PR #1, opened 2017-12-24
It is christmas

benibela:
and the duck needs to get dressed


samcarter8:
This would really close a circle as the `tikzducks` package was developed using texstudio :)

PR #17, opened 2018-01-23
be more specific about glibc on backtrace functionality

Piraty:
other libc implementations like musl libc don't provide execinfo nor
backtrace functionality.

sources:
- (fail logs on a build against musl)
- http://www.openwall.com/lists/musl/2015/04/09/3

### DIFF
--- a/src/debughelper.cpp
+++ b/src/debughelper.cpp
@@ -3,6 +3,12 @@
 #include "smallUsefulFunctions.h"
 #include <stdio.h>
 #include <stdlib.h>
+
+//backtrace is only available on GNU libc
+#ifndef __GLIBC__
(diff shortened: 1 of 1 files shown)


sunderme:
you patch breaks compilation on linux (see travis-ci).
Please fix !!!

Piraty:
yep, my fault.
good thing you guys have travis now.

sunderme:
@Benito: any comments ?

Piraty:
We were at this before, see [**here**](https://sourceforge.net/p/texstudio/bugs/1980/) and [**here**](https://sourceforge.net/p/texstudio/feature-requests/1186/).
Checking for `__GLIBC__` macro is what most devs do to ensure glibc specific stuff is there.

PR #55, opened 2018-03-07
Gives the option to compile using multiple cores (default 1)

joaofl:


sunderme:
/proc/cpuinfo is not available on OSX.
Furthermroe, why is the number of cores limited to 8 ?

joaofl:
What do you suggest to do for the OSX? I dont think checking which OS is a good solution, since it adds even mode platform specific code. Should I take the suggestion out?

I tough it was a good think to up-limit the number of cores, because specifying too many may crash the system. Most of the desktops nowadays are limited to 8, at most 16 cores...

What do you suggest?

sunderme:
the easiest solution is probably to compile with "make -j 4" without any questions (2-4 cores are typical).
BUILD.sh is anyway only a way to compile for the lazy ...

joaofl:
I wouldn't say lazy, but unexperienced. Makes it easier to contribute to the project as well. And a 4 fold compiling speedup makes it easier as well.

joaofl:
Ok. ill remnove the question as well.

sunderme:
okay, merged. Thanks.

PR #69, opened 2018-03-19
Support LaTeX tools for Japanese

tats-u:
Most of TeX users use pdfLaTeX to create documents, but Japanese don't.
Most of they usually use pLaTeX instead.  It is derived from "Publishing TeX", which is released by ASCII Corp. (publisher in Japan) in 1987 to support Japanese.  Unfortunately many class files for Japanese is compatible only with pLaTeX, not pdfLaTeX.  So are those used in societies and universities in Japan.
Makeindex and BibTeX are also modified to support Japanese ages ago.  They are renamed as `mendex` and `pbibtex` respectively.
To make the things worse, these programs for Japanese don't support Unicode under the hood, and modified programs to support Unicode are created.  They are called upLaTeX, upmendex, upBibTeX.
Japanese users of TeXStudio have to rewrite `latex` as `platex` for example to compile Japanese documents until now.  However, it is not identical.


sunderme:
why is it necessary to define extra commands ? it is not sufficient to cal platex in case of LATEX ? (man e.g. mendex in case of makeindex)

tats-u:
I have papered over this issue by the method you just proposed.    However, I'd like TeXStudio to avoid the bother of changing the command by, for example, adding commands like this PR, or check whether TeXLive runs on the Japanese locale and rewrite it if so.
FYI, [LyX](http://www.lyx.org), a GUI editor of TeX documents, separates tools for Japanese from those for other languages.


Possibly, non-Japanese people studying in Japanese are bothered by the issue you can't toggle latex (DVI mode of pdfLaTeX) and pLaTeX seamlessly.

tats-u:
To put it simply, the reasons why I define new commands for (u)pLaTeX and so on are:
1. LyX took the same method.
2. Simply, pLaTeX is not the DVI mode of pdfLaTeX
3. Someone may be use more than one of pLaTeX, upLaTeX, and the DVI mode of pdfLaTeX (I use pLaTeX, upLaTeX, and LuaLaTeX)

P.S.
I wish TeXStudio could automatically detect whether the locale is Japanese, and change the default compiler to `txs:///platex`, and the process to build and show documents to `txs:///compile | txs:///view` by later commits.  This allows Japanese users, especially students, professors, to compile TeX documents without settings as others can.  I wish this PR will be the 1st step in it.

P.P.S.
https://travis-ci.org/texstudio-org/texstudio/jobs/355300623
The failure of this build by Travis is so weird because downloading of packages spends too long time.  I wish I could try one more time.

timhoffm:
> The failure of this build by Travis is so weird because downloading of packages spends too long time. I wish I could try one more time.

job restarted.

sunderme:
please take all translation files file s(.ts , .qm) out of this pull-request.

tats-u:
@timhoffm Thank you, it successfully passed this time.
@sunderme Certainly, I will rebase the commit and push again with the force.

tats-u:
.ts & .qm files were removed.

tats-u:
Oh, Commands of chains (PS chain, DVI chain, and so on) must be modified.  Too early to be merged.

tats-u:
We have to divide commands `txs:///compile` into, for example, `txs:///make-pdf` and `txs:///make-dvi` in advance.
I will reopen this issue if it has been done.

PR #113, opened 2018-04-12
Update physics.cwl

thatlittleboy:
There are several matrix commands (e.g., `\dmat`, `\imat`) defined in the [`physics`](http://mirror.ox.ac.uk/sites/ctan.org/macros/latex/contrib/physics/physics.pdf) package that must be surrounded by a "matrix" environment, e.g.:

```
\mqty{\imat{2}}
\mqty(\dmat{1,2,3})
\mqty(\admat{1,2,3})
```

and these have corresponding entries in the existing `physics.cwl` file:

```
\mqty(\imat{number})#m
\mqty(\dmat{matrix element 1,matrix element 2,matrix element 3,...})#m
\mqty(\admat{matrix element 1,matrix element 2,matrix element 3,...})#m
```

However, these do not "work" since the `\imat`, `\dmat` etc. are still unrecognized commands, as far as TXS is concerned. (Pardon the colours, it was just so I could identify them easier in my original document.)

texstudio_2018-04-12_16-15-30

Since, as far as I know, there isn't support for classification of "nested" commands to be recognized (as in this case), I propose to include the `\imat`, `\dmat` etc. commands in the cwl, but with `#S` classification so that:

1. we avoid suggesting these (`\imat`, `\dmat` etc. ) to newbies as commands to use outside of `\mqty()`, `pmatrix` environments etc., since those will give an error ;
2. while still allowing TXS to recognize these as 'valid' commands.

### DIFF
--- a/completion/physics.cwl
+++ b/completion/physics.cwl
@@ -320,15 +320,15 @@
 \vqty{expression}#m
 \vu*{expression}#m
 \vu{expression}#m
-\diagonalmatrix{matrix element 1,matrix element 2,...}#Sm
-\dmat{matrix element 1,matrix element 2,...}#Sm
-\antidiagonalmatrix{matrix element 1,matrix element 2,...}#Sm
-\admat{matrix element 1,matrix element 2,...}#Sm
-\identitymatrix{n}#Sm
-\imat{n}#Sm
-\paulimatrix{n}#Sm
-\pmat{n}#Sm
-\xmatrix{matrix element}{n}{m}#Sm
-\xmat{matrix element}{n}{m}#Sm
-\zeromatrix{n}{m}#Sm
-\zmat{n}{m}#Sm
+\diagonalmatrix{matrix element 1,matrix element 2,...}#*m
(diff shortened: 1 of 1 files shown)


sunderme:
you need to remove the "S" from the flags (e.g. \imat{n}#Sm -> \imat{n}#m )

thatlittleboy:
@sunderme No, the `#S` is deliberate. I don't think it is a good idea for `\imat{n}` to appear in the autocompleter because it is not a valid command.

sunderme:
maybe you want to mark it as "unusual" command by marking it wirh "*" ?
Otherwise the completer is completely disabled and you can't even use it to change something from e.g. diagonalmatrix to zeromatrix ...

thatlittleboy:
@sunderme Okay, as I said, my only worry by having it appear in the completer is that newbies (to the physics package) might think it is okay to use it independently outside a matrix env. But there's no need for us to go into a deep discussion about it, either way is fine by me since it does not affect me. I'll make the changes.

sunderme:
merged

PR #184, opened 2018-06-14
added legend-related entries

jagjordi:


sunderme:
the added commands have arguments, please add the arguments into the cwl
(it really makes no sense to optimize the cwl manually and then not put in the commands thoroughly)

sunderme:
do you want to update this PR ?

sunderme:
no update

PR #191, opened 2018-06-16
Update siunitx.cwl, mathtools.cwl; New tcolorbox.cwl

thatlittleboy:
Changelog:

**siunitx.cwl**
- Removed `\cancel`: not a default command in `siunitx`; requires the `cancel` package to be loaded separately
- Added keyvals for `\sisetup`

**mathtools.cwl**
- Load `graphicx.cwl` alongside, since it is automatically loaded by `mathtools`
- Edited keyval syntax

**tcolorbox.cwl**
- Did my best to include everything. Most (90%) of the bulk is just repeated keyval options for the various commands/environments. If there is a way to declare the keyvals for a group of commands at once, then I can shrink this file down.

I'm not too sure the exact protocol for adding a new `.cwl`: I can only surmise that `completion.qrc` needs to be updated.


sunderme:
Thanks

PR #194, opened 2018-06-18
Add appimage to travis upload

LyzardKing:
This adds an automated appimage build to travis.
Like the windows and macos buils, since texstudio is being built it might as well be packaged and released on github as an appimage, available as soon as the other packages are.
At the moment I didn't test the push to github(but that's like the other ones you already have)
I had set up an upload to transfer.sh so I could download and test, and it seems that the images work.
I didn't include texlive, since I don't think the win/macos versions do, and I think it's better to keep them separate...

### DIFF
--- a/travis-ci/package.sh
+++ b/travis-ci/package.sh
@@ -83,7 +83,7 @@ cat "${TRAVIS_BUILD_DIR}/travis-ci/bintray.json"
 fi
 
 
-if [ "${QT}" = "qt5" ]; then
+if [ "${QT}" = "qt5Release" ]; then
 	print_info "Running linuxdeployqt"
 
 	make INSTALL_ROOT=appdir -j$(nproc) install ; find appdir/
@@ -117,7 +117,6 @@ if [ "${QT}" = "qt5" ]; then
 		"publish": true
 	}
 EOF
-
 
 cat "${TRAVIS_BUILD_DIR}/travis-ci/bintray.json"
 fi
--- a/travis-ci/package.sh
+++ b/travis-ci/package.sh
@@ -82,6 +82,46 @@ EOF
 cat "${TRAVIS_BUILD_DIR}/travis-ci/bintray.json"
 fi
 
+
+if [ "${QT}" = "qt5" ]; then
+	print_info "Running linuxdeployqt"
+
+	make INSTALL_ROOT=appdir -j$(nproc) install ; find appdir/
+	wget -c -nv "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage"
+	chmod a+x linuxdeployqt-continuous-x86_64.AppImage
+	unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
+	export VERSION=${TRAVIS_OS_NAME}-${VERSION_NAME}
+	./linuxdeployqt-continuous-x86_64.AppImage appdir/usr/share/applications/*.desktop -bundle-non-qt-libs -extra-plugins=iconengines/libqsvgicon.so
+	./linuxdeployqt-continuous-x86_64.AppImage appdir/usr/share/applications/*.desktop -appimage
+	mv "${TRAVIS_BUILD_DIR}/TeXstudio-${VERSION}-x86_64.AppImage" "texstudio-${VERSION}-x86_64.AppImage"
+
+
+	print_info "Preparing bintray.json"
+	cat > "${TRAVIS_BUILD_DIR}/travis-ci/bintray.json" <<EOF
+	{
+		"package": {
+			"name": "texstudio-linux",
+			"repo": "texstudio",
+			"subject": "sunderme"
+		},
+
+		"version": {
+			"name": "${VERSION_NAME}",
+			"released": "${RELEASE_DATE}",
+			"gpgSign": false
+		},
+		"files":
+		[
+			{"includePattern": "${TRAVIS_BUILD_DIR}/texstudio-${VERSION}-x86_64.AppImage", "uploadPattern": "texstudio-${VERSION}-x86_64.AppImage"}
+		],
+		"publish": true
+	}
+EOF
+
+
+cat "${TRAVIS_BUILD_DIR}/travis-ci/bintray.json"
+fi
+
 if [ "${QT}" = "qt5win" ]; then
 	print_info "package build into zip for win"
 	print_info "make installer"
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ matrix:
     os: linux
   - env: QT=qt5NoPoppler
     os: linux
+  - env: QT=qt5Release
+    os: linux
   - env: QT=qt5win
     os: linux
 dist: trusty
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,4 +59,13 @@ deploy:
     on:
       tags: true 
       condition: -f "${TRAVIS_BUILD_DIR}/texstudio-${TRAVIS_TAG}-win-portable-qt5.zip"
+  - provider: releases
+    skip_cleanup: true
+    api_key:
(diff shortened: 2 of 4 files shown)


LyzardKing:
This is for reference the latest image built by travis:
https://transfer.sh/5EkYj/texstudio-linux-2.12.9-201806181644-git_3152a46-x86_64.AppImage
(I disabled the push to transfer.sh in the PR)

probonopd:
According to the GitHub page, the official name is _TeXstudio_, and hence that capitalization is what I recommend for naming the AppImage (just like on the Mac, one would call it TeXstudio.app and TeXstudio.dmg).

Especially for continuous builds, also uploading the zsync file is the way to go, as it allows for binary delta updates. In which case renaming the AppImage would break the mechanism.

LyzardKing:
As far as I can tell the macos/windows variants don't bundle texlive, since it's not really a static thing.
One might need to add a part to texlive for example, and that would require a complete reinstall.
I was going for consistency with the other releases.
Regarding the name, I changed it from TeXstudio to texstudio (again, for consistency).
If that can break appimage updates or similar it can easily be changed by removing the cp line.

sunderme:
I don't see the point of bundling txs with texlive either. If so, you need to make sure that the bundle can be extended up to texlive full. And furthmore it would need to be updated with every tiny update of texlive ...

 How much access to external programs has an appimage ?

So, in general, I don't see the advantage of containers but maybe I am old fashioned and I like my programs lean and fast ...

LyzardKing:
As far as I can tell it has complete access.. It's not sandboxed by default.
You can with things like firejail, but it's an addon..
I't simply bundle of libraries, a part from the ones that are included in all the distros.
(In this case mostly qt stuff)
It's a nice way to test and get the latest version.

probonopd:
> As far as I can tell it has complete access.. It's not sandboxed by default.

Yes.

But my use case is:

As a user, I want to download _a single file_ and have a _fully working application_ without having to download additional stuff and without having to fiddle around.

LyzardKing:
The issue there is that there is no one static version.
So it would be making too many assumptions.
Most people that use Tex use it for scientific texts, and need specific
modules.
So they would need to reinstall anyway.
It might be more complicated to understand that you have a static version
that you have to manually change in the settings

LyzardKing:
On first run I do get these errors on mate. I think it's a missing qt library.. some qcursor thing? I can't really tell...


```
FAIL!  : QDocumentCursorTest::bidiMoving(with all markers 2) '(c.columnNumber())==(newColumn)' returned FALSE. (equal failed: got "10" !=expected "11"  reason:column)
   Loc: [src/tests/qdocumentcursor_t.cpp(727)]
FAIL!  : QDocumentCursorTest::bidiMoving(with all markers 2r) '(c.lineNumber())==(newLine)' returned FALSE. (equal failed: got "2" !=expected "1"  reason:line)
   Loc: [src/tests/qdocumentcursor_t.cpp(726)]
FAIL!  : QDocumentCursorTest::bidiMoving(with all markers 3r) '(c.columnNumber())==(newColumn)' returned FALSE. (equal failed: got "0" !=expected "11"  reason:column)
   Loc: [src/tests/qdocumentcursor_t.cpp(727)]
FAIL!  : QDocumentCursorTest::bidiMoving(with other markers 3) '(c.columnNumber())==(newColumn)' returned FALSE. (equal failed: got "8" !=expected "9"  reason:column)
   Loc: [src/tests/qdocumentcursor_t.cpp(727)]
FAIL!  : QDocumentCursorTest::bidiMoving(with other markers 3r) '(c.lineNumber())==(newLine)' returned FALSE. (equal failed: got "2" !=expected "1"  reason:line)
```

sunderme:
that are not errors but self-test results
txs should have been compiled for release instead of debug ....

LyzardKing:
Ah ok.
But, Travis is building a macos package.
Is it rebuilding after the tests?

LyzardKing:
I see now, in configure.sh only the linux builds are configured with debug.
to release we would have to either remove that or have another explicit target.
Something like qt5Release... with qmake texstudio.pro CONFIG-=debug

LyzardKing:
I'll temporarily add an upload to transfer.sh to test the non debug image. 
@sunderme Do you think the explicit release target is a good idea?

LyzardKing:
Ok...this works much better, and it's half the size, like the macos and windows variants..
For testing purposes this is the appimage generated:
https://transfer.sh/11AXln/texstudio-linux-2.12.9-201806190708-git_65c1d04-x86_64.AppImage
I'll now remove the upload to transfer.sh

I'm still not sure hot to get the "auto updates".
@probonopd How do I get the function where I can right click on the appimage/menu item and and have the update option?
Is that something that can be done with the travis integration?

probonopd:
> How do I get the function where I can right click on the appimage/menu item and and have the update option?

Short version:
You need to embed update information into the AppImage, and your user needs to be running `appimaged` and have AppImageUpdate on their `$PATH`.

Is that something that can be done with the travis integration? 

Yes. It also can be done without Travis CI.

Long version:

If you are using Travis CI, all you need to do is to export `GITHUB_TOKEN` as a secret variable and set up `uploadtool` as described at https://github.com/probonopd/uploadtool; `linuxdeployqt` will then automatically inject the needed update information into the AppImage and generate an `.AppImage.zsync` file in addition to the `.zsync` file.

To get the click update e.g., in the Dock, the user needs to run [appimaged](https://github.com/AppImage/AppImageKit/releases) and needs to have a file called "AppImageUpdate" (take [the AppImageUpdate AppImage](https://github.com/AppImage/AppImageUpdate/releases) and rename it that) on their `$PATH`, e.g., in `/usr/local/bin/AppImageUpdate`.

travis-ci/package.sh
```diff
@@ -82,6 +82,45 @@ EOF
...
+	chmod a+x linuxdeployqt-continuous-x86_64.AppImage
+	unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
+	export VERSION=${TRAVIS_OS_NAME}-${VERSION_NAME}
+	./linuxdeployqt-continuous-x86_64.AppImage appdir/usr/share/applications/*.desktop -bundle-non-qt-libs -extra-plugins=iconengines/libqsvgicon.so
+	./linuxdeployqt-continuous-x86_64.AppImage appdir/usr/share/applications/*.desktop -appimage
+	mv "${TRAVIS_BUILD_DIR}/TeXstudio-${VERSION}-x86_64.AppImage" "texstudio-${VERSION}-x86_64.AppImage"
+
+
+	print_info "Preparing bintray.json"
+	cat > "${TRAVIS_BUILD_DIR}/travis-ci/bintray.json" <<EOF
```
LyzardKing: @probonopd
I know nothing about github releases...
would the uploadtool have to replace this part?
Is there no easier way to do it?
I see that the tool is used to have continuous builds, but that's out of the scope for this PR.
I only want, on a release (when this code runs) to get an appimage and (if it's what's needed for an update)
a zsync file

LyzardKing:
@probonopd
I know nothing about github releases...
would the uploadtool have to replace the json part?
Is there no easier way to do it?
I see that the tool is used to have continuous builds, but that's out of the scope for this PR.
I only want, on a release (when this code runs) to get an appimage and (if it's what's needed for an update)
a zsync file

LyzardKing:
@sunderme 
Did I add the release properly?
I changed the travis.yaml and package.sh for that...
But as I said I don't have much experience with github releases

probonopd:
@LyzardKing 

> I know nothing about github releases...
> would the uploadtool have to replace the json part?

`provider: releases` uploads _releases_ only whereas `uploadtool` would upload the AppImage for _each build_ which is useful e.g., for continuous builds intended for testing.

You could even combine the two if you like.

LyzardKing:
@sunderme This is up to you.
I'm not sure how your build and upload system works at the moment.
@probonopd So uploadtool doesn't upload the release version?
So if one wanted to have release + continuous it would need both, is this the logic?
The disadvantage there is that in the github releases page the latest stable would be in second place, I guess...

sunderme:
I will adapt the upload strategies

LyzardKing:
Great! Thanks.

probonopd:
> @probonopd So uploadtool doesn't upload the release version?
So if one wanted to have release + continuous it would need both, is this the logic?

Yes.

sunderme:
the development versions are uploaded to bintray, only releases are presented on github

probonopd:
Yes, that's a good strategy as well; however in this case you won't get binary delta updates for the continuous builds (unless there is a way to upload two files to the same transfer.sh directory...)

sunderme:
possibly, but unless there is a good proposition, I am not going change this set-up

probonopd:
Fwiw, gitlab.com lets you keep build artifacts of each build in an unobtrusive way. Just saying because I was pleasantly surprised when I discovered this recently.

LyzardKing:
@sunderme So when a release comes out there'll be an appimage on github releases?
Is the daily build I can find here built with this appimage tool then?
https://bintray.com/sunderme/texstudio/texstudio-linux

If so it's good for me!

sunderme:
that is how it is set-up right now

sunderme:
yes

PR #195, opened 2018-06-19
Update glossaries.cwl

jagjordi:
added info for \glsdisp


sunderme:
if you want to use the command in the completer, you need to remove the `#S`
With that option, the command is used only for syntax checking

jagjordi:
I removed the #S, I use that command frequently so I thinks it makes sense to have it appear in the completer.

sunderme:
thanks

PR #199, opened 2018-06-25
Fix some Chinese translation errors.

p1gd0g:
There should not be two same translations.
I am a Chinese.


p1gd0g:
Done, thanks for guides.
BTW, I doubt the English description of "Show difference between two files in relation to base file".
E.g., my workflow is 
> open current doc -> "Show difference between two files in relation to base file" -> "select compare doc" -> "select base doc"

but the result looks likes 

> open current(base) doc -> "Show difference between two files in relation to base file" -> "select the first compare doc" -> "select the second compare doc"



I don't if it is clear, we may talk about it.

Thanks for your works for this great editor.

PR #211, opened 2018-07-01
Add Autocomplete file for AltaCV

foohyfooh:


foohyfooh:
There is one issue with this cwl file for AltaCV. The \cvsection shows the name of the sidebar file in the list of sections if one is specified in the square brackets. How can I make it that it always shows what is specified as the header and never the sidebar file name?

timhoffm:
Currently you can't. Structure-like commands are assumed to behave like `\section[short_title]{title}`. If there is an optional argument, TXS interprets it as the short_title to be used in table of contents.

sunderme:
are you going to do anything about the cvsection issue ?

timhoffm:
Not right now. Probably the correct solution is

When encountering a `L*` command in a cwl file:
- determine the argument named `short title`
- if that does not exist, determine the argument named title.
- Otherwise fall back to the last argument (or alternatively simply display the whole command).

The argument position and type (mandatory/optional) must be stored with the command definition.
I don't think we have infrastructure for this right now. Or do we?

When parsing a tex file and encountering a `L*` command:
- obtain the argument to use from the command definition.
- parse extract the correct text from the command args and use it in the structure.

PR #241, opened 2018-07-15
Add new screenshot to appdata file

Alexander-Wilms:
The screenshot now shows the system theme.


benibela:
Should we keep the screenshot on github or upload it to the webserver?

Alexander-Wilms:
Having it in the repo right next to the xml file is quite practical I think. But do whatever you prefer.

PR #279, opened 2018-08-13
Update cwls

thatlittleboy:
Changes:

- short text for `\caption` to be treated as text (with spell check)
- changed arg names for theorems to be more similar with `amsthm` documentation (doc calls them heading, not caption)
- make `\em` unusual (deprecated command)


sunderme:
thanks

PR #286, opened 2018-08-15
Update subcaption.cwl

lhecht:
add the \subref command
I dont know if it would be good to add \subref* as well?


sunderme:
thanks

Issue #289, opened 2018-08-16
Enhancement: By mouse over file structure add to the information the highest structure of the incluse


li-po:
### Actual behaviour
By hovering with the mouse an included file in the file structure, we get following informations:

- File name
- Line in the including file

### Enhanced behaviour
To the two information it would be very helpful to have also the title of the highest structure of the included file, probably most of the time _chapter_ or _part_

li-po:
This code does the job. Notes:

- it does assume that the file is `.tex`. Is this a problem?
- a two levels structure is shown (can be reduced to one by removing the internal loop)
- the top levels (one or more) are bold
- only the structure of opened files is shown
- the whole thing could be switched on/off by a configurable option, but I do not see a reasung to see this information as unnecessary

```
diff --git a/src/latexstructure.cpp b/src/latexstructure.cpp
index 8a07586e..6ce95285 100644
--- a/src/latexstructure.cpp
+++ b/src/latexstructure.cpp
@@ -214,6 +214,19 @@ QVariant LatexDocumentsModel::data ( const QModelIndex &index, int role) const
                        QString tooltip("<html><b>" + htmlTitle + "</b>");
                        if (entry->getCachedLineNumber() > -1)
                                tooltip.append("<br><i>" + tr("Line") + QString("</i>: %1").arg(entry->getRealLineNumber() + 1));
+            if (entry->document != nullptr) {
+                LatexDocument *dc = documents.findDocument(entry->title+".tex");
+                if (dc != nullptr) {
+                    foreach (StructureEntry *se, dc->baseStructure->children) {
+                        if (se->type == StructureEntry::SE_SECTION ) {
+                            tooltip.append(QString("<br><b>%1</b>").arg(se->title));
+                        foreach (StructureEntry *se2, se->children) {
+                            tooltip.append(QString("<br>&nbsp;%1 ").arg(se2->title));
+                        }
+                        }
+                    }
+                }
+            }
                        if (!entry->valid)
                                tooltip.append(QString("<br><font color=\"%1\">%2</font>").arg(missingFileColor.name(), tr("File not found.")));
                        return QVariant(tooltip);

```

li-po:
`if (entry->document != nullptr) {` is not necessary, I just forgot it

sunderme:
things would be simpler with pull-requests, see github docu

PR #293, opened 2018-08-18
Issue #289: Enhancement: By mouse over the file structure ...

li-po:


sunderme:
I have tried the new code, I am not convinced:

1. adding .tex to the title fails if the include already has the complete filename e.g. includedFile.tex
2. if there is no section defined in the include, nothing extra is shown
3. there is no visual indication that the additional information is section/subsection of the included file.

I would rather like to see the first lines of the included file as quotation similar to label/bibid tooltip behaviour (though probably for tooltips from the commands in the text). This usually does show section/subsection when they are present in the first lines ...

li-po:
1.  I could add a check if there is already an extension `.tex`, not an issue
2. With section do you mean in general part,chapter, section,.... ? I see such a case at best as the not most probable case. In this case the message "No sections" could be given, or nothing.
3. An header "Sections:" or "Contents:" could be added as caption. Alternatively the variable `se->level` could be used (I do not now its content right now) a `[P]` for part, `[SS]` for subsection a.s.o.  could be prepended to the sections.

> I would rather like to see the first lines of the included file as quotation similar to label/bibid tooltip behaviour 

I do not agree. The user shouldn't look for the content, he/she should see the content. If someone has an header with comment (copyright,...) he/she would not get added-value from the first lines. My opinion.

sunderme:
I have implemented the tooltip with showing the first lines.
If that is still insufficient for you, then you could add your functionality with config option

Issue #300, opened 2018-08-22
Long environment causes incorrect syntax highlighting


i3v:
### Environment
- TeXstudio: 2.12.10 (git 2.12.10)
- Qt: Qt Version 5.8.0, compiled with Qt 5.8.0 R
- OS: Windows 10
- TeX distribution: TexLive 2018

### Expected behavior
Syntax highlighting should work correctly regardless of the number of lines between `begin{scope}` and `end{scope}`.

### Actual behavior
It looks like the syntax checker forgets about "opening" statements after a fixed number of lines, and thus highlights innocent statements as incorrect in some cases.

### How to reproduce
Here's an MWE:

```
\documentclass{article}
\RequirePackage{graphicx}
\usepackage{tikz}                         


\newcommand{\somePlot}[1]{
	\begin{tikzpicture}
	\node (image) [anchor=south west,inner sep=0] { #1 };
	\begin{scope}[x={(image.south east)},y={(image.north west)}]
%
%
%
%
%
%
%
%
%
%
%
%
%
%
%
%
%
%
%  The number of commented out lines here is important.
% 
%  x
%  xx
%  xxx
%
%
%
%
		\draw [blue, line width=0.2em] (1em,1.2em) -- (5em,1.2em) {};
	\end{scope}
	\end{tikzpicture}%
}

\begin{document}
	\begin{figure}
		\somePlot{\includegraphics[width=0.2\linewidth]{example-image-a}}%		
	\end{figure}
\end{document}
```
It works OK for me, but TeXstudio highlights 3 statements:
![image](https://user-images.githubusercontent.com/920364/44484781-6021e600-a657-11e8-8128-224486693bb8.png)

If I remove a single line (e.g. line 30), and do "Idefix->Refresh Structure" after that (to force a refresh), the highlighting changes (now only 2 statements are highlighted):
![image](https://user-images.githubusercontent.com/920364/44484979-e3433c00-a657-11e8-9cd2-087222842029.png).

If I repeat that, there would be only 1 statement highlighted:
![image](https://user-images.githubusercontent.com/920364/44485076-2b625e80-a658-11e8-9106-d37dd283ed77.png)

Finally, after removing one more line, there would be no lines highlighted:
![image](https://user-images.githubusercontent.com/920364/44485122-58af0c80-a658-11e8-9c61-7b1080906eee.png)

i3v:
[...]

PR #301, opened 2018-08-22
add classification for \newcommands

thatlittleboy:
Added missing `#d` classification for `\newcommand` and friends, which might be the cause for #300.

Also, fixed `ensuremath` arg label to be in line with `latex2e.pdf` texdoc. `text` is wrong in any case, since one uses it like:

```
\newcommand{\ab}{\ensuremath{(\delta, \varepsilon)}}
```


thatlittleboy:
@i3v The `\ensuremath` 'fix' was just to remove TXS treating the argument as actual text and prevent spell-checking. For example,

```
\ensuremath{asdf askdf  alsdf}
```
should no longer complain about wrong spelling. (might have to disable `latex-document.cwl` from the settings to see the effect.)
It will still complain about math being outside math environment. Can't do too much else about that now I think.

The `\newcommand` ones, I think there might be an issue with the `def` argument label and/or the `#d` classifier. `\NewDocumentCommand` (and friends) from `xparse` package also exhibit similar behaviour (and those already have the `#d` classifier attached to them), so it's not specific to `\newcommand`, and also my insertion of `#d` in this PR is unlikely to be a fix for your issue.

But the `\newcommand{}{}` commands should have `#d` attached to them nonetheless.

i3v:
@thatlittleboy ,
OK, thanks for the details. 

I think I get why it was not working for me - that's because "latex-document.clw" [seem to be automatically loaded](https://github.com/texstudio-org/texstudio/blob/7819dd3124e5bad8f39a125fe00ee5a5643382a6/completion/cwlAliases.dat#L2-L3) (even though it was unchecked), when there's a `\documentclass{article}`.

And I also agree that it would be interesting to know, if it's intentional that the `formula` is not mentioned in the [cwl user manual](https://github.com/texstudio-org/texstudio/blob/d7acf2c97afafa1b1acaca3183f99109d0d154f9/utilities/manual/usermanual_en.html#L1447).

PR #309, opened 2018-08-30
Add minutes.cwl to completion.

dkolb:
Adds in completion for the minutes package as found at https://ctan.org/pkg/minutes


sunderme:
thanks

PR #330, opened 2018-09-23
src/main.cpp: set desktop file name to avoid a broken icon under KDE Wayland

Alexander-Wilms:
This fixes the missing icon in the title bar.

Before:
grafik

After:
grafik


timhoffm:
Thanks!

PR #341, opened 2018-10-01
Update class-beamer.cwl

samcarter8:
Added new variations of `\insert...` commands for frame numbers

(these were introduced in https://github.com/josephwright/beamer/commit/0a7af8a222ac680946df60ee3afa6328cfa6e2ba)


sunderme:
thanks

PR #358, opened 2018-10-20
Minor improvements (probably)

dima74:
Also please check [this fragment](https://github.com/texstudio-org/texstudio/blob/c86539b711e458890b07af2dcf83b25146ddb55c/src/tabdialog.cpp#L128-L137), it is `if ... else ...` statements with identical branches.


sunderme:
thanks

PR #379, opened 2018-11-11
cwl files update

thatlittleboy:
- New `cite.cwl`
- Updated `xcolor.cwl` to include `\rowcolors` command and more. Updated argument names to match documentation.


sunderme:
thanks

PR #393, opened 2018-11-18
Update French translations

dbitouze:
BTW, the changing line numbers are very annoying: they often lead to impossible automatic merge and I didn't find another solution than deleting my TXS' fork, fork again, clone, launch `linguist` with the new `texstudio_fr.ts` file and (hopefully possible) open also the old one I set aside in order to not have to rewrite all the missing translations.


sunderme:
lupdate should take care of it. Anyway changing line numbers are impossible to avoid.

sunderme:
if you used https://www.transifex.com/texstudio/texstudio , things would be simpler (for me). All other translations come from that site, German included.

dbitouze:
>lupdate should take care of it.

However, I always run it before running `linguist`.

>Anyway changing line numbers are impossible to avoid.

Sigh...

dbitouze:
>if you used https://www.transifex.com/texstudio/texstudio , things would be simpler (for me). All other translations come from that site, German included.

OK, I'll try to use it, though suggestions seems to work better with `linguist`. Could you please upload my last translations to `transifex`?

dbitouze:
Another advantage of `linguist` (over `transifex`) is that, when appropriate, it displays the dialog windows and hence the context, which is sometimes necessary for right translations.

sunderme:
your translation has been uploaded to transifex

dbitouze:
I proceed to a `git pull` for TXS 2.12.12 and run:
```bash
texstudio --update-translations && lupdate texstudio.pro && linguist translation/texstudio_fr.ts
```
in order to see if there were new strings to translate. There was one ("Babel") so I went to [transifex](https://www.transifex.com/texstudio/texstudio/translate/#fr/texstudio_ents/60319933) but it said there is 0 strings that are not translated.

PR #402, opened 2018-11-25
cwl improvements

thatlittleboy:
changelog

- new `enumitem.cwl`
- added commands to `class-beamer.cwl`, cf. [Why TeXStudio can't do auto-completion for Beamer?](https://tex.stackexchange.com/q/449273/)
- added `\bigl(%|\bigr)` variants (we only had `\big(%|\big)` before), cf. [Why use the control sequences \bigl, \biggl, \bigr or \biggr, as I can always use \big or \bigg?](https://tex.stackexchange.com/q/19480/)


sunderme:
thanks

PR #437, opened 2018-12-16
Fix error in minutes autocomplete.

dkolb:
When I originally submitted the `minutes` package's autocompletion I miscapitalized the `\decisiontheme` command.  Sorry about that!


sunderme:
Thanks

PR #463, opened 2019-01-06
Context menu: Add to ignore list

jonasstein:
The item Add to Dictionary was misleading. 
Words are not added to the Hunspell dictionary, but to a local ignore list in
~/.config/texstudio/*.ign


timhoffm:
That was changed the other way round [some time ago](https://sourceforge.net/p/texstudio/feature-requests/728/), because the notion of adding to a dictionary is easier to understand than an ignore list.

jonasstein:
That makes sense too. However if I select `add to dictionary` I would expect the words to be added to the users Hunspell dictionary. I close this here and open a feature request.

PR #466, opened 2019-01-10
Update texstudio_de.ts

0xflotus:


sunderme:
please use transifex.com/texstudio for translations, see webpage. Keeping transifex and pull-requests in sync would be tedious.

PR #499, opened 2019-02-06
Fontenc in quickstart

### LINKED ISSUE #347: Feature Request - Quick Start Wizard - fontenc option
### Environment
- TeXstudio:  2.12.10

### Expected behavior
An important missing feature in the quick start wizard is the possibility to select a fontenc option. Especially the T1 option of fontenc is useful for full glyph support when using utf8 encoding.
Since texstudio makes it possible to set the inputenc as utf8, support for fontenc is logical.

### Actual behavior
There is currently no fontenc option, only inputenc, in the quick start wizard.
The fontenc package needs to be manually added in the LaTeX file.

Timvde:
Fixes #347.

Source of font encodings: https://www.latex-project.org/help/documentation/encguide.pdf. I'm not 100% sure if they're all relevant, so it might be necessary to cut in the list.

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -29,8 +29,8 @@ qreal convertLatexLengthToMetre(const qreal &length, const QString &unit)
 }
 
 //options for the configmanager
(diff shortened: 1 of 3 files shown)


sunderme:
thanks

PR #509, opened 2019-02-16
Add Dutch as an option for babel in the wizard

Timvde:
Quite a few people around here use LaTeX, and we promote the use of TeXstudio in our LaTeX workshops, so it would be valuable to us if it was in the default list :)


sunderme:
thanks,
 Jan

Issue #260, opened 2018-07-28
Spell check of meaningful text inside commands (e.g. moderncv cventry{TEXT}) 


FranzFelberer:
### Environment
- TeXstudio: TeXstudio 2.12.10 (git 2.12.10)
- Qt: Qt-Version 5.11.1, kompiliert mit Qt 5.11.1 R
- OS: Mac 10.13.6 (17G65)
- TeX distribution: texlive

### Expected behavior
Spell check every meaningful text, that is shown in the output file

### Actual behavior
ignoring text entries for \cventry arguments that are displayed

### How to reproduce
open moderncv example from github
https://github.com/xdanaux/moderncv

sunderme:
txs uses the cwl files to determine the meaning of the arguments. Automatically scanned packages don‘t have that information and hence all arguments are considered to be non-text arguments. To change that you need to change the cwl file, see wiki and manual 
http://texstudio.sourceforge.net/manual/current/usermanual_en.html#CWLDESCRIPTION

sunderme:
I don‘t understand your first issue. It completes with the command which you want ?

2) the argument needs to be „text“ or „title“

FranzFelberer:
sorry for the confusion. I didn't make myself very clear:
1. my moderncv.cwl file is ignored by txs. Any idea why?

2. so the arguments should like below in order to spellcheck the second argument and use the text between % % for auto-completion?

\cventry{%<year%>--%<year%>}{text%<degree or job title%>}

FranzFelberer:
Doing what you mentioned solves the problem that the "manuell" spell checker checks this text, I was referring to the on-type spell checker that indicates mistakes while writing by a red unline-ripple. It shows this behavior if in the same argument the text has a property like \testbf.
See image: It indicates no error for the first "wrongwriten" but for the second which is in a \textbf environment.

![image](https://user-images.githubusercontent.com/20334762/43365865-af92acba-9333-11e8-9236-1113371b7973.png)

sunderme:
`\cventry{year-year}{degree or job title%text}{institution or employer%text}{city}{grade}{description%text}#n` works fine here

FranzFelberer:
Thank you, I got it working, but I experience some other strange behavior.

I defined a command in my tex file:
```
% switch between languages
\newcommand{\multilang}[3]{#1}	% #1 en, #2 fr, #3 de
```

In ~/.config/texstudio/completion/usercommands.cwl I add the line:
[...]

PR #511, opened 2019-02-18
Update class-moderncv.cwl

creat89:
I have modified the cwl file of  class-moderncv.cwl in order to allow the spell checker working in the entries of the CV. This is related to the issue #260


sunderme:
thanks

PR #545, opened 2019-03-22
Hanslollo patch translation

hanslollo:
A tiny change to the warning put out when a label is _defined multiple times,_ which should now be translated correctly. For reference, just google (with quotation marks) "multiple times defined" (returns 3k results, first few are just texstudio) versus "defined multiple times" (70k results of all kinds).


sunderme:
thanks

PR #558, opened 2019-04-10
[cwl] fix syntax of `\colorlet` in xcolor.cwl

muzimuzhi:
By [user manual of `xcolor`](http://mirrors.ctan.org/macros/latex/contrib/xcolor/xcolor.pdf), sec. 2.5.2, change the third argument of `\colorlet` from mandatory to optional.

image


sunderme:
thanks

PR #560, opened 2019-04-13
Fix rerunnable flag of user commands

MeanSquaredError:
The current version of TexStudio does not handle correctly rerunnable user commands. The user command can be rerunnable only if the user command with its arguments is a substring of an built-in rerunnable command (one of the *tex commands).

For example if the builtin pdflatex command is:
pdflatex -synctex=1 -interaction=nonstopmode %.tex
then creating a custom user command
pdflatex -synctex=1 -interaction=nonstopmode -output-directory=/root/tex %.tex
cannot be rerunnable even if the user enables the "Repeat contained compilation commands" for the new command.

The proposed patch fixes this issue, but checking only the basename of the new user command, ignoring the arguments. As long as the basename matches with one of the builtin rerunnable commands, the user command is marked as rerunnable and repeating can be enabled for it.


sunderme:
thanks

PR #562, opened 2019-04-15
Fix tooltip preview of inline math when it is preceded by other text

MeanSquaredError:
At the moment TexStudio does not handle correctly inline math that is preceded by other text on the same line. For example if we have the following line of inline math
abc $1+2$ cba
then hovering the mouse over the opening or closing dollar sign ($) will not invoke the preview tooptip.
The problem is caused by the fact that LatexEditorView::showMathEnvPreview does not detect correctly the starting position of the inline math text. The proposed patch fixes this issue.

Please note that previewing old-style display math (surrounded by double-dollar signs, like $$1+2$$ ) still does not work, but fixing it is more of an enhancement than a simple bugfix.


sunderme:
thanks

PR #574, opened 2019-04-20
Convert UTC time to local time in PDFInfoDock.

clerkma:
To fix the mismatch of 'CreationDate' between pdf file and PDFInfoDock.


sunderme:
is the pdf creation time always stored as UTC ?

sunderme:
> is the pdf creation time always stored as UTC ?

seems like it.

sunderme:
thanks

clerkma:
thanks

PR #575, opened 2019-04-20
 Avoid spinlock and UI block while compiling files 

manudroid19:
This commit makes UI responsive while compiling, and allows to continue typing.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1645,21 +1645,11 @@ bool BuildManager::waitForProcess(ProcessX *p)
 	REQUIRE_RET(!processWaitedFor, false);
 	processWaitedFor = p;
 	m_stopBuildAction->setEnabled(true);
-	QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));
-	QTime time;
-	time.start();
-	while (p && p->isRunning()) {
-		if (time.elapsed() > 2000)
-			qApp->instance()->processEvents(QEventLoop::AllEvents);
-		else
-			qApp->instance()->processEvents(QEventLoop::ExcludeUserInputEvents);
-		// workaround for high CPU consumption when waiting
-		// TODO: replace this loop by notification mechansim
-		ThreadBreaker::msleep(1);
-	}
-	QApplication::restoreOverrideCursor();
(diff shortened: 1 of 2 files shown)


manudroid19:
Does it need to pass for Qt4?😢

manudroid19:
I see in your webpage that you are still uploading and supporting Qt4 builds, so it must work there. I have also read a bit in Qt's documentation and the same thing can be achieved using the old Qt4's signal connect syntax, so it would work in all envs. I will try tomorrow, sorry I didn't realize because I made my local tests with qt5...

sunderme:
I am not sure that you really solve a problem here.
The original code is supposed to be blocking for 2 seconds until switching to a responsive mode.
The idea is (probably) that you should not start multiple compile processes before the build is finished.

manudroid19:
> 
> 
> I am not sure that you really solve a problem here.
> The original code is supposed to be blocking for 2 seconds until switching to a responsive mode.
> The idea is (probably) that you should not start multiple compile processes before the build is finished.

No, I have tested that (multiple compile processes) and that's already checked here so it's not allowed: [https://github.com/texstudio-org/texstudio/blob/ba6dcbbe6a0f741ee49a48df4308e7116779ac25/src/buildmanager.cpp#L1466](url)
Keep in mind that while the UI in not responsive, it's still receiving all the events from the user input, and after 2 sec it will process all of them. So, pressing the button while the UI is not responsive has the same effect and my solution has no issues with that.

The "2 sec wait" thing is almost for sure because getting stuck in a while(1) loop INSIDE the main event loop and check in every cycle if the process has finished and if there is user has input is equivalent to a nonsense 100% CPU load time while compiling. The "2 sec wait" is a workaround to lower that CPU usage, but it's still an inefficient, anti-pattern and battery-killer approach, while it also annoys people like me that like to hit F6 and continue typing while it compiles. 

It's described in the previous code comments, and it is marked as TODO: [https://github.com/texstudio-org/texstudio/blob/ba6dcbbe6a0f741ee49a48df4308e7116779ac25/src/buildmanager.cpp#L1656](url)

I will try now to make that work in Qt4. If you are okay with this, I will close this pull request and open a new one, so it's cleaner what is being done. I wait for your approval; I am a noob in big project's github project management.

PR #576, opened 2019-04-21
Avoid spinlock and UI block while compiling files

manudroid19:
We come from here: #575 
This pull request pretends to fix the UI blocking and high CPU usage caused by the spinlock in the previous code while compiling.
This still does not allow for multiple processes to be launched, as comented in #575 
This pull request is done to fix #575 problems with Qt4 and  provide a cleaner pull request done in one commit.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1645,19 +1645,9 @@ bool BuildManager::waitForProcess(ProcessX *p)
 	REQUIRE_RET(!processWaitedFor, false);
 	processWaitedFor = p;
 	m_stopBuildAction->setEnabled(true);
-	QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));
-	QTime time;
-	time.start();
-	while (p && p->isRunning()) {
-		if (time.elapsed() > 2000)
-			qApp->instance()->processEvents(QEventLoop::AllEvents);
-		else
-			qApp->instance()->processEvents(QEventLoop::ExcludeUserInputEvents);
-		// workaround for high CPU consumption when waiting
-		// TODO: replace this loop by notification mechansim
-		ThreadBreaker::msleep(1);
-	}
-	QApplication::restoreOverrideCursor();
+    QEventLoop loop; //This approach avoids spinlock and high CPU usage, and allows user interaction and UI responsivness while compiling.
+    connect(p, SIGNAL(finishedProcess()), &loop, SLOT(quit()));
+    loop.exec(); //exec will delay execution until the signal has arrived
 	bool result = processWaitedFor;
     processWaitedFor = nullptr;
 	m_stopBuildAction->setEnabled(false);
@@ -2458,6 +2448,7 @@ void ProcessX::onFinished(int error)
 		readFromStandardError();
 	}
 	ended = true;
+    emit finishedProcess();
 }
 
 #ifdef PROFILE_PROCESSES
--- a/src/buildmanager.h
+++ b/src/buildmanager.h
@@ -294,6 +294,7 @@ signals:
 	void processNotification(const QString &message);
 	void standardOutputRead(const QString &data);
 	void standardErrorRead(const QString &data);
+    void finishedProcess();
 private slots:
 	void onStarted();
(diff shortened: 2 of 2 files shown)


sunderme:
@benibela @timhoffm any comments on this change ?

sunderme:
okay, this seems to work as advertised.
but ...
why keep "waitforprocess" at all. If nothing is stopped, we could simply handle everything with signal/slot chains ... @benibela ?

manudroid19:
> 
> 
> okay, this seems to work as advertised.
> but ...
> why keep "waitforprocess" at all. If nothing is stopped, we could simply handle everything with signal/slot chains ... @benibela ?

Well, it could be done but it would require a major refactoring of the codebase, because there is a quite big call stack with a lot of functions waiting for return. We would need to get to the first call and refactor from there to make the code fully asyncronous.
This would be the definitive solution and I plan to start working on that once this pull request is approved, but it would require more time and a deeper analysis of the code base.
I also plan to disable the compile buttons while executing the process to make it noticeable for the user that a new compilation launch will have no effect while one process is running.

benibela:
The idea behind blocking typing during the compilation was that the user should not insert/delete lines, or the line numbers from the log won't match those of the document and then the errors would be shown at the wrong lines.


>We would need to get to the first call and refactor from there to make the code fully asyncronous. This would be the definitive solution and I plan to start working on  that once this pull request is approved, but it would require more time  and a deeper analysis of the code base. I also plan to disable the compile buttons while executing the process  to make it noticeable for the user that a new compilation launch will  have no effect while one process is running.


That would be great

manudroid19:
> The idea behind blocking typing during the compilation was that the user should not insert/delete lines, or the line numbers from the log won't match those of the document and then the errors would be shown at the wrong lines.

The point is that the current code doesn't do that either. Every input from the user will be processed after 2s and the new/deleted lines will be there unmatched with the errors. Furthermore, it takes a lot of CPU usage and it makes the user experience worse. 

It should also be obvious to the user that inserting/deleting lines since hitting the compile button has that effect. Note also that less experienced users won't be hitting compile and continue typing quickly, they will wait for the compile process to finish. People will a bit of practice will find this logical, and there is nothing that we can do about it.

timhoffm:
We do actually take a snapshot of the lines at compilation time to allow pdf syncing to still work if some lines were added/removed. See `LatexDocument.lineFromLineSnapshot` and related methods. That could in principle be used as well for the logs (However, it's less important here because you usually track log entries right after compilation, so the document has probably not deviated much when you edit during compilation).

manudroid19:
Didn't notice that, good to know. I don't think that extending that sync mechanism to logs is worth the effort.

Anyway, the relevant point here is that this pull request does not modify the program behaviour relative to line syncing. Any other concern about the request?

sunderme:
I merge the code as the orginal code does not do much else.

manudroid19:
> > We would need to get to the first call and refactor from there to make the code fully asyncronous. This would be the definitive solution and I plan to start working on that once this pull request is approved, but it would require more time  and a deeper analysis of the code base. I also plan to disable the compile buttons while executing the process  to make it noticeable for the user that a new compilation launch will  have no effect while one process is running.
> 
> That would be great

I'm starting to work in this. Refactoring the compiling code means getting throught the call stack and make the code synchronous with signals. For that, I have to get rid of all the code that depends on the returns, so I'm analyzing all the calls to `BuildManager::runCommand()`.

I've almost finished that but there is one call I can't get to know what it is for (line 359):

https://github.com/texstudio-org/texstudio/blob/6978ea91f85b6444f75a9acc682194a74535973f/src/scriptengine.cpp#L355-L365

I'm using the debugger, but I can't get TexStudio to run that line of code.
I have no idea what that method is for.
I don't know what "script engine" is for. I tried running macros, making glossaries, external commands, inserting tables... But the execution never enters that function.

Any explanation about what that class or method is for would be really welcome @sunderme @benibela @timhoffm

benibela:
It is for script macros. E.g. in a macro the user could write  `buildManager.runCommand("txs:///latex")`

Issue #572, opened 2019-04-19
Avoid flicker when pdf is reloaded


a-reisberg:
Whenever the pdf viewer reloads the pdf file (say, after a compilation), the screen flickers, which is quite distracting. So I've been playing with the code to try to circumvent this problem. I found something interesting but I don't quite understand. Probably you could give me some pointers.

In `PDFDocument.cpp`, if I comment out the following two lines
```C++
reloadPage();
windowResized();
```
then obviously, the pdf file doesn't get rendered anymore. But I can get it to load initially by scrolling up and down (my test pdf file has multiple pages). The interesting thing is that, now, suppose I recompile the file, and then double click on the pdf viewer to briefly activate zoom, then the page gets reloaded without and flicker. The content gets updated without any side effect.

So it seems like when content is reloaded, the viewer gets white-painted, then some processing is done before actual content is painted. This white-painting doesn't happen when magnifying glass is used. But I cannot pin down precisely how this happens to make the rendering part of content reloading sequence more like that of the magnifying glass tool. Your help would be really appreciated!

Thanks!!!

a-reisberg:
So after digging a bit more, I've narrowed down the issue a quite a bit. Essentially, within the reload sequence, whenever `update()` is called, the view flickers. I tried the following, which seemed to work most of the time (exactly when it works is written below):

- Comment out `windowResized();` inside `PDFWidget::setDocument` while keeping with `reloadPage`, just to make things simple.
- Inside `reloadPage`, comment out `update()` and replace it with the following
```C++
 if (!magnifier) magnifier = new PDFMagnifier(this, dpi);
 magnifier->show();
 magnifier->hide();
```

Then, provided that `magnifier` has been actually used before (by the user), and provided that only **one** pdf file is being displayed, the reload is completely seamless! If more than one pages are being displayed, some of the pages flicker some don't. This is quite puzzling.

Any insights would be really appreciated!

a-reisberg:
So I finally figured out how to make it not flicker anymore. Essentially, instead of running `update()` at file reload, just add and remove a dummy `QWidget` to `PDFWidget`. This will also trigger a repaint from Qt, but somehow, Qt handles this much more gracefully. The reload now is completely smooth!

The code is here: a-reisberg/texstudio@53d46e8702d90384420b716226db4d8ee769a556

Is there a better way to do this? If not, should I make a pull-request?

Thanks!

sunderme:
[...]

Issue #550, opened 2019-04-02
Automatically compiling a document, i.e. while typing


muelli:
### Environment
- TeXstudio: TeXstudio 2.12.14 (git n/a)
Using Qt Version 5.12.2, compiled with Qt 5.12.0 R
via flatpak
- Qt: 
- OS: 
- TeX distribution: 

### Expected behavior

I think I would like to see the whole document getting compiled while I type, overleaf style.
I know I can have this with almost trivially by calling smth like `latexmk -pvc` but maybe TeXStudio can even be a bit better than that.

I am aware of the partial update function, i.e. Alt+P but it doesn't work well for larger paragraphs or tables. Currently I press F5 or F7, whatever I can hit more quickly, but it's a bit awkward, because I have to lift my hand for pressing those buttons.  I like the way overleaf does it and I am well aware that it's a rather primitive approach.

sunderme:
not a direction I want to go
unless there are some pull-requests, this will not happen

by the way, with more experience, less compilations are needed

manudroid19:
I see two problems with this: 

-First, the mouse icon changes to "loading" (spinning circle or whatever) whenever you clock compile, compile & see or any compilation order. It would be reaaally great if we could disable this in settings so hitting compile doesn't stop the workflow. 
-Second, we can use the latex -pvc but if we run it outside the program.
If we put the -pvc in TexStudio latexmmk command, the spinning mouse will be there forever and pdf won't update, no matter if we save the file. 

Note that my first comment about the spining mouse icon when compiling is a general improvement that I would love to see implemented, independently from the subject of this issue thread.

Keep up the good work!!!

benibela:
implemented in 448af99890e736963edf8b6042617b8b61055d4c

>-First, the mouse icon changes to "loading" (spinning circle or whatever) whenever you clock compile

There was no spinning mouse here on linux. perhaps the cat ate it



> If we put the -pvc in TexStudio latexmmk command, the spinning mouse will be there forever and pdf won't update, no matter if we save the file.

It only waits for latex commands. If you mark latexmk as non latex command, it should work

benibela:
>-First, the mouse icon changes to "loading" (spinning circle or whatever) whenever you clock compile, compile & see or any compilation order. It would be reaaally great if we could disable this in settings so hitting compile doesn't stop the workflow.

Actually, the spinning mouse (Qt::WaitCursor) it not shown during compilation, but when reloading the pdf

PR #577, opened 2019-04-22
Reduced flicker on pdf reload and slightly improved painting performance

a-reisberg:
See #572

### DIFF
--- a/src/pdfviewer/pdfrendermanager.cpp
+++ b/src/pdfviewer/pdfrendermanager.cpp
@@ -244,7 +244,7 @@ QSharedPointer<Poppler::Document> PDFRenderManager::loadDocument(const QString &
 	return document;
 }
 
-QPixmap PDFRenderManager::renderToImage(int pageNr, QObject *obj, const char *rec, double xres, double yres, int x, int y, int w, int h, bool cache, bool priority, Poppler::Page::Rotation rotate)
+QPixmap PDFRenderManager::renderToImage(int pageNr, QObject *obj, const char *rec, double xres, double yres, int x, int y, int w, int h, bool cache, bool priority, bool delayMode, Poppler::Page::Rotation rotate)
 {
 	if (document.isNull()) return QPixmap();
 	if (pageNr < 0 || pageNr >= cachedNumPages) return QPixmap();
@@ -284,6 +284,10 @@ QPixmap PDFRenderManager::renderToImage(int pageNr, QObject *obj, const char *re
 	if (img.isNull() && renderedPages.contains(pageNr)) { // try cache first
 		img = *renderedPages[pageNr];
 	}
+
+    if (!img.isNull() && delayMode)
+        QMetaObject::invokeMethod(info.obj, info.slot, Q_ARG(QPixmap, img), Q_ARG(int, pageNr));
+
 	//if(img.isNull()) // not cached, thumbnail present ? (fix crash?)
 	//	img=QPixmap::fromImage(page->thumbnail());
 	if (!img.isNull() && !partialImage) { // if a image was found, scale it apropriately
--- a/src/pdfviewer/pdfrendermanager.cpp
+++ b/src/pdfviewer/pdfrendermanager.cpp
@@ -244,7 +244,7 @@ QSharedPointer<Poppler::Document> PDFRenderManager::loadDocument(const QString &
 	return document;
 }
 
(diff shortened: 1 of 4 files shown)


sunderme:
I still think that

'#if QT_VERSION >= 0x050000
￼	        else {
￼	            QTimer::singleShot(delayTimeout, obj, [=]() {QMetaObject::invokeMethod(info.obj, info.slot, Q_ARG(QPixmap, img), Q_ARG(int, pageNr));});
￼	        }
￼	#endif
￼	    }'

does not do what it is supposed to do. Please take it out.

a-reisberg:
Could you please elaborate? I tested (note the `pgrNr + 100` below)
```C++
if (delayTimeout >= 0) {
    if (!img.isNull())
        QMetaObject::invokeMethod(info.obj, info.slot, Q_ARG(QPixmap, img), Q_ARG(int, pageNr + 100));
#if QT_VERSION >= 0x050000
    else {
        QTimer::singleShot(delayTimeout, obj, [=]() {QMetaObject::invokeMethod(info.obj, info.slot, Q_ARG(QPixmap, img), Q_ARG(int, pageNr));});
    }
#endif
}
and then `qDebug() << pageNr` inside `PdfWidget::setImage` and got the correct value.
```

benibela:
You can just create a new slot (any method can be marked as slot when the class is derived from QObject ), then it also works on qt4

And use tabs instead of spaces

Once it does not flicker anymore, it would also be easier to implement live previewing (e.g. for #550)

sunderme:
how is it supposed to work ?
you call setImage after delayTimeout, with no image yet calculated. There is another singleShot timer which call update which draws the background if the image is not present. So I don't see what this call is supposed to do and the syntax is odd, even considering the new Qt5 syntax (inokeMethod is calling a slot but you already are using a method (singleShot) which can directly address slots, so it is the same thing done twice)
Even it might work for now, no guarantees that it works on all qt versions properly.

Please, just remove that part of the code.

a-reisberg:
If you look at how `setImage` is written:
```C++
void PDFWidget::setImage(QPixmap, int pageNr)
{
    qDebug() << pageNr;
	forceUpdate = true;
    update(pageRect(pageNr));
}
```
it doesn't even use the `img` argument (in fact, before it didn't even use the `pageNr` argument). It's just used as a signaling device all along. So in the case where `img` hasn't been computed yet, `setImage` will be called with a null `img`, which is fine, since it doesn't use it at all. `setImage` will then call `update()`, which will call `renderToImage` again to actually render that page.

Now, that's the path for qt5. For qt4, those lines won't even be compiled. So for `qt4` and below, a less refined path is used inside `delayedUpdate()` itself, by having a timer which calls `update()`:
```C++
// Update eventually if loading takes too long. For Qt5 and above, lambda expression for SLOT has been used to handle
// this within renderToImage itself.
#if (QT_VERSION < 0x050000)
    QTimer::singleShot(1000, this, SLOT(update()));
#endif
```

Now, according to @benibela, it can also be made to work for qt4. @benibela Could you please elaborate? I don't know much about Qt. Thanks!

sunderme:
the problem with a slot is that you can't send parameters from singleshot without the qt5/lambda methodology ...

a-reisberg:
@sunderme I just had an idea of how to do it. Let me try...

sunderme:
by the way, the singleshot(...,functor) has only been introduced since qt 5.4 ...

a-reisberg:
Ok, now it should work with all Qt versions.

src/pdfviewer/pdfrendermanager.cpp
```diff
@@ -25,6 +25,20 @@ const int kMaxPageZoom = 1000000;
 const qreal kMaxDpiForFullPage = 1000.0;
 
 
+SetImageForwarder::SetImageForwarder(QObject *parent, QObject *obj, const char *rec, QPixmap img, int pageNr):
+    QObject(parent), obj(obj), rec(rec), img(img), pageNr(pageNr)
+{
+}
+
+void SetImageForwarder::setImage() {
+    QMetaObject::invokeMethod(obj, rec, Q_ARG(QPixmap, img), Q_ARG(int, pageNr + 100));
```
sunderme: where is the pageNr+100 coming from ?

a-reisberg: Ooops. Sorry. That's from debugging session. Will fix now.

PR #580, opened 2019-04-24
Add padding to small PDF files (size < 1024).

clerkma:
To fix [Specific PDF not viewable in TeXstudio](https://tex.stackexchange.com/questions/486175/specific-pdf-not-viewable-in-texstudio).
This is a bug related to poppler's `MemStream` class when load PDF with `Poppler::Document::loadFromData`.


sunderme:
thanks

PR #584, opened 2019-04-28
Added calculus and calculator package completion

ppizarror:


sunderme:
I would like to see an adaptation to cwl command definition, i.e. that commands that define other commands are marked with #d 
It is not possible to define more than one command per command in txs but at least the other commands should be defined properly

ppizarror:
Hi, I've used the classification format defined in manual, I hope i made it well.
Also, thanks for Texstudio, love it!!
Greetings from Chile 😄

sunderme:
thanks

PR #585, opened 2019-04-28
Disable compile & build buttons while executing compile proccess

manudroid19:
This pull request wants to make noticeable for users that a compile command while a proccess is executing will have no effect, by disabling the compile buttons. They should first stop the current process.

imagen

To achieve this, the buildManager instance needs to have a reference to the QAction buttons. I just created 2 new public attributes in the BuildManager class that are set up when the buttons are initialized in Texstudio instance. If there is any other prefered approach, using private attributes and a setter, for example, I have no problem with that.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1644,13 +1644,17 @@ bool BuildManager::waitForProcess(ProcessX *p)
 {
 	REQUIRE_RET(!processWaitedFor, false);
 	processWaitedFor = p;
-	m_stopBuildAction->setEnabled(true);
     QEventLoop loop; //This approach avoids spinlock and high CPU usage, and allows user interaction and UI responsivness while compiling.
     connect(p, SIGNAL(finishedProcess()), &loop, SLOT(quit()));
-    loop.exec(); //exec will delay execution until the signal has arrived
+	m_stopBuildAction->setEnabled(true);
+    m_buildAction->setEnabled(false);
+    m_compileAction->setEnabled(false);
+    loop.exec(); //exec will delay execution until the signal has arrived    
 	bool result = processWaitedFor;
     processWaitedFor = nullptr;
 	m_stopBuildAction->setEnabled(false);
+    m_buildAction->setEnabled(true);
+    m_compileAction->setEnabled(true);
 	return result;
 }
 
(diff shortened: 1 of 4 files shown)


timhoffm:
I would rather not couple the BuildManager to the GUI. The preferred way would be create appropriate signals `BuildManager::buildStarted` BuildManager::buildCompleted` and connect them to a slots `Texstudio::disableBuildButtons`, `Texstudio::enableBuildButtons`.

manudroid19:
> I would rather not couple the BuildManager to the GUI. The preferred way would be create appropriate signals `BuildManager::buildStarted` `BuildManager::buildCompleted` and connect them to a slots `Texstudio::disableBuildButtons`, `Texstudio::enableBuildButtons`.

Done. I created only one signal and slot using a boolean argument. I also refactored the "stop build" code, since it was already in BuildManager (where we don't want it to be) and I put everything inside appropiate ManagedActions in Texstudio class, just where all the QActions are. Now I think everything is more clear.

Two questions:
- Now the "stopcompile" QAction is created from Texstudio class instead of BuildManager, as it should be. In consequence, the call to `tr("Stop Compile")` is done from a different class now. Does this affect the translation engine? I don't really know how it works.
- In the waitForProcess function there is some code that I wanted to clean and I don't understand: 
https://github.com/texstudio-org/texstudio/blob/3ebd5ad1fdef7f897bf2d24133574c822856ffd2/src/buildmanager.cpp#L1643-L1655
What are the lines 1645, 1646 1651 and 1652 for?? I also don't understand why p, which is a reference to a ProcessX instance, gets casted to a QPointer<ProcessX> to check if the instance was destroyed... when nobody tried to destroy it... I just don't get why there is a return here.

sunderme:
@timhoffm any comments ? can we merge it ?

manudroid19:
> 
> 
> @timhoffm any comments ? can we merge it ?

I have answered the first question myself by teasing myself if the Build button was correctly translated. It works correctly (tested with spanish translation), so the seems like like the `tr()` Qt function works globally and it's not class dependent.

So as far as I'm concerned, this can be merged. 
An answer to the second question would be appreciated to further optimize the code in future pull requests.

manudroid19:
just to clarify: from my side, this can be merged. I have tested and translations work fine. 

The other question was to further optimize the previous code and has nothing to do with the mergeability of this commit.

timhoffm:
Sorry didn't have time to review and test this so far.

sunderme:
to get it further, I merge the PR. Anything more can be corrected directly in the repository

PR #587, opened 2019-05-03
minor fixes in biblatex.cwl

i3v:
In accordance with [biblatex package v3.12 manual](http://mirror.macomnet.net/pub/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf) :

1. Added missing `\defbibenvironment` macro
2. Fixed missing optional argument in `\defbibheading`


sunderme:
thanks

PR #634, opened 2019-06-05
Add an editor option to remove trailing whitespace on save

MeanSquaredError:
The editor class (QCodeEdit) supports an option that allows removal of trailing whitespace on file save, but this option is not accessible from the TexStudio configuration settings.

This proposed patch adds an extra editor setting called "Remove Trailing Whitespace on Save". It is a checkbox that allows the user to enable/disable removal of trailing whitespace from the saved text files.

Please note that unlike other editors the whitespace is removed only from the file saved on disk, the current text in the editor is not changed. This is the way it is implemented in QCodeEdit and I did not see the need to change that.

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -1555,6 +1555,49 @@ Then you can select a new shortcut by one of the following ways:
           <string>Editor</string>
          </property>
          <layout class="QGridLayout">
+          <item row="12" column="0">
+           <widget class="QCheckBox" name="checkBoxRealTimeCheck">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+            <property name="text">
+             <string>Inline Checking:</string>
+            </property>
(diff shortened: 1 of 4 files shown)


Dagefoerde:
Funny, this is the exact feature I was looking for today. I'd love to see this in TeXstudio. Thanks for providing a patch!

sunderme:
Have you tried that the saving without spaces works as expected ?

MeanSquaredError:
Yes, I am already using the latest TexStudio on Linux/Fedora Core 28 with this patch and it trims trailing spaces successfully. In order to use this feature you have to check the box "Remove trailing whitespace on save" in the Editor settings, because it is off by default.

Also it trims the trailing space in the saved file on disk only. The trailing space in the editor in not trimmed, but it is not a problem for version control because it uses the file on disk.

Did you notice any problems with this feature on any platform?

MeanSquaredError:
Actually I it is Linux/FC29, not FC28, but other than that I have been using it successfully for a couple of days and it works fine for me.

MeanSquaredError:
I can provide Linux RPMs for anyone who wants to test it.

sunderme:
thanks

Issue #643, opened 2019-06-19
Picture configure_commands.png in user manual is not contained in osx installation


muzimuzhi:
### Environment
- TeXstudio: 2.12.16
- Qt: 5.12.3, compiled with Qt 5.12.3 R
- OS: 10.14.5

Picture [`configure_commands.png`](https://github.com/texstudio-org/texstudio/blob/master/utilities/manual/configure_commands.png) is used by `usermanual_en.html`, and contained in the repo. Also, the latest commit that changes the picture is 2 years ago, and this picture file seems not special among other pictures in the same sub-directory of code repo.

But, this `configure_commands.png` is missing in osx installation (cannot find it in directory `/Applications/texstudio.app/Contents/Resources`).
https://github.com/texstudio-org/texstudio/blob/639c36d2701b821162b5c9415990736e5a3f5aaf/utilities/manual/usermanual_en.html#L121

muzimuzhi:
Closed in commit https://github.com/texstudio-org/texstudio/commit/d7e834cbaaa25ddfb584f65e1e089fe2d608a276.

PR #646, opened 2019-06-22
Add 2 missing images used by manual

muzimuzhi:
There are in total 3 images used by user manual but missing in package for macOS. Issue #643 (fixed by https://github.com/texstudio-org/texstudio/commit/d7e834cbaaa25ddfb584f65e1e089fe2d608a276) reports one of them, and this pr adds the rest 2.

Other change:
 - Substitute tabs by spaces in order to unify indent symbols

PS:
 - Is it necessary to keep items in arrays of file `texstudio.pro` in *strict alphabetic order*?


sunderme:
thanks

Issue #661, opened 2019-07-13
TeXstudio gets confused whith multiple log/PDF files


ysalmon:
### Environment
- TeXstudio: 2.12.16
- Qt: 5.9.5 R
- OS: Kubuntu 18.04
- TeX distribution: texlive 2019

### Actual behavior
TXS has a nice feature to specify additional path for searching log files. This is useful when one compiles with a specific output directory and generally this works well.

However, I am dealing with a collaborative project where we have a makefile that does not make use of outputdir. So depending on whether I run the makefile (to compile all documents and assemble them) or click on the Compile button in TXS (to preview the document I am working on), the log files does not get written at the same place.

The difficulty is that when a log file is present in the source directory, TXS considers it to be "the" log file, even when it is stale because I made changes to the tex file and compiled to my output dir. This leads me to not being able to spot errors I would introduce in the code, because the resulting PDF might still look globally sane, and the error is buried in the log file that TXS does not display or care about.

EDIT : think about it, the same problem probably arises when displaying the PDF file, which would explain why I do not see defects either in the preview.

### Proposed behavior
I understand this might be tricky, but I think that when the "additional path for log/PDF file"-feature is activated, TXS should look for log files in every place (including the default) and display the most recent log/PDF, instead of giving precedence to the file in the default location.

sunderme:
fixed with PR

PR #662, opened 2019-07-13
Use the most recent log

ysalmon:
This PR addresses my concern in #661.

This works satisfactorily for viewing the log file in the following sense : if I use my standard compile command from with TXS (which puts log file in ./_build), the new log file gets used (and errors reported) even if there is a stale log file in the base dir obtained from another compilation process.

I replicated the change for locating the PDF file but cannot extensively test it as my current routine does copy the PDF back from ./_build to the base dir and thus overwrites a stale PDF.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -2141,9 +2141,13 @@ QString BuildManager::findFile(const QString &defaultName, const QStringList &se
             }
         }
 	}
-	if (mostRecent && mr != nullptr)
-        return mr->absoluteFilePath();
-	return "";
+	if (mostRecent && mr != nullptr) {
+        QString result = mr->absoluteFilePath();
+        delete mr;
+        return result;
+    } else {
+        return "";
+    }
 }
 
 void BuildManager::removePreviewFiles(QString elem)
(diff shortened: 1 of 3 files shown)


src/buildmanager.cpp
```diff
@@ -2115,9 +2129,20 @@ QString BuildManager::findFile(const QString &defaultName, const QStringList &se
...
+                            delete mr;
+                        mr = new QFileInfo(fi);
+                    }
+                } else
+                    return fi.absoluteFilePath();
+            }
+        }
 	}
+	if (mostRecent && mr != nullptr)
+        return mr->absoluteFilePath();
```
sunderme: I guess that you have a resource leak here ...
and please use indentation, better braces, to connect which return is called when

ysalmon: Oh yes ; I ordinarily program in Python or OCaml, so manual memory management is not something I am used to !

It should be better with my most recent commit.

ysalmon: Indentation was inconsistent (tabs/spaces). I switched back to tab-indentation as was the case originally in that method (however other methods in the same file are space-indented…).

sunderme:
thanks

src/buildmanager.cpp
benibela: C++ has  automatic memory management, too. Pointer variables just needs to be declared with `unique_ptr<QFileInfo> mr` rather than `QFileInfo* mr ` and assigned with the constructor/reset rather than `=`. Or here an empty QFileInfo  could have been used rather than a pointer and checked if it exists rather than if it is null

PR #667, opened 2019-07-16
cwl: booktabs - fix \specialrule

muzimuzhi:
According to doc of `booktabs` package, sec. 4, change the first argument of `\specialrule` from optional to mandatory.
image


sunderme:
thanks

PR #670, opened 2019-07-19
Fix BUILD.sh for Fedora

MeanSquaredError:
Hi,
On recent Fedora versions (e.g. Fedora 29) the qmake executable is called qmake-qt5, so BUILD.sh fails. The proposed changeset fixes detection of QTDIR and QMAKE variables for Fedora systems. Additionally it allows the user to specify explicitly the QMAKE value. By default it assumes the old behavior, i.e. QTDIR="" and QMAKE="qmake"

### DIFF
--- a/BUILD.sh
+++ b/BUILD.sh
@@ -2,20 +2,6 @@
 echo "TeXstudio compilation :"
 echo "----------------------------------------"
 
-#autodetection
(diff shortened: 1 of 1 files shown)


sunderme:
i assume that it works.

PR #687, opened 2019-08-11
Some cwl changes: new files, keyval fixing, and several updates

muzimuzhi:
Main changes
 - add new cwl files for `keyval`, `blindtext`, and `mwe` packages
 - fix missing `=` in keyval list `key#true,false`, <BR>this influences cwl files for `datatool`, `graphicx`, `hyperref`, `listings`, `mathtools`, `siunitx`, and `yathesis`
 - update cwl files for `float`, `graphicx`, `lipsum`, and `pgfplots` to their latest version, including minor fixing

I can split this pr into three if the maintainers think the topics covered is a bit dispersive.


sunderme:
Leaving away = in case of true/false is intentional because keyval=true is the same like just "keyval" (which I prefer)...

muzimuzhi:
Using the following test, I confirmed that cwl lines like `key=#true,false` can still help user type in `key=true` and `key=false`. 
 * I will revert the related changes. 
 * Also, `tcolorbox.cwl` contains 1000+ `key=#true,false` lines. I will change them to `key#true,false` form.

------

Test (use `|` to denote the position of cursor):
 - load `graphicx`, then input `\includegraphics[|]{}`
 - hit "ctrl + space" and the auto-completion window pops up
 - select key `draft`, hit "enter" to get `\includegraphics[draft|]{}`
 - type in `=`, then hit `ctrl + space` again, and the auto-completion window containing options "true" and "false" pops up

muzimuzhi:
Other two things:
 - `tcolorbox.cwl` is a 40k+ line file, I will check if we can reduce its size later (not in this pr of course).
 - It is better to document the special use and effect (as described in the above "test") of `key#true,false` in `usermanual_en.html`

sunderme:
Thanks

Issue #690, opened 2019-08-15
Compiling of custom commands gets stuck


dirkziegenbalg:
### Environment
- TeXstudio: 2.12.16 (git 2.12.16-52-ga7fd0a3) starting with 2.12.16
- Qt: 5.12.1
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior

custom command should close an eventually open PDF in Adobe Acrobat and start compiling afterwards

### Actual behavior

Command gets stuck after sending the DDE command to Acrobat and the compiling interface (buttons, etc.) is deactivated. The compilation command is not executed. Calling the custom commands separately doesn't cause any issues.  It's not possible to start any compilation until Texstudio is restarted.

The command runs well with Texstudio 2.12.14. Issues started with version 2.12.16 and is still present with the latest git version.

### How to reproduce
Use the following custom build command:

close-adobe:Close ----> dde:///"C:\Program Files (x86)\Adobe\Acrobat DC\Acrobat\Acrobat.exe":acroviewA19/control/[DocClose("?am.pdf")][FileClose("?am.pdf")]"

create:create_with_close ---->   txs:///close-adobe | txs:///pdftex_user

pdftex_user:create_pdf ----> pdflatex.exe -shell-escape -synctex=1 -interaction=nonstopmode %.tex

MeanSquaredError:
Did you build it on Windows yourself or did you use a custom build?
I would say that the dde commands are the prime suspects, although I do not have a TexStudio Windows build for testing.

You can try replacing pdflatex.exe with a no-op (empty) batch file. E.g. something like this 
pdftex_user:create_pdf ----> noop.bat, where noop.bat is an empty file. And then see if the UI lockup still persists. If the lockup persists then DDE is the suspect.

dirkziegenbalg:
I downloaded the file from bintray (build from 11.8.19) but the issue 
also exists in the official 2.12.16 version.

Just checked again with the build from 11.8.19 and the issue persists 
with the noop.bat. TXS UI is till locked up and the batch file is not 
executed.

When I reinstall version 2.12.14 every runs fine as expected. The batch 
file as well as pdflatex are executed.


Am 16.08.2019 um 12:43 schrieb MeanSquaredError:

MeanSquaredError:
Hi,
Does the UI lockup happen with TexStudio 2.12.16 if you replace the command in close-adobe:Close with an empty noop.bat? That is 

close-adobe:Close ----> noop.bat

[...]

PR #695, opened 2019-08-19
Fix a lockup in BuildManager::waitForProcess

MeanSquaredError:
This pull request fixes the issue reported in https://github.com/texstudio-org/texstudio/issues/690
The bug is caused by a race condition in BuildManager::waitForProcess. If that method is given a process that has already ended, then the method will wait forever.


sunderme:
thanks

dirkziegenbalg:
thanks

PR #696, opened 2019-08-23
Disable all GCC optimizations when building a debug binary

MeanSquaredError:
Currently BUILD.sh and texstudio.pro enable GCC optimizations (-O2) when building a debug binary. This complicates debugging of texstudio because many local variables are optimized out.
The proposed pull request disables all optimizations for debug builds.


sunderme:
Thanks

PR #708, opened 2019-08-29
Fix a crash when undoing/redoing code with labels and other structure elements

MeanSquaredError:
There is an easily reproducible crash that occurs when a user uses multiple undo/redo on code that contains labels or other structure elements.

How to reproduce the crash:

1. Create a simple tex file with only two lines:
\label{abc}
\label{def}

2. In the structure view panel on the left open tree with the labels. Right click on the "def" label (the second label), select "Find usages", then in the "Replace by" box enter "def2" (any different name will do), then click the [Replace all] button.

3. In the structure view panel on the left right click on the "abc" label (the first label), select "Find usages", then in the "Replace by" box enter "abc2" (any different name will do), then click the [Replace all] button.

4. Then in the main text window click the undo (CTRL-Z) and redo (CTRL-SHIFT-Z) shortcuts in the following order.

CTRL-Z (undo)
CTRL-Z (undo)
CTRL-SHIFT-Z (redo)
CTRL-SHIFT-Z (redo)
CTRL-Z (undo)
CTRL-Z (undo)
CTRL-SHIFT-Z (redo)
CTRL-SHIFT-Z (redo)
CTRL-Z (undo)
CTRL-Z (undo)
CTRL-SHIFT-Z (redo)
CTRL-SHIFT-Z (redo)

It usually takes 2-3 undo-redo loops until TexStudio crashes.

The reason for the crash is that the structure view panel on the left uses a QTreeView control which implements QItemSelectionModel that keeps the current selection in persistent indexes (QPersistentModelIndex)
These persistent indexes should be updated when the code in LatexDocument::patchStructure adds and removes rows. However patchStructure does not send update notifications for most changes (it only notifies for category addition/removal).
So once the persistent indexes go out of sync with the actual entries in the tree view and sooner or later it tries to reference an already freed StructureEntry item (the data item for QPersistentModelIndex). Such attempt to work with an already freed StructureEntry data causes a crash.

The only reasonable solution is to keep any persistent indexes (QPersistentModelIndex) in sync with the QTreeView items which means that we must notify the control of any changes to the data items either explicitly through changePersistentIndex/changePersistentIndexList or implicitly by calling the member functions beginInsertRows/beginMoveRows/beginRemoveRows/endInsertRows/endMoveRows/endRemoveRows.

The proposed patch uses the second approach since the code already has most of the infrastructure for that it arguably it is the proper solution anyway because these member functions will call take care of the calculations that we would have to do if we were to use changePersistentIndex/changePersistentIndexList.

The updated code seems to work correctly for me and it does not crash. However it is a significant change so maybe it needs more thorough testing.

### DIFF
--- a/src/latexdocument.cpp
+++ b/src/latexdocument.cpp
@@ -2194,30 +2194,57 @@ void LatexStructureMergerMerge::mergeChildren(StructureEntry *se, int start){
 		mergeStructure(oldChildren[i]);
 }
 
+bool LatexDocument::IsInTree (StructureEntry *se)
+{
+	Q_ASSERT(se);
+	while (se) {
+		if (se->type == StructureEntry::SE_DOCUMENT_ROOT) {
+			return true;
+		}
+		se = se->parent;
+	}
+	return false;
+}
+
 void LatexDocument::removeElementWithSignal(StructureEntry *se)
 {
+	int sendSignal = IsInTree(se);
 	int parentRow = se->getRealParentRow();
 	REQUIRE(parentRow >= 0);
-	emit removeElement(se, parentRow);
+	if (sendSignal) {
+		emit removeElement(se, parentRow);
+	}
 	se->parent->children.removeAt(parentRow);
 	se->parent = nullptr;
-	emit removeElementFinished();
+	if (sendSignal) {
+		emit removeElementFinished();
+	}
 }
 
 void LatexDocument::addElementWithSignal(StructureEntry *parent, StructureEntry *se)
 {
-	emit addElement(parent, parent->children.size());
+	int sendSignal = IsInTree(parent);
+	if (sendSignal) {
+		emit addElement(parent, parent->children.size());
+	}
 	parent->children.append(se);
 	se->parent = parent;
-	emit addElementFinished();
+	if (sendSignal) {
+		emit addElementFinished();
+	}
 }
 
 void LatexDocument::insertElementWithSignal(StructureEntry *parent, int pos, StructureEntry *se)
 {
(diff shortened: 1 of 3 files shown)


MeanSquaredError:
The crash happens in Linux with Qt 5.11.3. TexStudio is the latest source downloaded from github and built from source.

sunderme:
This will take a little more time to review, I take a look during the week-end.

MeanSquaredError:
Sure. It is a non-trivial change so it takes some time to verify it.

I was looking for a simpler fix, but after trying a few simpler hacks, none of them worked reliably. The problem is that QTreeView/QItemSelectionModel relies on keeping valid persistent indexes to the structure view elements (rows of LatexDocumentsModel), so one has to maintain these persistent indexes somehow.

Maybe another solution would be to clear the selection after each change, but it is more of a hack and we wouldn't have a guarantee that it would always work.

Another part of the problem is that LatexDocument::patchStructure is quite big and difficult to maintain and modify. I guess it might be a good idea to refactor it and split it into smaller functions. But then agian that is a different issue.

MeanSquaredError:
Actually memory corruption starts quite early on during the undo/redo cycle. The two renames and the first two CTRL-Z (undos) are sufficient for the persistent indexes to get corrupted and to start using invalid StructureEntry pointers. In my case running a debug build of TexStudio under gdb or valgrind makes the debugger the second undo is sufficient to make the debugger complain about access to freed memory. The release build takes several of the double undo/redo cycles to crash.

sunderme:
test compile looks okay. I will merge it and then we can see if something unexpected comes up

PR #713, opened 2019-09-02
minor changes to declare-catch regex

muzimuzhi:
Main changes:
 - add support for `\edef`, `\gdef`, and `\xdef`, whose syntax is identical to `\def`'s
 - allow empty default value for optional argument, e.g., `\newcommand\abc[1][]{code}`
 - allow optional star right after `\newenvironment`, see [related doc](https://latexref.xyz/_005cnewenvironment-_0026-_005crenewenvironment.html#g_t_005cnewenvironment-_0026-_005crenewenvironment)


sunderme:
thanks

PR #714, opened 2019-09-04
Allow the user to optionally skip the installation after building

MeanSquaredError:
In some cases when debugging TexStudio it makes sense to build it without a following installation, for example if there is already an installed stable version of TexStudio and the newly build version will be run from the source directory.

The proposed patch adds a new option to BUILD.sh that allows the user to optionally skip the installation after building.


sunderme:
I don‘t see the necessity , but I don‘t use build.sh either...

thanks

PR #723, opened 2019-09-13
Add CWL files for the MinionPro font

MeanSquaredError:
Add CWL files for the MinionPro font. This works around https://github.com/texstudio-org/texstudio/issues/716 when TXS crashes with memory allocation error while generating the .cwl files. Most likely the problem happens while it tries to generate MinionPro-FontDef.cwl

It looks like the users computer has some kind of memory allocation limit applied to TXS, but we were unable to find out how exactly it is applied. Alternatively it could also be some kind of memory corruption bug but I am unable to fully reproduce the issue on my computer and Valgrind does not complain about any memory corruption while TXS parses the MinionPro .sty files.


muzimuzhi:
More minor improvement can be done. I will prepare a cleaner cwl (group) for MinionPro in at most 2 days.

sunderme:
after looking into documentation, I would suggest to generate MnSymbols.cwl (which is the math symbols) and maybe MinionPro.cwl, just to load MnSymbol.

Is there going to be an update here ?

MeanSquaredError:
Do you mean to create a stub MinionPro.cwl that would contain a single command including MnSymbols.cwl ?

Regarding the update - I am not sure. Maybe @muzimuzhi can comment on the update since he suggested it. If he has an update, he can submit a new pull request and then I will close this one. Alternatively he can send me the updated files and I will update this pull request.

sunderme:
Basically,yes

MeanSquaredError:
OK, I will prepare an update that uses MnSymbols.cwl

MeanSquaredError:
I replaced the previous MinionPro CWL files with other two files as per your suggestion:

`MnSymbol.cwl` this is the autogenerated CWL file for the MnSymbol package.
`MinionPro.cwl` this is a stub CWL file which includes MnSymbol.cwl

sunderme:
nearly there.
I would remove the #S in MnSymbol.cwl as the commands are really used for completion, as they valid math symbols.

MeanSquaredError:
Updated again to remove the S classifier from the commands in MnSymbol.cwl

sunderme:
Thanks

sunderme:
Do you want to put your name into the about dialog ?

MeanSquaredError:
Sure, please add "Vesselin Atanasov" to the list.

Issue #712, opened 2019-09-02
Feature Request: support xparse-style declarations when auto-generating cwl


muzimuzhi:
Currently, only latex2e-style declarations like `\newcommand` and `\newenvironment` are analysed in the process of auto-generating cwl files.

Nowadays, more and more new packages are (re)written using new declaration commands provided by `xparse` pacakge (which is based on latex3). Before the related cwl files are manually added into txs, maybe it is better to support `xparse`-style declarations at code level first.

A bnf-like description of these new declaration could be [found here](https://github.com/muzimuzhi/personal-notes/blob/master/sub/latex-xparse.md#syntax)), and the first part is as follow
```abnf
xparse-declare ::= cmd-declare <cmd name> arg-spec <definition code>
                 | env-declare <env name> arg-spec <begin code> <end code>

cmd-declare    ::= "\\" prefix ["Expandable"] "DocumentCommand"
env-declare    ::= "\\" prefix "DocumentEnvironment"
prefix         ::= "New" | "Renew" | "Provide" | "Declare"

arg-spec       ::= argument*
argument       ::= processor* [modifier] arg-type
```
Parsing `<cmd namd>` is not a problem, but counting the exact number of arguments may take some effort.

PS: I wonder if regex can parse `> {\SplitList{,}} m` to `m`, since it seems need a stack (to skip paired braces).

muzimuzhi:
This relates function `parseLine` in file `latexstyleparser.cpp`.

sunderme:
do you have an example package which uses xparse as a method t test against ?

muzimuzhi:
For example, packages `menukeys`, `lipsum`, and `siunitx` (ordered by length of source code, from shorter to longer). Some, but not all of the packages using `xpares` can be found in https://ctan.org/topic/expl3.

------

If counting the number of arguments is not considered, the regex of `\NewDocumentCommand`-like and `\NewDocumentEnvironment`-like ones can be
```cpp
// for convenience, "\" is not escaped
QRegExp reXparseCom("\\(?:New|Provide|Declare)(?:Expandable)?DocumentCommand\s*\{\s*(\\\w+)\s*\}");
QRegExp reXparseComNoBrace("\\(?:New|Provide|Declare)(?:Expandable)?DocumentCommand\s*(\\\w+)");
QRegExp reXparseEnv("\\(?:New|Provide|Declare)DocumentEnvironment\s*\{\w+\}")
```
[...]

PR #730, opened 2019-09-25
Generate .cwl commands based on xparse definitions

MeanSquaredError:
@muzimuzhi @sunderme

This patch implements the feature request #712
by adding support for parsing xparse definitions and generaing .cwl definitions for the corresponding commands and environments.

The implementation has a couple of limitations:

A) When an xparse command is used to define a command or environment the name of the command/environment and the argument definition must be on the same line.

So this definition will be recognized

`\NewDocumentCommand{\foo}{m o}`
`{#1}`

while this definition will not be recognized

`\NewDocumentCommand{\foo}`
`{m o}{#1}`

All of the packages that I checked use the first type of definition so it shouldn't be a real problem in most cases. However the xparse documentation gives an example where the the argument definition is on the next line, so such a definition would not be recognized.
Adding support for definitions (where the arguments definition group is on the next line) would require a major rewrite of the parser.

B) The following xparse argument types are recognized:
`m o O s r R d D` are handled (sort of) properly (the .cwl language is not flexible enough to handle all the edge cases but these should work more or less).
`! +` (modifiers) are ignored (probably the right thing to do anyway).

`> {...}` (processing filters) are ignored (this is probably the right thing to do anyway).
`v b e E` are not supported and ignored. I guess I could add support for these if there are real-life packages that use them (and if I am able to figure out what these argument types actually do).
`l u g G` are obsolete and not supported by the parser implementation. Again I could add support for them if there are any real-life packages that use these argument types.

Implementing xparse required some changes to the parser so it probably needs extensive testing. I tried it with a few sample files and it seems to work fine, but more testing is required.
This is a small sample class file that I used while testing the xparse cwl support:

`\NeedsTeXFormat{LaTeX2e}`
`\ProvidesClass{myclass}[2019/09/25 My test class]`
`\LoadClass[letterpaper,oneside]{amsart}`
`\NewDocumentEnvironment { twice }{ +b O{\ttfamily} rAB }`
`{#2#1#2} {}`
`\NewDocumentCommand{\foo}{dAB s}{#1}`

The xparse environment and command are meaningless but they are simple and you can test the parser by making small modifications to the command and environment argument definitions.


sunderme:
Have you checked that the parsing of normal style files does not fail ?

MeanSquaredError:
The updated parser code generates correctly the .cwl files for the amsart class and the related packages that amsart requires. It also works correctly for a custom class based on the amsart class.

sunderme:
Thanks

Issue #728, opened 2019-09-25
How do I prohibit biber runs when storing the bbl file in a subdirectory?


bushfeuer:
To keep my working folder clean, I configured pdflatex to store the aux files in a subdirectory with `pdflatex.exe -aux-directory=aux`.
I also configured biber to store its output to the same folder:
`biber.exe --output_directory aux`
I then added this folder to the additional search path in the "build" menu of Texstudio by adding `.\aux`.

### Environment
- TeXstudio: 2.12.16
- Qt: 5.12.1
- OS: Windows 10 Pro x64
- TeX distribution: MikTeX

### Expected behavior
Texstudio doesn't run pdflatex and biber multiple times every time I use the "Build&View" or "Compile" commands.

### Actual behavior

Once the bbl file is not stored in the same directory as the tex files, Texstudio uses biber every time I use the "Build&View" or "Compile" commands. This is not the case when the bbl file is in the directory of the tex files. 

Is there any way to prevent the automatic execution of biber when the bbl file is in a different directory? The procedure takes quite some time compared to a single run of pdflatex...

MeanSquaredError:
I have never used biber but looking at the source code at
https://github.com/plk/biber/blob/dev/bin/biber
more specifically line 25, GetOptions(...)
it looks like the output-directory should be specified with an extra equal sign after the option name, i.e.
`--output-directory=path\to\your\directory`
Not sure if the Perl parser really needs the equal sign or it can use space instead of it.
Also you may try specifying the absolute pathname of your output directory, so the full command line could look like this:
`biber.exe --output_directory "C:\Users\MyUser\MyDir\aux\\"`
(note the double backslash ( \\ ) at the end of the path in order to prevent the backslash from escaping the double quote ".

Overall, you can try running biber from Windows command prompt and once you are sure that it works correctly and outputs the file in the correct directory, specify it as a command in TexStudio.

MeanSquaredError:
Sorry, ignore my previous message, it seems that the actual problem is with TexStudio and not with Biber

MeanSquaredError:
I will try preparing a patch for that, but I have a few other patches in the queue, so it may be a week or so before it is ready.

sunderme:
How does pdflatex find the bbl/blg files ?

MeanSquaredError:
[...]

PR #735, opened 2019-10-01
When looking for .bbl files also check in the additional PDF search paths

MeanSquaredError:
Right now TXS decides whether to recompile the .bib files by comparing the timestamps of the .bib files with the timestamp of the output .bbl file. If the .bbl file is created in a directory other than the input .tex file, TXS is unable to find the output .bib file and runs the .bib compiler each time when the .tex document is compiled.

This slows down recompilation of the .tex file as reported in https://github.com/texstudio-org/texstudio/issues/728
The proposed patch works around this issue by also looking in the additional PDF search directories when searching for the compiled .bbl file.

On a side note:

1. We probably need support for a directory containing all kinds of temporary output files (bibliography, synctex, etc.). By default it should be the directory where the .tex and/or .pdf files are located but it could be a separate directory. I don't have a clear idea how this should be organized so I am mentioning it in case someone wants to discuss this feature.

2. The different findFile, findFileInPath, findAbsoluteFilePath seem to need overhaul and/or unification. The different logfile methods could move some of their code into the overhauled search functions too.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -5916,12 +5916,14 @@ void Texstudio::runBibliographyIfNecessary(const QFileInfo &mainFile)
 
 	QList<LatexDocument *> docs = rootDoc->getListOfDocs();
 	QSet<QString> bibFiles;
-	foreach (const LatexDocument *doc, docs)
-		foreach (const FileNamePair &bf, doc->mentionedBibTeXFiles())
+	foreach (const LatexDocument *doc, docs) {
+		foreach (const FileNamePair &bf, doc->mentionedBibTeXFiles()) {
 			bibFiles.insert(bf.absolute);
-    if(bibFiles.isEmpty()){
-        return; // don't try to compile bibtex files if there none
-    }
+		}
+	}
+	if(bibFiles.isEmpty()) {
+		return; // don't try to compile bibtex files if there none
+	}
 	if (bibFiles == rootDoc->lastCompiledBibTeXFiles) {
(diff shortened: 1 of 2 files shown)


MeanSquaredError:
Actually thinking about it I am not sure if the additional PDF search paths is the correct setting that should be used.

We could use the additional log search paths (usually log files are placed in the temporary output directory).

Or we could have a separate setting, although it is probably not worth to have a separate setting just for the bibliography temp files.

So I am not really sure about the correct solution with the current organization of directories in TXS.

sunderme:
I would go for the log directory as they are probably moved there as well whereas the pdf is not.

MeanSquaredError:
Sounds right, given that  -aux-directory in miktex/pdflatex seems to change the output directory for the intermediate files only and not for the final .pdf.

I will update the patch to use the logfile directory instead.

MeanSquaredError:
I updated the code so now it searches for the .bbl files in the .tex file directory and in the additional logfile directories. After some testing it works reliably for me.

On an unrelated note, both with the old bibliography update code (before this pull request) and my new code I am hitting now and then some kind of weird caching bug. It seems that even if you update the timestamp of the .bib file QFileInfo still reads the old timestamp. I suspect some kind of interaction between the linux filesystem and QFileInfo caching (or maybe a Qt bug), but I cannot reproduce it reliably. It happens very rarely and now it completely stopped happening. So if someone can reproduce it reliably let me know.

sunderme:
Thanks

PR #737, opened 2019-10-04
Revert "add zhlipsum.cwl"

OsbertWang:
Reverts texstudio-org/texstudio#736
There is one bug for the zhlipsum.cwl.
```TeX
#keyvals:\zhlipsum
name=#simp,trad,nanshanjing,xiangyu,zhufu,aspirin
before=
after=
inter=
#endkeyvals
```
These codes cannot give the autocomplete list in texstudio.


sunderme:
why don't you prepare a final cwl and then we can put in that one ?

muzimuzhi:
`zhlipsum` package takes special treatment of the possible optional arguments which `\zhlipsum` accepts. Currently, the cwl-analysing mechanism cannot handle this form of flexibility.

For example, the first optional argument of `\zhlipsum` can either be a number range or a key-value list. But these cwl lines
```latex
\zhlipsum[num]
\zhlipsum[keyvals]

#keyvals:\zhlipsum
key1
key2
#endkeyvals
```
won't work if the user types in `\zhlipsum[|]` (`|` represents the position cursor) and expect a completion pop-up showing list `key1, key2`.

------
A subset of the full-usage of `zhlipsum` package, like
```latex
\zhlipsum
\zhlipsum[num][keyvals]

#keyvals:\zhlipsum
...
#endkeyvals
```
could be a workaround.

OsbertWang:
> why don't you prepare a final cwl and then we can put in that one ?

I'm afraid that I cannot fix this bug in a short time, and I don't want this zhlipsum.cwl with defects is included. It's my fault to hand in a wrong file. Sorry.



> Even a subset of the full-usage of `zhlipsum` package, like
> 
> ```
> \zhlipsum
> \zhlipsum[num][keyvals]
> 
> #keyvals:\zhlipsum
> ...
> #endkeyvals
> ```
> 
> cannot make the completion function work.
> 
> Maybe some enhancement should take on [`latexpackage.cpp`](https://github.com/texstudio-org/texstudio/blob/master/src/latexpackage.cpp).

Well, `\zhlipsum[num][keyvals]` cannot but `\zhlipsum[keyvals]`can. However, `\zhlipsum[keyvals]` brings another bug that texstudio will treat number as wrong argument when I type number in it.

sunderme:
`#keyvals:\zhlipsum#c` disables the syntax checking...
ambiguities with two arguments can't be resolved, where one argument is something else as the first argument with 2 arguments.

OsbertWang:
> `#keyvals:\zhlipsum#c` disables the syntax checking...
> ambiguities with two arguments can't be resolved, where one argument is something else as the first argument with 2 arguments.

Yes. As muzimuzhi mentioned, maybe some enhancement should take on latexpackage.cpp.

sunderme:
as completion file for a non-essential package, the cwl is good enough

Issue #689, opened 2019-08-12
Feature Request: add special treatment for commands like "\newcount"


muzimuzhi:
As far as I guess, since cwl file only supports options inside braces (`{}`, `[]`, etc), commands that normally used without braces around their arguments and defines (mostly one of) their arguments as new macros are specially treated in `latexdocument.cpp`. For example, 
```latex
\def\newcmd<parameter text>{<def>} % and \edef, \gdef, \xdef
\newcount\mycount
```

However, there are other command that share the same syntax with `\newcount` but have not been specially treated. This will cause wrong highlight effect in txs, see
![image](https://user-images.githubusercontent.com/6376638/62860224-0fa50c00-bd32-11e9-9d36-ad6636919c37.png)
The full list of these `\newcount`-like commands is
```latex
\newdimen
\newskip
\newmuskip
\newbox
\newtoks
\newread
\newwrite
\newlanguage
\newfam
\newinsert
```

PS: These macros, plus `\newcount`, were originally introduced in plain tex format, and were then adopted by latex2e format. Hence one can see the occurrences of them both in documents about tex with plain tex (like [`texdoc texbytopic`](http://mirrors.ctan.org/info/texbytopic/TeXbyTopic.pdf), Chap. 31) and in documents about latex2e (like [`texdoc source2e`](http://mirrors.ctan.org/macros/latex/base/source2e.pdf), Sec. b.9).

The related source lines could be
https://github.com/texstudio-org/texstudio/blob/a7fd0a3dc064f685d68b8a6792bb825f4669a085/src/latexdocument.cpp#L838-L852

sunderme:
I have implemented it in the cwl, see commit 99cafd4f0df6c49319cdc13b9891fc7672b8cdff

muzimuzhi:
The added lines to cwl file only recognize usages with braces like 
```latex
\newbox{\somebox} % here txs recognizes \somebox as a new cmd
...
\somebox % here txs highlights \somebox as normal cmd
```

But what I want, and what is usually used in latex package source files is the form without braces:
```latex
\newbox\someBox % here txs highlights both \someBox as "unrecognized command"
...
\someBox
```
That is to say, commit https://github.com/texstudio-org/texstudio/commit/99cafd4f0df6c49319cdc13b9891fc7672b8cdff does not solve the question shown in the screenshot I have posted:
[...]

PR #742, opened 2019-10-06
Changes to cwl files and related section in manual

muzimuzhi:
Changes to cwl files:
 - move `\dashbox`, which is valid only inside `picture` latex environment, from `latex-dev.cwl` to `latex-document.cwl`, since other `picture`-only commands are all recorded in the latter cwl file.
 - remove redundant argument of arg-free commands (`\<xxx>name`)
 - set arg-name and classifier for `\(re)newenvironment`
 - add some `%l` to dimension related arguments
 - unify name of arguments
 - other minor changes

Changes to related section in manual:
 - supplement descriptions for `key` and `envname`, according to the related source lines:
https://github.com/texstudio-org/texstudio/blob/6ca046695f953300856c7b817eab5004c40d8460/src/latexpackage.cpp#L676-L677
 - add after-punctuation spaces


muzimuzhi:
> * set arg-name and classifier for `\(re)newenvironment`

After this change, the following lines of source code might be redundant:
https://github.com/texstudio-org/texstudio/blob/6ca046695f953300856c7b817eab5004c40d8460/src/latexdocument.cpp#L827-L858

This has the similar situation in commit https://github.com/texstudio-org/texstudio/commit/d538bab7c91e9b845e0be698d1aac49b7e9b2d77, which can be traced back to issue #689.

sunderme:
thanks

sunderme:
do you want to be mentioned in about dialog ?

muzimuzhi:
> do you want to be mentioned in about dialog ?

It is my honor. Either "muzimuzhi" or "Yukai Chou" is welcome.

PR #749, opened 2019-10-12
Fix Qt4 build

Timvde:
When refactoring for using QRegularExpression in Qt5, the wrong function
was called in Qt4.

The CI showed you broke the Qt4 build: https://travis-ci.org/texstudio-org/texstudio/jobs/597073296

NOTE: since I don't have Qt4 locally, I couldn't actually test and build this. Make sure to test that the CI passes before merging.


sunderme:
Thanks

PR #753, opened 2019-10-15
Fix gcc9 compile warnings

MeanSquaredError:
This is the first set of cosmetic patches that fix the issues mentioned in https://github.com/texstudio-org/texstudio/issues/748
This set fixes the GCC compiler warnings. The patches fix the following groups of GCC warnings:

1. -Wcast-function-type warnings when including QtScript headers. These are the compiler warnings reported in https://github.com/texstudio-org/texstudio/issues/748
Even though I initially thought that the warnings are caused by TexStudio code it turned out that the warnings are caused entirely by code in QtScript/qscriptengine.h
I reported the issue in the Qt bugtracking system at https://bugreports.qt.io/browse/QTBUG-79242 but until they fix it I have added a bit of code to TXS src/scriptengine.h to silence the warnings. Given that it is a low-priority issue it may take a while until it is fixed in the QtScript headers.

2. -Wdeprecated-copy warnings in the TXS code reported by GCC9. These have been fixed by adding explicit default constructors and operator= to some classes in TXS.

3. -Wdeprecated-copy warnings in the Qt headers. I have added a bit of code to texstudio.pro to silence the warnings when building with GCC and Qt version is between 5.0.0 and 5.12.x
These warnings have been fixed in Qt 5.13.x so we don't silence them with newer versions of Qt.

4. Warnings about two unused labels in src/pdfviewer/synctex/synctex_parser.c. Fixed by removing the unused labels.

5. Warning about unused variable in rc/pdfviewer/PDFDocks.cpp. Fixed by removing the unused variable.

6. -Wcatch-value warnings in several files. gcc9 complains when we catch polymorphic objects by value instead of reference. Fixed by catching the exception objects by reference.

7. -Wimplicit-fallthrough warnings in several files. Fixed by adding the magic comment "// Intentional fallthrough" in the corresponding case statements. gcc recognizes the magic comment and silences the corresponding warnings.

8. Warning about about implicit declaration of vasprintf in one source file. Fixed by defining _GNU_SOURCE just before including stdio.h.

9. For the debug build removed some -W flags which are already enabled by -Wall and -Wextra.

### DIFF
--- a/src/bidiextender.cpp
+++ b/src/bidiextender.cpp
@@ -33,7 +33,7 @@ HKL getCurrentLanguage()
 	try {
 		kb::XKeyboard xkb;
 		return xkb.get_group() + 1;
-	} catch (X11Exception) {
+	} catch (X11Exception &) {
 		return 0;
 	}
 
@@ -67,7 +67,7 @@ bool isProbablyLTRLanguageCode(HKL id)
 		id --;
         if (id >= 0 && id < static_cast<HKL>(installedLangSymbols.size()))
 			return isProbablyLTRLanguageRaw(installedLangSymbols[id]);
-	} catch (X11Exception) { }
+	} catch (X11Exception &) { }
 	return false;
 #else
 	return false;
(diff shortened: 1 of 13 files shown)


sunderme:
Thanks

Issue #754, opened 2019-10-16
External open not working with custom build dir


MaxSagebaum:
I have configured my environment to create everything in the 'build' directory with '-output-directory=build' for pdflatex. I also added this directory to the additional search path and there are no auxiliary or pdfs files left in the main directory. Everything works so far.

Only if I click on 'External Viewer' I get the error:

Process started: xdg-open "sample".pdf

This command misses the directory prefix. 
TexStudio displays the correct pdf file in the 'build' folder. But, the command behind the button does not use the full path.

MaxSagebaum:
I forgot:

### Environment
- TeXstudio: 2.12.16
- Qt: 5.12.4, compiled with Qt 5.12.1 R
- OS: Fedora 5.2.17-200.fc30.x86_64
- TeX distribution: texlive

MeanSquaredError:
See
https://tex.stackexchange.com/questions/272151/texstudio-change-external-pdf-viewer-command-to-open-pdf-from-sub-directory

More specifically see the second reply where they set the PDF file location using "?am.pdf"

MaxSagebaum:
Thank you for the reply. I missed this command and I could get it to work with:

```
xdg-open ?a)/build/?m.pdf > /dev/null
```

Nevertheless, I think the default behaviour of the button and the open command should be to open the pdf file found by TexStudio. It is not very user friendly to have to configure this in a manual fashion.

Maybe a new 'p' special char would help in this case. It would point to the pdf file found by TexStudio. The default command could then be:

```
xdg-open ?p > /dev/null
```

MeanSquaredError:
OK, let me try to prepare a patch

MeanSquaredError:
I added an implementation of this feature as a pull request here https://github.com/texstudio-org/texstudio/pull/761

However I am not sure if the suggested design (where ?p expands to the PDF pathname) is a correct one. See my thoughts on it on the page with the pull request.

gabriele-v:
Hello @MeanSquaredError, sorry to bring up this but I've tried to make the ?p working but without success.

If I've understood correctly, i have to change from `?am.pdf` to `?p{pdf}am.pdf` correct?

Thanks

MeanSquaredError:
@gabriele-v 

This is how I use it in the `External PDF Viewer` setting:
```
okular ?p{pdf}:ame"
```
In your setting you seem to be missing the colon symbol.

PR #761, opened 2019-10-23
Add the ?p command-line expansion that searches for compiled PDF file

MeanSquaredError:
This pull request implements the feature request from https://github.com/texstudio-org/texstudio/issues/754

Basically it adds the ?p command-line expansion which expands to absolute pathname of the PDF file. The file name of the PDF file is taken from the basename of the main .tex file with .pdf appended to it.
The searched directories are the directory that contains the main .tex file plus the additional PDF search directories.

I tested the patch and it seems to be working OK.

That being said, I am not sure that the selected design decision is correct. 

A) This pull request implements the ?p expansion for .pdf files, but TexStudio can also generate .ps or .dvi files (or any other image type for that matter). So while ?p solves the problem with .pdf files, it does not solve the problem with the other image types.
B) We can only search for files with the same basename as the main .tex file. What if someone compiles PDF files with some other name?

I can think of several possible workarounds that can be implemented in code:

1. Add the file extension in braces, e.g. ?p{pdf} or ?p{dvi} or ?p{ps} , etc. This solves the problem A (file extensions other than .pdf) but does not solve the problem with other filenames.

2. Use a new, third expansion syntax (besides % and ?). For example we could use expansions enclosed in { and } and inside them allow the following expansions:
%% is the % character
%b is the main file basename
%e is the main file extension
For example, let's say that we compile our tex files by adding the -compiled suffix to them and setting the extension .pdf. Then main.tex would be compiled to main-compiled.pdf

Then our command would look like this:

xdg-open {%b-compiled.pdf} > /dev/null

3. Maybe there is some other way to design the expansions so that they would allow solving problems A and B.

I am still unsure of the correct approach, so any comments, thoughts or suggestions are welcome.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -577,6 +577,29 @@ QStringList BuildManager::parseExtendedCommandLine(QString str, const QFileInfo 
 	return result;
 }
 
+/*
+ * Select a file which provides the pathname parts used by the "ame" expansions. Currently we can select
+ * one of the following files:
+ *
+ * - Master (root) .tex file (default).
+ * - Current .tex file. Selected by the c: prefix.
+ * - A file with the same complete basename as the master file and a chosen extension. The search for this
+ *   file is done in the master file directory and then the extra PDF directories. Selected by the p{ext}:
+ *   prefix.
+ *
+ * TODO: If selector ?p{ext}: is not flexible enough then maybe we should implement another selector:
+ * ?f{regexp_with_basename}:
+ *
+ * It will be processed like this:
+ *
+ * 1. regexp_with_basename undergoes %-token replacement
+ *    %m is replaced by the complete basename of the master file.
+ *    %% is replaced by %
+ * 2. The expression from 1 is used to search the master file directory, the current file
+ *    directory and the extra PDF directories.
+ * 3. If step 2 finds a matching file it is used as a selected file. If step 2 does not
+ *    find a file, then some reasonable default is used.
+ */
 QFileInfo BuildManager::parseExtendedSelectFile(QString &command, const QFileInfo &mainFile, const QFileInfo &currentFile)
 {
 	QFileInfo selectedFile;
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -535,6 +535,7 @@ QStringList BuildManager::parseExtendedCommandLine(QString str, const QFileInfo 
 				else if (command == "m") command = selectedFile.completeBaseName();
 				else if (command == "e") command = selectedFile.suffix();
 				else if (command.isEmpty() && !commandRem.isEmpty()); //empty search
+				else if (command == "p") command = findPdfFile(QString(), mainFile);
 				else continue; //invalid command
 
 				command.append(commandRem);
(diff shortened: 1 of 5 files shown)


benibela:
There is the colon prefix syntax to get paths to other files than the main file. e.g. `?c:a` to get the absolute path to the current file.

There `p:` for the pdf could be added and then you can do `?p:a` or `?p:ame` ...

MeanSquaredError:
Yes, I was thinking of using p: to specify search for pdf file, that is a file with the same basename as the main file and extension .pdf

Then I thought that we probably want to use the same approach for the DVI and PS viewers too. 
So the question is how to specify the expansion for DVI and PS files?
Maybe use expansions like 
?p:am.dvi
?p:am.pdf
?p:am.ps
This would imply that the ?p:am expansion would have to look-ahead the following extension on the command-line(.dvi .pdf .ps) and to modify the search so that it would look for a file with the extension that follows after ?p:am

MaxSagebaum:
I do not know how hard the look-ahead would be to implement. From a documentation and readability perspective it would be quite hard to understand since it is the only command that uses this kind of functionality. Especially if the creation script creates some additional files based on the output pdf name.

Then you would have for a dependency file: ?p:am.pdf.d

If you would like to add a postfix this would look like ?p:am)-depend.pdf. Here, the look-ahead would need to look over '-depend' to find the prefix.

If you have some other dot in there ?p:am.gen.pdf the 'p' command would look for a '.gen' file which might not exist.

Therefore I would vote for ?p{pdf}:am.pdf which uses pdf two times but would make it clear that 'p' searches for a pdf file or any other file extension the user can then specify

benibela:
There is also already some kind of full search, when you use `*` in the  name. Although that has the issue that it is searched before running any commands, so on the first compilation, when there is only the tex, it will not find any file, and thus not run anything.

MeanSquaredError:
@benibela Yes, I looking at the code and noticed both the c: modifier and the * character which enables full search. Now the question is how to combine all these things together without breaking anything.

The way I see the new functionality it must fulfill the following requirements:

1. Don't break any existing settings. After adding the new code TXS must keep working the same way with any existing settings that users have configured so far.
2. The new expansion characters must be easily understandable for users other than the developers who wrote the code.
3. The new code must be easily maintainable.
4. The new expansion characters must be flexible enough. At the very least they must work not only for .pdf but also for other extensions like .dvi and .ps

Almost impossible task :-)

Anyway, after giving it a bit of thought I would say that IMHO the best new syntax is :

?p{pdf}:am.pdf or the equivalent ?p{pdf}:ame

Here is how I imagine it should work:

1. Generic syntax should be ?p{ext}:am.ext or ?p{ext}:ame
2. The system would allow expansion modifiers c: and p{ext}:
without a modifier it will select the master file (as it does now).
with c: it will select the current file (as it does now)
with p{ext} it will do the following:
A) It will build a filename consisting of the master basename and .ext as an extension.
B) It will search the master directory and the additional PDF search directories for the filename built in A.
C) If a matching file is found, then its absolute pathname will be selected for further processing. If no matching file is found we can either select some reasonable default (e.g. basename.ext in the master file directory) or generate an error.
3. If will do the pathname expansion like it does normally using the remaining letters but using the found absolute pathname from 2 instead of the current or master .tex file.

So with this semantics the proper way to reference the generated .pdf file will be :

?p{pdf}:am.pdf or the equivalent ?p{pdf}:ame
(I prefer the second syntax but the first one will work too).

It will be at least 3-4 days until I find the time to implement these changes, so if anyone has any comments or ideas, you are welcome to give ideas.

MeanSquaredError:
Changed the expansion to ?p{ext}: as per the discussion above. Please do not merge yet, because I still need to add the explanation on the Configure/Commands screen and to add a simple test.

MeanSquaredError:
OK, I added an explanation about the ?p{ext}: file selector on the configuration dialog and also added a simple test for that selector. So I think the PR is suitable for a review (and possible inclusion into the source tree if it is good enough).

I also added a comment into the source code documenting the idea for another file selector (in case ?p{ext}: is not flexible enough). I will copy&paste it here:

`TODO: If selector ?p{ext}: is not flexible enough then maybe we should implement another selector:`
` ?f{regexp_with_basename}:`
` It will be processed like this:`
` 1. regexp_with_basename undergoes %-token replacement`
`    %m is replaced by the complete basename of the master file.`
`    %% is replaced by %`
` 2. The expression from 1 is used to search the master file directory, the current file`
`    directory and the extra PDF directories.`
` 3. If step 2 finds a matching file it is used as a selected file. If step 2 does not`
`    find a file, then some reasonable default is used.`

sunderme:
thanks
the manual would need an update ...

MeanSquaredError:
@sunderme , @timhoffm , @benibela 

I thought about documenting the ?p option in the manual too after I added the code but could not find the manual source code anywhere. I guess I am missing something obvious, but can you please point me to the URL where the manual source is located? If it is not available in a repository then I can just send you the proposed changes in the documentation as plaintext.

sunderme:
it is in the repository under utilities/manual/…

PR #762, opened 2019-10-23
Remove redundant calls to std::move

MeanSquaredError:
In src/latexdocument.cpp there are three functions which end with
`return std::move(result);`

Modern C++ compilers starting with C++11 (which includes GCC9) are allowed to do "copy elision" and construct the result directly in the destination address without performing an extra object construction/copy even if the object copy/move constructor has side effects. So the call to std::move is redundant and GCC9 warns about it.

More information about copy elision can be found here:
https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c/
https://en.cppreference.com/w/cpp/language/copy_elision

The proposed patch removes the redundant calls to std::move()

### DIFF
--- a/src/latexdocument.cpp
+++ b/src/latexdocument.cpp
@@ -1394,7 +1394,7 @@ QMultiHash<QDocumentLineHandle *, int> LatexDocument::getBibItems(const QString 
 			}
 		}
 	}
-	return std::move(result);
+	return result;
 }
 
 QMultiHash<QDocumentLineHandle *, int> LatexDocument::getLabels(const QString &name)
@@ -1409,7 +1409,7 @@ QMultiHash<QDocumentLineHandle *, int> LatexDocument::getLabels(const QString &n
 			}
 		}
 	}
-	return std::move(result);
+	return result;
 }
 
 QMultiHash<QDocumentLineHandle *, int> LatexDocument::getRefs(const QString &name)
@@ -1424,7 +1424,7 @@ QMultiHash<QDocumentLineHandle *, int> LatexDocument::getRefs(const QString &nam
 			}
 		}
 	}
-	return std::move(result);
+	return result;
 }
 
 /*!


MeanSquaredError:
The changes build correctly but Travis CI seems to timeout randomly when it tries to fetch a gpg key, so it breaks the build randomly.

timhoffm:
Restarted the failed job.

What's the effect of this PR on older compilers? Do we have to require a C++11 compiler with this?

MeanSquaredError:
1. On new compilers that support both std::move and we gain a very small (most likely zero) speed improvement.

2. On old compilers that support "return value optimization" (a type of copy elision) but don't support std::move we would not be able to compile the code before the PR.

3. On old compilers that support std::move but don't support "return value optimization"  we would get a slight slowdown. However I cannot think of a popular compiler that implemented std::move before "return value optimization". Most compilers implemented "return value optimization" after 2003 when it was made legal by C++03. Regarding std::move, I think it was defined in C++11 in 2011.

4. On old compilers that have neither std::move nor return value optimization we wouldn't be able to compile TXS before this PR. Not sure whether we would be able to compile it after. The PR removes std::move, but it is quite likely the current TXS code depends on newer libraries and language features anyway.

Overall I think the main benefit is that we get one less compiler warning and slightly simpler code (I have never liked or used std::move :-)

I wouldn't really bother with it but since I've been assigned https://github.com/texstudio-org/texstudio/issues/748 I decided to get rid of all the compiler/linker warnings and to fix the failing tests.

sunderme:
txs still builds on gcc 4.8. Unless it is proven that we can still build on centos-7 , I would not touch this issue. (see https://build.opensuse.org/package/show/home:jsundermeyer/texstudio-qt4-2.12.16)

MeanSquaredError:
The patch does not break the ability of the code to build anywhere because it removes the call to std::move. Basically it turns
return std::move(result);
into
return result;

That "return result;" will build anywhere where Qt4 can build because result is a QList.
Also given that the current version builds on centos7/gcc4.8 it means that gcc4.8 supports std::move and therefore it also supports return value optimization, so there won't be a slowdown either.

MeanSquaredError:
Actually after looking at the history for src/latexdocument.cpp it seems that "return result;" was replaced with "return std::move(result);" just a couple of weeks ago, so given that it build fine on centos7 until then I am pretty sure that this change will not break anything because it just reverts that change.

sunderme:
okay, let's have a try.

PR #765, opened 2019-10-26
Fix QSearchReplacePanelTest::findSpecialCase2

MeanSquaredError:
This PR fixes the test QSearchReplacePanelTest::findSpecialCase2.

It needed a Qt call to make the search panel active and then another Qt call to process the queued window activation event.


sunderme:
Thanks.
I wanted to look into it but ...

PR #766, opened 2019-10-27
Fix the FAIL messages in HelpTest::packageDocFile

MeanSquaredError:
While running tests I noticed that HelpTest::packageDocFile was failing and the error messages did not tell clearly what was wrong with the help system. It turned out that the problem was that texdoc was not in the search path of TeXstudio.

This PR changes the error messages to make them clearer and it adds a separate error message for the case when TeXstudio is unable to find texdoc.


sunderme:
thanks

PR #770, opened 2019-10-30
Fix tests that failed in non-EN locales by translating UI strings.

MeanSquaredError:
This PR fixes tests in two files (latexcompleter_t.cpp and structureview_t.cpp) that failed when run in non-English locales. There are a few strings that get translated in different locales and the tests did not expect the translated strings. The fixes are implemented by applying tr() for the corresponding context class.

This fixes all the test errors reported in https://github.com/texstudio-org/texstudio/issues/748


sunderme:
Thnaks

PR #793, opened 2019-11-12
Cwl changes

muzimuzhi:
Main changes:
 - Append etex primitives to `tex.cwl`.
Since there is a latex package named `etex`, to save the namespace, etex primitives are appended to `tex.cwl`, rather than reside in `etex.cwl`.
 - Add predefined `x11names` color names to `xcolor.cwl`, and move always available color names out from `#ifOption`.
 - Some changes to `tikz.cwl` including updating list of tikz libraries and adding some frequently used commands and options.
`tikz` provdes hundreds of options, this pr adds only small part of them.


sunderme:
thanks

PR #799, opened 2019-11-17
Fix token type names

MeanSquaredError:
Currently the latex tokenizer does not provide the names for tokens of type "shorttitle" and "formula". This makes debugging of tests more difficult because shorttitle and formula tokens are displayed as "UnknownTokenType".

This PR adds the names for "shorttitle" and "formula" tokens.


sunderme:
thanks

Issue #802, opened 2019-11-21
Question about execCommand()


MeanSquaredError:
@timhoffm , @sunderme , @benibela 

Hi,

This is a question for the developers of TXS. I am trying to cleanup the code in latexstyleparser.cpp which invokes external programs like pdflatex but fails to set the search path for these programs so their invocation fails on some platforms (the most recent such bug is https://github.com/texstudio-org/texstudio/issues/768 ) There are various places where QProcess is used, so I'd like to write a common wrapper for QProcess or preferably use an existing wrapper function. The wrapper would setup the search path for the executable before actually running the executable pretty much like ProcessX does.

ProcessX is too specialized, so it is not suitable for a generic QProcess wrapper. However there is one other method,
`QString execCommand(const QString &cmd,QString additionalPaths="");`
from utilsSystem.cpp

It seems to be doing almost everything that I need and it even receives an optional string argument `additionalPaths`. However this `additionalPaths` is not the path that the command is searched for, but the paths that are added to the PATH of the new process.

After some looking through the code it seems that the only caller that specifies additional paths is `KpathSeaParser::kpsewhich()` which executes kpsewhich and passes it some additional paths which are taken from `BuildManager::additionalSearchPaths`.

It really seems to me that the original intention was to specify the paths where to search for the executed command and not to augment the environment of the executed command itself. After all what would someone want to adjust the PATH variable of kpsewhich? It has its own path variables and (I think but might be wrong) that kpsewhich does not use PATH when searching.

So I am planning to modify execCommand so that `additionalPaths` would be the paths where to search for the actual executable (instead of just augmenting the environment of the executable itself).

Can anyone comment on this? Do you see any pitfalls?

MeanSquaredError:
OK, cleaned up the program execution code and moved it to a separate class called `ExecProgram` which is similar to ProcessX but much simpler. If everything works well it will be a replacement for `execCommand`. If tests show no problem I will provide a patch which will also fix  #768

MeanSquaredError:
OK, I built a test branch with the new ExecProgram code. It passes the tests and seems to work OK in the sense that it autogenerates .cwl and runs correctly various external commands (granted I did not do heavy testing). If it fixes #768 then I will make it into a PR.

muzimuzhi:
@MeanSquaredError 
[...]

PR #803, opened 2019-11-22
Cleanup execution of external programs. Fix #768

### LINKED ISSUE #768: When does the cwl-auto-generating function run?
### Environment
- TeXstudio: 2.12.16
- Qt: Using Qt Version 5.12.3, compiled with Qt 5.12.3 R
- OS: macOS 10.14.6
- TeX distribution: macTeX 2019

### Description

Under what condition does the cwl-auto-generating function run? Here are two failing examples I encountered:
 - I have used `nicematrix` package during recent weeks for several times but there is no `nicematrix.cwl` found in directory `~/.config/texstudio/completion/autogenerated`.
 - There used to be a nearly empty `tkz-euclid.cwl` in the above directory. But if I delete it and then reopen texstudio and compile a tex file using `tkz-euclid` package (and compile that file for several times), file `tkz-euclid.cwl` is never re-generated.

The most recent auto-generated cwl files are created on Oct 13, 2019, and the second recent date is Oct 10, 2019. I have checked that neither texstudio or macOS is upgraded between that two dates.

MeanSquaredError:
This is the PR following the discussion in #768 and after implementing what I suggested in #802 

Basically TXS has many places where it invokes different external programs. Two of the most often used ones are in ProcessX (where it executes txs:/// commands) and LatexStyleParser (where it invokes kpathsea/miktex external programs in order to parse tex files and generate .cwl definitions).

In all these places the code uses different ad-hoc implementations to 
1. Adjust the system PATH so that QProcess would be able to find executables without absolute path.
2. Wait for program to complete.
3. Check whether it completed successfully.
4. Read its output.

All these adhoc solutions use different approaches, #ifdefs for the different platforms and generally some of them may work for one platform but others will fail, just like with #768 where running of the build chain from ProcessX worked but running pdflatex from Latexstyleparser failed. Additional problem was that different solutions use different mixes of latin1, utf8 and local text encoding which is another source of possible issues (the correct approach as implied by qgetenv/qputenv is to use only local 8-bit encoding).

So what this patch does is to 

1. Implement a simple class ExecProgram which has two wrapper methods around QProcess.start().

ExecProgram.execAndWait (synchronous)
ExecProgram.execAndNoWait (asynchronous)

Both wrappers adjust the path to include the paths visible from the command shell and optionally to include caller-provided additional search directories.

2. Change LatexStyleParser, ProcessX, KpathseaParser and a couple of other places to use ExecProgram instead of ExecCommand().

3. Remove the no-longer used ExecCommand

### DIFF
--- a/src/execprogram.h
+++ b/src/execprogram.h
@@ -1,0 +1,25 @@
+#ifndef EXECPROGRAM_H
+#define EXECPROGRAM_H
+
+#include "mostQtHeaders.h"
+
+class ExecProgram
+{
+public:
+	// Input parameters
+	QString program;
+	QStringList arguments;
+	QString additionalSearchPaths;
+	QString workingDirectory;
+
+	// Output parameters. Only assigned by synchronous execAndWait
+	bool normalRun; // If false then program either did not run or crashed
+	int exitCode;
+	QString standardOutput;
+	QString standardError;
+
+	void execAndWait (void);
+	void execAndNoWait (QProcess &proc) const;
+};
+
+#endif // EXECPROGRAM_H
--- a/src/execprogram.h
+++ b/src/execprogram.h
@@ -6,20 +6,24 @@
 class ExecProgram
 {
 public:
-	// Input parameters
-	QString program;
-	QStringList arguments;
-	QString additionalSearchPaths;
-	QString workingDirectory;
-
-	// Output parameters. Only assigned by synchronous execAndWait
-	bool normalRun; // If false then program either did not run or crashed
-	int exitCode;
-	QString standardOutput;
-	QString standardError;
+	ExecProgram(void);
+	ExecProgram(const QString &progNameAndArguments, const QString &additionalSearchPaths, const QString &workingDirectory = QString());
+	ExecProgram(const QString &progName, const QStringList &arguments, const QString &additionalSearchPaths = QString(), const QString &workingDirectory = QString());
 
 	bool execAndWait(void);
 	void execAndNoWait(QProcess &proc) const;
+
+	// Input parameters
+	QString m_program;
+	QStringList m_arguments;
+	QString m_additionalSearchPaths;
+	QString m_workingDirectory;
+
+	// Output parameters. Only assigned by synchronous execAndWait
+	bool m_normalRun; // If false then program either did not run or crashed
+	int m_exitCode;
+	QString m_standardOutput;
+	QString m_standardError;
 };
 
 #endif // EXECPROGRAM_H
--- a/src/execprogram.h
+++ b/src/execprogram.h
@@ -18,7 +18,7 @@ public:
 	QString standardOutput;
 	QString standardError;
 
-	void execAndWait (void);
+	bool execAndWait (void);
 	void execAndNoWait (QProcess &proc) const;
 };
 
--- a/src/execprogram.h
+++ b/src/execprogram.h
@@ -18,8 +18,8 @@ public:
 	QString standardOutput;
 	QString standardError;
 
-	bool execAndWait (void);
-	void execAndNoWait (QProcess &proc) const;
+	bool execAndWait(void);
+	void execAndNoWait(QProcess &proc) const;
 };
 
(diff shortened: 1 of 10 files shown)


src/latexstyleparser.cpp
```diff
@@ -736,28 +734,18 @@ QStringList LatexStyleParser::readPackageTexDef(QString fn) const
...
 	}
-	QStringList args;
-	QString result;
-	args << "-t" << "latex" << "-l" << "-p" << fname;
-	myProc.start(texdefDir + "texdef", args);
-	args.clear();
-	myProc.waitForFinished();
-	if (myProc.exitCode() != 0)
+	execProgram.execAndWait ();
+	if (!execProgram.normalRun || (execProgram.exitCode != 0)) {
```
sunderme: to my taste, a method like runFinishedWithoutError could simply things here (and later)

src/latexstyleparser.cpp
```diff
@@ -823,29 +811,19 @@ QStringList LatexStyleParser::readPackageTracing(QString fn) const
...
-	args << "-draftmode" << "-interaction=nonstopmode" << "--disable-installer" << tf->fileName();
-	myProc.setWorkingDirectory(QFileInfo(tf->fileName()).absoluteDir().absolutePath());
-	myProc.start(texdefDir + "pdflatex", args);
-	args.clear();
-	myProc.waitForFinished();
-	if (myProc.exitCode() != 0)
+	execProgram.arguments << "-draftmode" << "-interaction=nonstopmode" << "--disable-installer" << tf->fileName();
+	execProgram.workingDirectory = QFileInfo(tf->fileName()).absoluteDir().absolutePath();
+	execProgram.execAndWait();
+	if (!execProgram.normalRun || (execProgram.exitCode != 0)) {
```
sunderme: like here, see above

src/execprogram.h
```diff
@@ -0,0 +1,25 @@
+#ifndef EXECPROGRAM_H
+#define EXECPROGRAM_H
+
+#include "mostQtHeaders.h"
+
+class ExecProgram
+{
+public:
+	// Input parameters
```
sunderme: personally, I would like an constructor which allows setting the main parameters ...

MeanSquaredError:
OK, I changed the execAndWait so that it returns a boolean indicating normal termination with exit code 0. Also simplified the code that call execAndWait to check the return value.
Did not rename execAndWait -> runFinishedWithoutError, because I could not come up with a good new name for execAndNoWait, but if you have a good name for it too, then I can rename both methods.

Now what is left is to add the constructor(s).

MeanSquaredError:
OK, I added three different constructors to ExecProgram to initialize the member variables and changed the calling code to use the constructors.

I also prefixed the names of the data members of ExecProgram with m_ to prevent name clashes of the data members with the constructor arguments.

I think the PR can now undergo another review.

MeanSquaredError:
Ooops, Codacy complained :-) let me fix it.

src/latexstyleparser.cpp
MeanSquaredError: OK, changed it so that `execAndWait` effectively returns 
`normalRun && (execProgram.exitCode == 0)`
So now we are checking the result when calling `execProgram.execAndWait`

src/latexstyleparser.cpp
MeanSquaredError: OK, changed it so that `execAndWait` effectively returns 
`normalRun && (execProgram.exitCode == 0)`
So now we are checking the result when calling `execProgram.execAndWait`

src/execprogram.h
MeanSquaredError: Added the constructors.

MeanSquaredError:
OK, fixed the issues by initializing m_normalRun and moving all the assignments to the constructor initializer list. Hopefully Codacy is OK with it now.

sunderme:
looks good

sunderme:
thanks

muzimuzhi:
@MeanSquaredError Many thanks.

PR #810, opened 2019-11-22
"Show Line Numbers" setting improvements

### LINKED ISSUE #804: Line numbering too deep in advanced settings
Enabling line numbers in a fresh TeXstudio installation takes a lot of clicks:
Options -> Configure TeXstudio -> Show Advanced Options -> "OK" -> Adv. Editor -> Show line numbers

Line numbers shouldn't be an advanced setting, especially considering that error messages always report line numbers.

Two suggestions:

* Move "Show line numbers" from "Adv. Editor" to "Editor" in settings.
* Make "All line numbers" the default setting.

### Environment
- TeXstudio: 2.12.16
- OS: all
- TeX distribution: all

thomas-daniels:
Resolves #804 

* Puts "Show Line Numbers" in Editor (rather than Adv. Editor), makes it a 'normal' option rather than an advanced option.
* Makes "All Line Numbers" the default value.

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -1549,7 +1549,7 @@ Then you can select a new shortcut by one of the following ways:
           <string>Editor</string>
          </property>
          <layout class="QGridLayout">
-          <item row="12" column="0">
+          <item row="13" column="0">
            <widget class="QCheckBox" name="checkBoxRealTimeCheck">
             <property name="sizePolicy">
              <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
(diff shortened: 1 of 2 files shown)


sunderme:
I am not sure that I want to see line numbers as default ...
I agree with putting the option as non-advanced

thomas-daniels:
@sunderme I can recreate the PR with only the first commit, what's your opinion on that?

sunderme:
> @sunderme I can recreate the PR with only the first commit, what's your opinion on that?

That would be great !

thomas-daniels:
okay, moving to #811

PR #814, opened 2019-11-26
Fix bug in QEditorTest which prevents debug builds from starting

MeanSquaredError:
There is a bug in the QEditorTest which may prevent debug builds of TXS from starting.

If latex-document.cwl is disabled in the configuration and a fresh debug build of TXS is started it will try to run all the tests and QEditorTest will cause a failing debug assertion that will invoke the crash handler which will ask the user whether he wants to terminate TXS or to try to recover. Recovering does not work reliably (sometimes TXS works and sometimes it just locks up afterwards).

The problem is caused by the fact that QEditorTest relies on \section blocks being foldable and there are several places in the test where it tries to fold/unfold \section blocks. When latex-document.cwl is disabled \section is just an unknown command and its blocks are not foldable. QEditorTest tries to fold them and causes a debug assertion to fail in QNFADefinition::fold()

The proposed PR fixes this issue in QEditorTest by checking if a block is foldable before actually trying to fold it. If the block is not foldable then the corresponding test just fails, which is better than crashing the whole application.

### DIFF
--- a/src/tests/qeditor_t.cpp
+++ b/src/tests/qeditor_t.cpp
@@ -10,7 +10,7 @@
 QEditorTest::QEditorTest(QEditor* ed, bool executeAllTests):allTests(executeAllTests)
 {
 	editor=ed;
-    if (editor->getFileCodec()!=nullptr) defaultCodec = editor->getFileCodec();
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

PR #821, opened 2019-11-30
Small framework for debug logging

MeanSquaredError:
This PR adds a small framework for extra debug logging that can be used to log debug information to a file even when Qt logging is disabled globally from qtlogging.ini
I wrote it while investigating #768 so I decided to clean it up and send it as a PR because it may be handy on other occasions too. Key features:

1. Enabled by defining the configuration macro DEBUG_LOGGER (which is undefined by default). Can be enabled from BUILD.sh but it is off by default even for debug builds.

2. Adds several debug macros DLOG_DEBUG, DLOG_INFO, DLOG_WARNING, DLOG_CRITICAL, DLOG_FATAL. These macros correspond to the Qt logging levels, but they work independently of the Qt logging framework, so they will work even if Qt logging has been disabled as long as TXS is compiled with DEBUG_LOGGER. These macros have the benefit that they are completely optimized out and do not leave any trace in the code when DEBUG_LOGGER is undefined. Unlike that the standard Qt debug logging "qDebug() << some_funciton()" will still leave some code even when QT_NO_DEBUG is defined.

3. Standard Qt logging through qDebug(), qWarning(), etc. still works as usual.

4. When the logging framework is enabled it adds a new command-line option "--debug-logfile /pathname/to/logfile.txt" which enables debug logging for the corresponding run of TXS. When logging to a file the following things happen:

4.1. All debug messages (both from the DLOG_XXXXXX macros and from the standard qDebug(), qWarning(), ... are written into the specified debug logfile.

4.2. All Qt debug messages from TXS are enabled regardless of the global setting in qtlogging.ini. This is quite handy on platforms like Fedora where Qt disables globally the debug messages, so in order to view debug messages from TXS one has to enable debug messages globally but then gets a number of debug messages from other unrelated programs.

5. The PR adds two extra classes DebugLoggerMain and DebugLoggerQt. The classes are only visible when built with DEBUG_LOGGER. The .cpp and .h files are placed in src/debug/ and the PR also moves a couple of other unrelated files ( debughelper.* ) into that directory.

6. I tried to make the as unintrusive as possible, so it makes minimal changes to the other TXS code. The logging framework works with Qt4 and Qt5.


sunderme:
thanks

PR #824, opened 2019-12-05
add Italian dictionary and thesaurus from PLIO

jacopodonati:
These files are the same files you can [find here](https://extensions.libreoffice.org/extensions/italian-dictionary-thesaurus-hyphenation-patterns). However, importing `dict-it.oxt` directly into TeXstudio gives an error because it seems unable to find a Hunspell dictionary.


sunderme:
thanks

PR #826, opened 2019-12-05
Fix a bug in cursor movement through soft-wrapped lines

MeanSquaredError:
This PR fixes a bug in the TXS editor (qcodeedit) where the cursor moves incorrectly through wrapped lines when soft line breaks are enabled.

TXS editor cursors have a feature that they internally call "column memory". Basically it is the ability to remember the current horizontal position of the cursor. That horizontal position is remembered and when the cursor moves vertically without the user moving it horizontally, the editor tries to position it horizontally at the same horizontal offset that it had during the last horizontal movement. That is not possible on each line, but the editor finds the closest possible horizontal position.

That column memory works correctly in most cases, but it fails (and gets changed) during some movements through wrapped lines with soft linebreaks.

I am attaching an image that shows the issue.

all_lines

 The image constist of 5 sequential screenshots where the user tries to select the first three lines of text (the second line is wrapped and occupies 3 screen lines). The cursor is marked with a red vertical rectangle. Between each screen the user presses SHIFT+Down_Arrow to select the next line.

The intention of the user is to select the 3rd line completely and that is expected since the initial cursor offset was the leftmost position (column 0) on line 1. However on last screen the cursor is not at column 0 of line 3 but at column 12 of line 3, so selecting the whole line 3 would require the user to press SHIFT+DOWN and then one extra keypress SHIFT+HOME to remove the extra selection caused by incorrect column memory that was changed from column 0 to column 12.

This is an annoying bug caused by the fact that in several places when restoring the current position the code does new_x_position = qMax(currrent_position, m_max) (here m_max is the saved original X position, the "column memory")
This works fine in most cases except when going through soft-wrapped lined that have an initial offset added by tabs. The correct solution is just to do new_x_position = m_max.

The proposed patch renames m_max to m_savedX, fixes the positioning and adds six basic tests for the fix (tests are in QDocumentCursorTest::columnMemory() ).

### DIFF
--- a/src/qcodeedit/lib/document/qdocument.cpp
+++ b/src/qcodeedit/lib/document/qdocument.cpp
@@ -118,9 +118,9 @@ struct RenderRange
 	int wrap;
 };
 
-template<typename T> struct CacheMeta { 
-	static inline bool exists(const T&); 
-	static inline void clearArray(T*, int size); 
+template<typename T> struct CacheMeta {
+	static inline bool exists(const T&);
+	static inline void clearArray(T*, int size);
 };
 
 template<> bool CacheMeta<int>::exists(const int& v) { return v >= 0; }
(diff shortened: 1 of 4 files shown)


MeanSquaredError:
Just to add about this "column memory" the popular text editors like (KDE KATE) implement the column memory correctly and restore the cursor position on line 3 to column 0 unlike qcodeedit which places it on column 12.

sunderme:
thanks

Issue #828, opened 2019-12-09
Failed to compile latest git version of texstudio with phonon on ubuntu 19.10.


hongyi-zhao:
H,

I try to compile the latest git version of texstudio with phonon on ubuntu 19.10.  The steps is as follows:

```
sudo apt-get build-dep texstudio
sudo apt-get install build-essential debhelper devscripts qtbase5-dev qt5-default qt5-qmake  libqt5svg5-dev qtscript5-dev qttools5-dev libpoppler-qt5-dev zlib1g-dev pkg-config libphonon4qt5-dev
qmake
make -j4
```
And I meet the following errors:

```
/usr/bin/ld: .obj/PDFDocument.o: in function `PDFMovie::realPlay()':
PDFDocument.cpp:(.text+0x35e): undefined reference to `Phonon::VideoPlayer::isPlaying() const'
/usr/bin/ld: PDFDocument.cpp:(.text+0x374): undefined reference to `Phonon::VideoPlayer::isPaused() const'
/usr/bin/ld: PDFDocument.cpp:(.text+0x380): undefined reference to `Phonon::VideoPlayer::currentTime() const'
/usr/bin/ld: PDFDocument.cpp:(.text+0x38b): undefined reference to `Phonon::VideoPlayer::totalTime() const'
/usr/bin/ld: PDFDocument.cpp:(.text+0x39a): undefined reference to `Phonon::VideoPlayer::seek(long long)'
/usr/bin/ld: .obj/PDFDocument.o: in function `PDFMovie::mouseReleaseEvent(QMouseEvent*)':
PDFDocument.cpp:(.text+0x3e1): undefined reference to `Phonon::VideoPlayer::isPlaying() const'
/usr/bin/ld: PDFDocument.cpp:(.text+0x3ed): undefined reference to `Phonon::VideoPlayer::pause()'
/usr/bin/ld: .obj/PDFDocument.o: in function `PDFMovie::PDFMovie(PDFWidget*, QSharedPointer<Poppler::MovieAnnotation>, int)':
PDFDocument.cpp:(.text+0xacf): undefined reference to `Phonon::VideoPlayer::VideoPlayer(QWidget*)'
/usr/bin/ld: PDFDocument.cpp:(.text+0xc9b): undefined reference to `Phonon::MediaSource::MediaSource(QString const&)'
[...]

PR #829, opened 2019-12-10
Workaround for a qmake bug when building with phonon support

MeanSquaredError:
This PR fixes https://github.com/texstudio-org/texstudio/issues/828
There is a bug in Qt 5.12.x where `QT += phonon4qt5` does not add `-lphonon4qt5` to the linker command line. The proposed PR adds a  workaround for this issue which shows up in Ubuntu 19, Fedora 30 (and probably in any distribution that uses Qt 5.12.x)


sunderme:
Thanks

Issue #842, opened 2019-12-27
Question: Cleaning up the C++ code in src/latexparser/


MeanSquaredError:
Hi,

I started working on the script that reformats the TeX source code and in the process I reached the conclusion that the script needs to be able tokenize/parse the TeX source. I started writing my own tokenizer/parser in QtScript but in the process I reached the conclusion that I don't want to reinvent the wheel and that the best approach would be to export some of the TXS latex-parsing code from src/latexparser/ to the userspace scripts. This would also allow me to use the syntax information from the .cwl files, something which a parser written in pure QtScript wouldn't be able to do.

Looking at the code in src/latexparser/ it seems that the current code there is in the process of being rewritten. I can spend some time and help with the rewrite if someone tells what is the rewrite plan?

It seems that at least some of the methods and data members of "class LatexParser" and "struct LatexReader" are marked as obsolete.

It seems that the aim is to get rid of "class LatexParser" and "struct LatexReader" and to replace their usage with usage of methods of namespace "Parsing". However at least one of the methods (i.e. Parsing::latexDetermineContexts2) uses LatexParser.

So maybe the plan is to incorporate the methods of Parsing:: into class LatexParser ?

Can anyone, please clarify on the planned changes to the latex parsing code?
@sunderme , @timhoffm , @benibela

sunderme:
There is no overall plan.
With the Qt4 removal, it may be a good time to clean up code.
LatexReader is the older functionality which was superseded by the token system, but probably not all places have been adapted. So removing old LatexReader code may be a first step.

sunderme:
By the way,if infrastructure needs to be adapted, a c++ implementation may be simpler if there are no other benefits...

MeanSquaredError:
Thank you for the clarification.
I will remove LatexReader and will provide a patch for that.

Regarding the .tex reformatting functionality, the script approach is (sort of) attractive because it would allow more people to test it before the reformatting code goes into the main code. After all, I suspect that fewer people will be able to build and install a custom branch of TXS (as compared to just installing a custom macro). And besides it doesn't hurt to have the tokenizer/parser available from the user scripts because someone else might find it useful too.

On the other hand exporting the latex parsing functionality to QtScript implies exporting the `Token` class and
1. `Token` is not a Qt object. Turning it into a Qt object might imply a speed penalty even outside of QtScript and this is something that we really want to avoid, so it must be checked carefully. Or the code to export `Token` and its properties to to QtScript has to be written manually.

[...]

PR #843, opened 2019-12-27
Remove LatexReader from SyntaxCheck

MeanSquaredError:
Based on the discussion in https://github.com/texstudio-org/texstudio/issues/842 this is the first of several PRs that remove LatexReader from the code. This PR provides two changes:

1. Remove `LatexReader` from the code in `SyntaxCheck`. This is an easy change because the function where it was used (`SyntaxCheck::checkLine` with 5 arguments) is actually dead code (unused) and there is another overload (`SyntaxCheck::checkLine` with 7 arguments) that is being used. That overload with 7 args uses the new `Token` class instead of the obsolete `LatexReader`. So we are just removing dead code.

2. The PR also changes two places in `SyntaxCheck::run` where we used the literal constant 3 instead of the correct enum `QDocumentLine::UNCLOSED_ENVIRONMENT_COOKIE`


sunderme:
thanks

Issue #856, opened 2020-01-03
Bibliography update not working when using build directory


Jamen1993:
### Environment

- TeXstudio: 2.12.16 and 2.12.18
- Qt: 5.12.1
- OS: Windows 10 1909
- TeX distribution: texlive
- Language: German

### Expected behavior

TS is able to detect whether the bibliography was changed and decides if the bibliography tool must be run when doing a build.
I'd also expect this behaviour if I configured pdflatex and the bibliography tool to use a build directory.

### Actual behavior

The bibliography tool runs with every build, whether the bibliography was changed or not.

### How to reproduce

1. Setup pdflatex and biber to use a build directory
![grafik](https://user-images.githubusercontent.com/17482299/71734926-5adcca00-2e4d-11ea-8527-6563146189de.png)
2. Create the build directory in your working directory
3. Create a basic project with bibliography

`File: test.tex`
```latex
\documentclass[11pt,a4paper]{article}
\usepackage[utf8]{inputenc}
\usepackage[T1]{fontenc}
\usepackage[british]{babel}
\usepackage{amsmath}
\usepackage{amsfonts}
\usepackage{amssymb}
\usepackage{graphicx}

\usepackage{biblatex}
\bibliography{literature}

\begin{document}
	Test
	
	\appendix
	\nocite{*}
	\printbibliography
\end{document}
```
`File: literature.bib`
```bibtex
@article{test,
	author = {Bach, Sandra},
	title = {A very good article},
	journaltitle = {Bad Journal},
	date = {2019-08}
}

@article{test2,
	author = {Bach, Sandra},
	title = {A very good article again},
	journaltitle = {Bad Journal},
	date = {2019-10}
}

@article{test3,
	author = {Bach, Sandra},
	title = {A very good article once again},
	journaltitle = {Bad Journal},
	date = {2019-12}
}
```

4. Build the project twice and watch biber run the second time even though the bibliography has not changed.

sunderme:
duplicat eto #855

Jamen1993:
Are you sure? This doesn't work in 2.12.16 as well.

MeanSquaredError:
@Jamen1993 
It might be resolved after #855 was fixed. So can you try it with the latest development build of TXS from
https://bintray.com/sunderme/texstudio/texstudio-win#files
and see if the problem with biber is fixed?

Jamen1993:
[...]

PR #860, opened 2020-01-06
Fix for issue #856

MeanSquaredError:
This PR fixes issue #856
The problem was caused by the fact that when searching for an existing .bbl file we were passing a relative pathname (actually just a .bbl filename) for that existing .bbl file. Since the user also specified relative pathname for the additional log files, the generic search function `BuildManager::findfile()` failed to find the .bbl file.

The proposed fix replaces `BuildManager::findfile()` with a separate class `FindInDirs`. That class has the following features:

1. It does everything that `BuildManager::findfile()` could do
2. It enforces search in absolute directory pathnames. This enforcing is implemented by specifying an absolute directory in the constructor ( `resolveDir`). Any search directory that is loaded afterwards is checked and if it is not absolute, then it is resolved to an absolute directory by prepending `resolveDir` to it. This enforcing of absolute pathnames takes care of the problem with the lookup of .bbl filenames.
3. The return value from `FindInDirs::findAbsolute()` is always absolute. On the other hand `BuildManager::findfile()` could return a relative pathname if it was passed a relative pathname and that relative pathname was found relative to current directory.

Ideally `FindInDirs` will be used to replace all the other search functions like `findResourceFile()`, etc. but it probably needs to be extended with extra properties/methods to achieve that.

### DIFF
--- a/src/findindirs.cpp
+++ b/src/findindirs.cpp
@@ -20,17 +20,12 @@ void FindInDirs::loadDirs(const QString &dirs)
 void FindInDirs::loadDirs(const QStringList &dirs)
 {
 	foreach(const QString &oneDir, dirs) {
-		loadOneDir(oneDir);
+		m_absDirs.push_back(
+			QDir::isAbsolutePath(oneDir) ?
+			oneDir :
+			m_resolveDir + QDir::separator() + oneDir
+		);
 	}
-}
-
-void FindInDirs::loadOneDir(const QString &dir)
-{
-	QString resolved =
-		QDir::isAbsolutePath(dir) ?
-		dir :
-		m_resolveDir + QDir::separator() + dir;
-	m_absDirs.push_back(resolved);
 }
 
 QString FindInDirs::findAbsolute(const QString &pathname)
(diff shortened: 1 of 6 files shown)


sunderme:
I have two remarks
1) i would like to see documentation/comments, especially on code which is used all ober the place
doxygen comments arenpreffered and supported in qtcreator. Try typing /*! above any function declaration
2) the check for existence and readability could be moved into that class as well.

src/findindirs.cpp
```diff
@@ -0,0 +1,66 @@
...
+	loadDirs(splitPaths(dirs));
+}
+
+void FindInDirs::loadDirs(const QStringList &dirs)
+{
+	foreach(const QString &oneDir, dirs) {
+		loadOneDir(oneDir);
+	}
+}
+
```
sunderme: not sure that we need that extra function.
loadDirs should be sufficient

MeanSquaredError:
OK, I did the following changes:

1. Added Doxygen documentation for the `FindInDirs` class.
2. Added an extra argument to `FindInDirs::FindInDirs` that allows check for file readability.
3. Removed `FindInDirs::loadOneDir`

sunderme:
perfect

PR #862, opened 2020-01-07
Fix debug builds on MinGW platform 

MeanSquaredError:
This PR fixes the debug builds on the MinGW platform.

Currently the debug builds use `txs_assert()` and `txs_assert_x()`. However these two functions are only defined on platforms that support GNU libc. MinGW on the other hand does not use/support glibc. So currently the debug builds fail on MinGW.

The proposed PR changes the debug code so that it only uses `txs_assert()` and `txs_assert_x()` when the crash handler and its helper functions ( `txs_assert()` and `txs_assert_x()` ) are actually defined.

### DIFF
--- a/src/modifiedQObject.h
+++ b/src/modifiedQObject.h
@@ -4,8 +4,8 @@
 
 #include <qglobal.h>
 
-
-#ifndef NO_CRASH_HANDLER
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #867, opened 2020-01-08
Fix linked libraries on Windows

MeanSquaredError:
This PR fixes a build error on Windows 7, Qt 5.12.6 and MinGW 7.3.0 (bundled with Qt)

Currently when trying to build using the following configuration
`qmake.exe "CONFIG+=debug" "CONFIG-=debug_and_release" "CONFIG-=release"`
it fails to find the zlib and Shlwapi libraries.
zlib is used by the internal PDF viewer and QuaZIP
Shlwapi is used by SyncTeX (part of the internal PDF viewer) when it is built on Windows.

### DIFF
--- a/texstudio.pro
+++ b/texstudio.pro
@@ -314,6 +314,9 @@ isEmpty(USE_SYSTEM_QUAZIP) {
 
 include(src/pdfviewer/pdfviewer.pri)
 LIBS += -lz # Internal PDF viewer and internal QuaZIP require zlib
+win32: {
+    LIBS += -lShlwapi # SyncTeX/Windows uses Shlwapi
+}
 
 # ###############################
 
--- a/texstudio.pro
+++ b/texstudio.pro
@@ -315,7 +315,7 @@ isEmpty(USE_SYSTEM_QUAZIP) {
 include(src/pdfviewer/pdfviewer.pri)
 LIBS += -lz # Internal PDF viewer and internal QuaZIP require zlib
 win32: {
-    LIBS += -lShlwapi # SyncTeX/Windows uses Shlwapi
+    LIBS += -lshlwapi # SyncTeX/Windows uses shlwapi
 }
 
 # ###############################
--- a/texstudio.pro
+++ b/texstudio.pro
@@ -313,6 +313,7 @@ isEmpty(USE_SYSTEM_QUAZIP) {
 }
 
 include(src/pdfviewer/pdfviewer.pri)
+LIBS += -lz # Internal PDF viewer and internal QuaZIP require zlib
 
 # ###############################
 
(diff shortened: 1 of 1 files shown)


MeanSquaredError:
The qt5win build environment fails.
That build environment defines `-DWIN32` to specify a Windows build but SyncTeX (which is a part of the internal PDF viewer) recognizes Win32 via one of the following defines
```
#if defined(_WIN32) || defined(__WIN32__) || defined(__TOS_WIN__) || defined(__WINDOWS__)
#define SYNCTEX_WINDOWS 1
#endif
```
So effectively Travis-CI / q5win builds SyncTeX in Unix mode, which is different from the actual Windows/MinGW builds where SyncTeX is built in Windows mode so it requires Shlwapi.

So I think that the proper fix would be to :
1. Add one of the above defines `_WIN32`, `__WIN32__`, `__TOS_WIN__`, `__WINDOWS__` to the qt5win build environment so that SyncTeX is actually built in Windows mode.
2. Add the shlwapi.lib library file to the qt5win build environment (it can be taken from the windows platform SDK/kit).

MeanSquaredError:
Actually, thinking again, I would say that the proper place to get shlwapi.lib from would be MinGW installation and not the Windows platform SDK.

sunderme:
I would not change the qt5win build environment if there is no benefit.
To allow building on mingw, I'd suggest to detect mingw and act accordingly... (or just use a flag in qmake)

sunderme:
i guess on linux, it is lowercase shlwapi ...

MeanSquaredError:
That's a good catch! I changed the case of shlwapi and hopefully that fixes the qt5win build.

There is a small glitch with the qt5win build, though. Currently it defines the `WIN32 ` macro, but 
`src/pdfviewer/synctex/synctex_parser_utils.c`
does not check for `WIN32` (however it checks for `_WIN32`, `__WIN32__`, `__TOS_WIN__`, `__WINDOWS__`).

On real Windows+mingw 7.3, the macros `_WIN32` and `__WIN32__` are defined so in `synctex_parser_utils.c` the Windows code branch is compiled. On the other hand qt5win only defined `WIN32` and in `synctex_parser_utils.c` the Unix code branch is compiled.

Effectively this means that the Windows code branch in SyncTeX is not tested. Not a big deal for me, just mentioning it.

MeanSquaredError:
Great, I see that the qt5win build succeeded, so it was the letter case indeed. This pretty much takes care of the Release builds on native Windows+mingw platform.

There is still another issue that breaks Debug builds on Win+mingw, but I will have to find the time to cleanup that fix a bit and submit it as a separate PR in the next couple of days.

PR #891, opened 2020-01-17
Cleanup texstudio.pro

MeanSquaredError:
This PR cleans up the project file texstudio.pro. More specifically it implements these changes:

1. Currently when creating a debug build + NO_TESTS, the test classes from src/tests/ are compiled and built into the executable. This PR separates the debug builds from the NO_TESTS option, so when building a debug executable it is possible to exclude the tests from being linked in.

2. The lists of .cpp and .h files for the test classes are moved to src/tests/tests.pri. Makes texstudio.pro simpler.

3. The conditional code for the command-line options NO_TESTS, USE_SYSTEM_HUNSPELL and USE_SYSTEM_QUAZIP is moved into the corresponding .pri files. Again this simplifies texstudio.pro.

4. The defines for -lz and -lshlwapi are moved into the corresponding .pri files (pdfviewer and quazip).

5. A few extra messages are added, so texstudio.pro it will always say whether it is using static or system libraries for hunspell, quazip and whether tests are built.


sunderme:
thanks

PR #872, opened 2020-01-09
Improve the documentation of the script command system()

MeanSquaredError:
Based on the discussion here https://github.com/texstudio-org/texstudio/issues/868 I am providing an expanded documentation for the `system()` script command.

PR #892, opened 2020-01-17
Expand the documentation for system()

MeanSquaredError:
Based on the user comment here https://github.com/texstudio-org/texstudio/pull/872#issuecomment-573415276 I am providing a PR that expands the documentation for system() and provides two examples.


sunderme:
thanks

PR #894, opened 2020-01-20
Add doxygen comments to QDocumentLineHandle cookie methods

MeanSquaredError:
While working on changes to the tokenizer code I came across an incorrect comment about QDocumentLineHandle::getCookieLocked which said that the method needs external logging. Most likely the incorrect comment was just a copy&paste error, but I decided to add doxygen code for the QDocumentLineHandle cookie methods, documenting the parameters, return values and thread safety of these methods.

I am using the doxygen comments for my own reference, but they may be useful to someone else too so I am providing a PR for them

### DIFF
--- a/src/findindirs.cpp
+++ b/src/findindirs.cpp
@@ -1,6 +1,19 @@
 #include "findindirs.h"
 #include "utilsSystem.h"
 
+/*!
+ * \brief     Creates a search object with search modifiers.
+ * \param[in] mostRecent From all matching files return the most recent one.
+ * \param[in] checkReadable A matching file must be readable.
+ * \param[in] resolveDir An absolute directory path that is used to turn
+ * loaded search directories into absolute ones. For each loaded
+ * directory a check is made if it is absolute. Absolute directories
+ * are used unchanged while relative directories have resolveDir
+ * prepended to them.
+ * \param[in] dirs Optional search directories to load into the search object.
+ * Multiple directories should be separated by the platform-specific
+ * directory separator.
+ */
 FindInDirs::FindInDirs(bool mostRecent, bool checkReadable, const QString &resolveDir, const QString &dirs) :
 	m_mostRecent(mostRecent),
 	m_checkReadable(checkReadable),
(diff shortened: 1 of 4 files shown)


sunderme:
qce i.e. all qdoc* files are documented in the .cpp file , not in .h.
Maybe better to keep that scheme ...
This probably true for most of the other sources as well.

MeanSquaredError:
The documented methods
`QDocumentLineHandle::getCookie`
`QDocumentLineHandle::getCookieLocked`
etc.

are all defined in the `qdocumentline_p.h`, that is all of the method bodies are inside `qdocumentline_p.h`. So there is no .cpp with the method definitions that we could move the documentation to.

A couple of `QDocumentLine` methods have their method definitions in `qdocument.cpp`, so  I guess I could move the documented method bodies to `qdocument.cpp` and move the doxygen comments there too.

What should I do ?

sunderme:
Probably best to move the two definitions to qdocument.cpp then (in my opinion).
I really would like to have documentation in one place...

MeanSquaredError:
OK, I moved the cookie method definitions to `qdocument.cpp` and also moved the doxygen comments to `qdocument.cpp`.

Since this is the preferred documentation style for TXS, I also moved the same way the doxygen comments for the `FindInDirs` class.

sunderme:
The auto test fails on travis ci.
Can you check that everything is still okay ?

MeanSquaredError:
I checked the details here
https://travis-ci.org/texstudio-org/texstudio/builds/641239547?utm_source=github_status&utm_medium=notification
It seems that the build that errs is `QT=qt5`, while the other 4 builds are OK,
```
QT=qt5osx
QT=qt5NoPoppler
QT=qt5Release
QT=qt5win
```

I also checked the actual error message and it seems that there is no specific error, just a lockup of the build system:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

I remember that a while ago there were three different PRs that erred in a similar way (timeout during build). IIRC @timhoffm just restarted the builds and they passed the tests correctly. So maybe the easiest thing would be to restart it and see if that fixes the problem?

sunderme:
It stalls in the auto run execution.
Can you start texstudio -execute-tests on your build to be sure that there is not a locking error ?

MeanSquaredError:
You are right, it locks up on my computer too, while running the CodeSnippet test.
Could be a bug in the PR itself or could be that it is triggering some other bug.

It is ironic that the code change is so small and I did not see a chance for it breaking anything, so I did not bother to run the tests before submitting the PR.

I will investigate it this evening and will provide an update/fix for the PR.

sunderme:
I have found the issue.
my bug ...

MeanSquaredError:
Perfect timing - I got down to investigating the issue 15 minutes ago but was lucky to check my email and saw that you just solved the issue :-)

What is interesting is that it seems that moving the cookie method definitions outside of the header file triggered the lockup.

My hypothesis is that separating the cookie method definitions from the declarations prevented gcc from inlining the code in the debug builds thus slowing down the code in the debug builds slightly. Just enough for the race condition to happen.

Anyway, thanks for solving the issue!

Issue #888, opened 2020-01-16
Please test 3.0.0alpha1


sunderme:
Since there have been some fundamental changes on spell checking and syntax highlighting, please test this version.
In case of crashes , backtraces or a description what has been done, would be helpful.

foreachthing:
Any chance to get an up-to-date [changelog](https://github.com/texstudio-org/texstudio/blob/master/utilities/manual/CHANGELOG.txt)?
Currently, 2.12.20 is missing... and 3.00.00 has no entries at all. Thanks!

sergiokapone:
For  3.0.0alpha1 I found a bubs:
1. Comments inside tikzpicture environment highligted as tikz code (need to be a common comment)
2. tikz-keyword "node" spellchecked.

tmelorc:
In Linux Mint 18.
```
TeXstudio 3.00.00 (git n/a)
Using Qt Version 5.7.0, compiled with Qt 5.7.0 D
```

Open Terminal does not work. Nothing happens.

paulinsmaier:
Creating a new tex file inside a document, using \input{xyz} and right clicking with: xyz does not exist, should it be created leads to a crash

MeanSquaredError:
@paulinsmaier Which platform/OS did you try TXS on?

paulinsmaier:
> @paulinsmaier Which platform/OS did you try TXS on?

windows 10, version 1909, Build 18363.535

MeanSquaredError:
@paulinsmaier 
Does TXS create any kind of backtrace or crash dump file?

paulinsmaier:
> @paulinsmaier
> Does TXS create any kind of backtrace or crash dump file?

It should be called: texstudio_backtrace*.txt, right? Where should it be located? I did a search on all files, but did not find any

MeanSquaredError:
@paulinsmaier 
Yes, the backtrace file is called texstudio_backtrace*.txt
TXS uses the "temporary path" for that file for which the programming documentation says "On Windows this is usually the path in the TEMP or TMP environment variable".

So I would check :
1. The `c:\temp` or `c:\tmp` directory if your system has one.
2. Your home directory `c:\users\your_login_name`
3. The TXS appdata directory (same directory where `texstudio.ini` is located)

On a side note, yesterday sunderme committed a bugfix for a somewhat related issue, so maybe you could try the latest windows build and see if it fixes the crash that you observed:

[...]

PR #898, opened 2020-01-21
Fix the "Open Terminal" command under Linux

MeanSquaredError:
This PR fixes the bugreport for non-working terminal command ( https://github.com/texstudio-org/texstudio/issues/888#issuecomment-576333120 )

The problem was that `gsettings get org.gnome.desktop.default-applications.terminal exec` returns the command with a trailing line feed that we did not remove. This was fixed by calling .trimmed() on the returned command.


sunderme:
Thanks

tmelorc:
Thanks a lot. It works now.

jonasstein:
Thank you. Works fine on Gentoo with
```
TeXstudio 3.00.00 (git 3.0.0alpha6-1-g80b1f228)
Using Qt Version 5.15.0, compiled with Qt 5.15.0 R
```

Issue #897, opened 2020-01-20
[RFE] add \ar to tikz-cd.cwl and \tikztostart to tikz.cwl


chsk-uk:
Minor request for enhancement: please

* add `\ar` to `tikz-cd.cwl` (this is an alias for `\arrow`), and
* add `\tikztostart` to `tikz.cwl` (alongside `\tikztonodes` and `\tikztotarget`).

Thank you!

muzimuzhi:
I will add them (and maybe more) in two days. 

Currently, I am afraid that cwl files for `tikz`-related packages are more or less incomplete in both options and commands/environments.

chsk-uk:
Excellent. Thank you for the quick fix!

PR #900, opened 2020-01-23
cwl - update tikz-cd.cwl, solve #897

muzimuzhi:
Main changes
* Add usual commands and options to `tikz-cd.cwl`. 
* Add `\tikztostart` to `tikz.cwl`


sunderme:
thanks

PR #904, opened 2020-01-25
Mark LatexToken methods as const.

MeanSquaredError:
While working on the Latex code formatter I came across a problem - some of the LatexToken methods cannot be called on const objects even though the methods do not change the objects.

This PR fixes this issue.


MeanSquaredError:
While working on the Latex code formatter I came across a problem - some of the LatexToken methods cannot be called on const objects even though the methods do not change the objects.

This PR fixes this issue.

PR #921, opened 2020-02-05
Improvement to configure MiKTeX portable under Windows.

OgreTransporter:
If you use MiKTeX portable under Windows, no environment variables are created and MiKTeX is not necessarily installed in C:\MikTeX, i.e. TeXstudio cannot find it. The same applies if you have installed different versions of MiKTeX, because for some special cases only an old version works.

The patch adds a button to the configuration page under Windows. With this button you can then select pdflatex.exe in the MiKTeX directory and the other commands are automatically searched in the same directory.

Not interesting for Linux and macOS, but it makes work easier under Windows if you don't have to reset all commands manually.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1495,6 +1495,7 @@ void BuildManager::checkLatexConfiguration(bool &noWarnAgain)
 	}
 }
 
+#ifdef Q_OS_WIN32
 void BuildManager::swapWinSearchPath(QString &miktex, QString &texlive)
 {
 	QString oldMiktexBinPath = miktexBinPath;
@@ -1504,6 +1505,7 @@ void BuildManager::swapWinSearchPath(QString &miktex, QString &texlive)
 	miktex = oldMiktexBinPath;
 	texlive = oldTexliveWinBinPath;
 }
+#endif
 
 bool BuildManager::runCommand(const QString &unparsedCommandLine, const QFileInfo &mainFile, const QFileInfo &currentFile, int currentLine, QString *buffer, QTextCodec *codecForBuffer )
 {
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1495,6 +1495,16 @@ void BuildManager::checkLatexConfiguration(bool &noWarnAgain)
 	}
 }
 
+void BuildManager::swapWinSearchPath(QString &miktex, QString &texlive)
+{
+	QString oldMiktexBinPath = miktexBinPath;
+	QString oldTexliveWinBinPath = texliveWinBinPath;
+	miktexBinPath = miktex;
+	texliveWinBinPath = texlive;
+	miktex = oldMiktexBinPath;
+	texlive = oldTexliveWinBinPath;
+}
+
 bool BuildManager::runCommand(const QString &unparsedCommandLine, const QFileInfo &mainFile, const QFileInfo &currentFile, int currentLine, QString *buffer, QTextCodec *codecForBuffer )
 {
 	if (waitingForProcess()) return false;
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -3555,13 +3555,10 @@ void ConfigManager::updateManagedOptionObjects(ManagedProperty *property)
 		property->writeToObject(o);
 }
 
+#ifdef Q_OS_WIN
 void ConfigManager::searchLaTeX()
 {
-#ifdef Q_OS_WIN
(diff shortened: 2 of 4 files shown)


src/buildmanager.cpp
```diff
@@ -1495,6 +1495,18 @@ void BuildManager::checkLatexConfiguration(bool &noWarnAgain)
 	}
 }
 
+#ifdef Q_OS_WIN32
+void BuildManager::swapWinSearchPath(QString &miktex, QString &texlive)
```
sunderme: not sure that this special method is necessary useful.
You should have direct access to the miktexBinPath... (and the name is confusing: swap usually means exchanging the both given variables)

src/configmanager.cpp
```diff
@@ -3546,3 +3554,36 @@ void ConfigManager::updateManagedOptionObjects(ManagedProperty *property)
...
+			if (ql == NULL) continue;
+			QPushButton* pb = (QPushButton*)ql->widget();
+			if (pb == NULL) continue;
+			if (pb->property(PROPERTY_WIDGET_TYPE) == CG_PROGRAM) w = pb->property(PROPERTY_ASSOCIATED_INPUT).value<QWidget *>();
+		}
+		if (w == NULL) continue;
+		QString newCommandLine = tempCommands.value(getCmdID(w)).guessCommandLine();
+		if (!newCommandLine.isEmpty()) setText(w, newCommandLine);
+	}
+	buildManager->swapWinSearchPath(tmpMiktex, tmpTexlive);
```
sunderme: why is the path forgotten again ?

src/buildmanager.cpp
OgreTransporter: The purpose is, if a MiKTeX version was detected, to change the variable to the desired target path only for the duration of the manual search. After the search the path is changed back again (your second comment). This is helpful if you have installed a MiKTeX version regularly and are working with a portable version in parallel.

src/configmanager.cpp
OgreTransporter: To restore the original state before calling the function.

sunderme:
This sounds an awful lot like a solution for one user. Not many will use a portable miktex in parallel to an installed one (why would anyone?)
So, I can imagine an open dialog when the auto search fails, but for that specific purpose alone ?
You rather could change the search path for that run, maybe let the txs path take precedence over PATH...

OgreTransporter:
Adding a dialog box and asking for a MiKTeX path when no LaTeX can be found was my first thought. I could add that as well.

I chose this button to make it easier to switch between two versions. For example if the template from publisher A only works with MiKTeX 2.9.7000, but publisher B needs 2.9.7269.

The patch is designed not to change or affect the existing functions. Hence the double exchange of the variables.

Translated with www.DeepL.com/Translator (free version)

sunderme:
After thinking about the approach a bit, I will not go forward with this approach.
There is already an option for additional command path ($PATH)  (options/build/commands ($PATH)).
At the moment that path is added to the system $PATH, hence no miktex installation can be overruled.
But we could change that or you remove all miktex-installations from $PATH and select the active one only with the "additional path" setting.
The main advantage is that any command settings are untouched and the PR functionality is already there without more buttons.

sunderme:
another more elegant approach may be to manipulate "additonal path" via a macro script, changing between different installations on the fly

timhoffm:
IMHO the best and at the same time most generic approach would be to define "environments" as a set of common commands. Think of it as a kind of dropdown menu on the commands tab, which lets you choose a consistent set of commands.

This could be "MiKTeX 2.9.7000" or "MiKTeX 2.9.7269" or "TeXlive xyz"  or whatever the user needs. We could provide auto-detection for the common distributions and install locations, but the user would be able to additionally configure it completely as needed.

One can debate if the SVN commands should be included in the environment or not. I don't have a strong opinion on that.

OgreTransporter:
Ok, I have now set MiKTeX 2.9.7000. Now I have to change to 2.9.7269. How do I do this fast? As additional path I have now set C:\MiKTeX-portable\v2.9.7269\texmfs\install\miktex\bin\x64. And then? The commands are still all set to 2.9.7000. The function getMiKTeXBinPathInternal does not search in the additional directories but only in the global PATH variable.

A profile solution like TeXnicCenter works of course.

sunderme:
the idea is, *not* to set any path information in the commands (e.g. just latex ...) and finding the appropriate installation just via the additional path.

sunderme:
Since this approach will not be included, I close the PR.
We can discuss potential profile support via the issues, once it is clearer what is actually needed

Issue #918, opened 2020-02-04
Question about LatexDocument::checkNextLine


MeanSquaredError:
Hi,

I am looking at the latest code of `SyntaxCheck` and one thing caught my attention.
It seems that in `SyntaxCheck::run` once it processes one line it optionally emits the signal `checkNextLine` which is handled by `LatexDocument`. In turn `LatexDocument::checkNextLine` processes the signal and then calls `SyntaxCheck::putLine` to queue the next line for processing.

As far as I see `LatexDocument::checkNextLine` does not really need to access the members of the `LatexDocument` object except for `LatexDocument::unclosedEnv` and the latter data member is not really dependent in any way on the `LatexDocument` class.

Is there a reason why `LatexDocument::checkNextLine` is not moved into `SyntaxCheck`? I think that this will make the code simpler and will also make it easier to clear some nasty race conditions in the tests in `SyntaxCheckTest`. Currently the tests in SyntaxCheckTest don't work very reliably especially when some other program loads the CPU during the tests. There is no easy way to fix that because due to the asyncronous nature of the `checkNextLine` event we cannot reliably catch the moment when `SyntaxCheck` has finished the processing of the queued lines.

I would expect that removing `checkNextLine` will simplify things a bit. So is there something (e.g. a design decision) that I am missing which forces the use of this separate signal?

sunderme:
`int index = indexOf(dlh,hint);` is the function of document. 
what race condition ?

MeanSquaredError:
`LatexDocument::indexOf` is inherited from `QDocument::indexOf`
On the other hand every `QDocumentLineHandle` object has a member `QDocument::m_doc` which points to the owner document, so we can always call `QDocument::indexOf` through the `QDocumentLineHandle` object alone.

In fact there is the convenience method `QDocumentLineHandle::line()` which delegates the call to `QDocument::indexOf`.

So I would say that we should be OK to move checkNextLine from `LatexDocument` into `SyntaxCheck` (unless I am missing something else too).

Regarding the race condition:
The code that we use to wait for the `SyntaxCheck` to finish is 
```
do{
        edView->document->SynChecker.waitForQueueProcess(); // wait for syntax checker to finish (as it runs in a parallel thread)
        QApplication::processEvents(QEventLoop::AllEvents,10); // SyntaxChecker posts events for rechecking other lines
}while(edView->document->SynChecker.queuedLines());
```
[...]

PR #922, opened 2020-02-05
Fix tests that use SyntaxCheck

MeanSquaredError:
As mentioned in https://github.com/texstudio-org/texstudio/issues/918 there is a race condition in the `SyntaxCheck::waitForQueueProcess`. This race condition causes any tests that use this method to return prematurely. This happens mostly when the CPU load is high and when it happens tests start failing like this:

```
PASS   : SyntaxCheckTest::checkkeyval(key/value, with -, error key)
FAIL!  : SyntaxCheckTest::checkkeyval(key/value,error value) '(!formats.isEmpty())==(error)' returned FALSE. (equal failed: got "0" !=expected "1" )
   Loc: [src/tests/syntaxcheck_t.cpp(267)]
FAIL!  : SyntaxCheckTest::checkkeyval(key/value, with -, error val) '(!formats.isEmpty())==(error)' returned FALSE. (equal failed: got "0" !=expected "1" )
   Loc: [src/tests/syntaxcheck_t.cpp(267)]
PASS   : SyntaxCheckTest::checkkeyval(2 key/value)
FAIL!  : SyntaxCheckTest::checkkeyval(2 key/value,error key) '(!formats.isEmpty())==(error)' returned FALSE. (equal failed: got "0" !=expected "1" )
   Loc: [src/tests/syntaxcheck_t.cpp(267)]
FAIL!  : SyntaxCheckTest::checkkeyval(2 key/value,error value) '(!formats.isEmpty())==(error)' returned FALSE. (equal failed: got "0" !=expected "1" )
   Loc: [src/tests/syntaxcheck_t.cpp(267)]
PASS   : SyntaxCheckTest::checkkeyval(2 key/value,composite key)
PASS   : SyntaxCheckTest::checkkeyval(yathesis without options)
PASS   : SyntaxCheckTest::checkkeyval(yathesis with valid option)
FAIL!  : SyntaxCheckTest::checkkeyval(yathesis with ilegal option) '(!formats.isEmpty())==(error)' returned FALSE. (equal failed: got "0" !=expected "1" )
   Loc: [src/tests/syntaxcheck_t.cpp(267)]
PASS   : SyntaxCheckTest::checkkeyval(article)
```
The above errors are just an example but in general it affects the tests in
`SyntaxCheckTest::checkkeyval`
`SyntaxCheckTest::checktabular`
`SyntaxCheckTest::checkArguments`
`LatexEditorViewTest::inMathEnvironment`

I have described in https://github.com/texstudio-org/texstudio/issues/918 what exactly is causing the race condtion.

The proposed PR fixes the issue by making the following changes:
1. Replaces `SyntaxCheck::stopped` with a command message dequeue and a state variable.
2. The command dequeue allows sending three different commands to the `SyntaxCheck` worker thread: PAUSE, UNPAUSE, STOP.
3. The pause/unpause commands allow pausing/unpausing of the worker thread which in turn allows avoiding the race condition that happens with the existing code.

There is also another minor multithreading issue with the current code - currently `SyntaxCheck::stop` is a variable shared between the different threads, but it is not protected. Ideally it should be either protected by a lock or even better marked as atomic. I don't think that the non-atomic variable really causes any issues because any modern processor handles word-sized variables in an atomic way, but it is still better to be on the safe side and mark them explicitly as atomic. So the PR makes the replacement variable SyntaxCheck::mState of `QAtomicInt` type.

I tried overloading the CPU while running tests the modified version of `SyntaxCheck` and could not reproduce the test errors. So the race condition is either fixed or just better hidden :-)

### DIFF
--- a/src/syntaxcheck.cpp
+++ b/src/syntaxcheck.cpp
@@ -25,6 +25,7 @@ SyntaxCheck::SyntaxCheck(QObject *parent) :
 	mLinesLock.lock();
 	stopped = false;
 	mLines.clear();
+	mLinesEnqueuedCounter.fetchAndStoreOrdered(0);
 	mLinesLock.unlock();
 }
 
@@ -59,6 +60,7 @@ void SyntaxCheck::putLine(QDocumentLineHandle *dlh, StackEnvironment previous, T
     newLine.hint=hint;
 	mLinesLock.lock();
 	mLines.enqueue(newLine);
+	mLinesEnqueuedCounter.ref();
 	mLinesLock.unlock();
 	//avoid reading of any results before this execution is stopped
 	//mResultLock.lock(); not possible under windows
@@ -272,21 +274,28 @@ void SyntaxCheck::setFormats(QMap<QString, int> formatList)
 /*!
 * \brief wait for queue to be empty. Used for self-test only.
 */
-void SyntaxCheck::waitForQueueProcess()
+void SyntaxCheck::waitForQueueProcess(void)
 {
-    wait(1);
-	while (!crashed && mLinesAvailable.available() > 0) {
-		wait(1);
-	}
-}
+	int linesBefore, linesAfter;
 
-/*!
-* \brief check if queue is empty. Used for self-test only.
-* \return queue is not empty
-*/
-bool SyntaxCheck::queuedLines()
-{
-	return mLinesAvailable.available() > 0;
+	/*
+	 * The logic in the following loop is not perfect because it could terminate the loop too early if it takes more
+	 * than 10ms between the call to mLinesAvailable.acquire() and the following call to mLinesAvailable.release().
+	 * Implementing the check properly requires bi-directional communication with the worker thread with commands to
+	 * pause/unpause the worker thread which complicates the code too much just to handle testing.
+	 */
+	linesBefore = mLinesEnqueuedCounter.fetchAndAddOrdered(0);
+	forever {
+		for (int i = 0; i < 2; ++i) {
+			QApplication::processEvents(QEventLoop::ExcludeUserInputEvents | QEventLoop::ExcludeSocketNotifiers, 1000); // Process queued checkNextLine events
+			wait(5); // Give the checkNextLine signal handler time to queue the next line
+		}
+		linesAfter = mLinesEnqueuedCounter.fetchAndAddOrdered(0);
+		if ((linesBefore == linesAfter) && !mLinesAvailable.available()) {
+			break;
+		}
+		linesBefore = linesAfter;
+	}
 }
 
 /*!
--- a/src/syntaxcheck.h
+++ b/src/syntaxcheck.h
@@ -101,12 +101,11 @@ public:
 	void stop();
 	void setErrFormat(int errFormat);
     QString getErrorAt(QDocumentLineHandle *dlh, int pos, StackEnvironment previous, TokenStack stack);
-	void waitForQueueProcess();
+	void waitForQueueProcess(void);
 	static int containsEnv(const LatexParser &parser, const QString &name, const StackEnvironment &envs, const int id = -1);
 	int topEnv(const QString &name, const StackEnvironment &envs, const int id = -1);
 	bool checkCommand(const QString &cmd, const StackEnvironment &envs);
 	static bool equalEnvStack(StackEnvironment env1, StackEnvironment env2);
-	bool queuedLines();
 
 	void setLtxCommands(const LatexParser &cmds);
     void setSpeller(SpellerUtility *su);
@@ -126,6 +125,7 @@ private:
 	QQueue<SyntaxLine> mLines;
 	QSemaphore mLinesAvailable;
 	QMutex mLinesLock;
+	QAtomicInt mLinesEnqueuedCounter; //!< Total number of lines enqueued from beginning. Never decremented.
 	bool stopped;
     int syntaxErrorFormat;
 	LatexParser *ltxCommands;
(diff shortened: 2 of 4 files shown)


MeanSquaredError:
Regarding `3. I am not convinced that you really need a pause.`

I guess I could remove the pause/unpause, but I cannot think of a better way to workaround the race condition. How would we handle the case when the SyntaxCheck thread is past the `mLinesAvailable.acquire()`, so `mLinesAvailable.available()` returns zero and `SyntaxCheck::waitForQueueProcess` does the check exactly at this moment?

Do you have any specific suggestion that would prevent this ?

One much simpler solution would be to ditch the proposed code changes and just increase the current `wait()` in `SyntaxCheck::waitForQueueProcess` from 1ms to 5ms. That would not remove the race condition but it would pretty much resolve it for any modern computer and since it in just in the test code I guess it is not a big deal.

If you want I can change the PR so that it will do just that.

MeanSquaredError:
Actually it the race would almost never happen if `SyntaxCheck::waitForQueueProcess` would do in loop
1 `QApplication::processEvents(...)`
2. wait(5)
3. Check if there are queued lines.

That would pretty much resolve it because 5 ms should be enough for the next line to be queued on any moderately fast computer. However processing events should precede the wait because otherwise the pending `checkNextLine` signal would hang in the event queue.

So if that hack is OK with you, I can replace the current code in the PR with it.

sunderme:
I would rather go for reference counting, i.e. count all outstanding checkNextLine (and reduce the number when putLine is executed)
Just make sure that the number is adapted even if checkNextLine does not actually put the line into the syntax checker.

sunderme:
but your suggestion probably works as well and it is only for testing ...

MeanSquaredError:
I am not sure that I understand correctly your suggestion. Do you mean that:

Do you mean that we should :

1. Have an extra variable for pending `checkNextLine`, e.g. `int cnlPending`. This variable should be increased before `emit checkNextLine` and then decreased before return from `LatexDocument::checkNextLine`
2. In `SyntaxCheck::waitForQueueProcess` we should check two things, the number of queued lines (`mLinesAvailable.available()`) and the number of lines pending for `checkNextLine`.

If that is the suggestion, then there are a few problems that I can see:

1. When should we really increase `cnlPending` ? If it is just before `emit checkNextLine`, then we still have a window where the race condition will happen. If `SyntaxCheck::waitForQueueProcess` does the check after `SyntaxCheck` worker has done `mLinesAvailable.acquire()` but before we do `++cnlPending`, then it will decide incorrectly that the `SyntaxCheck` work has been completed.
Even if we move the `++cnlPending` just after `mLinesAvailable.acquire()`, the race could theoretically still happen, so we would need a very special custom kind of semaphore that would do the `acquire()` and `++cnlPending` atomically, which is a fair amount of code.

2. We don't really know how many `LatexDocument` objects handle the `checkNextSignal` event. If there are multiple documents handling the event we should take care to only `--cnlPending` for lines that are actually destined for the specific `LatexDocument`. However in this case it is not clear how to handle lines that have already been deleted (e.g. the owner document has been closed while the line is processed by `SyntaxCheck`). In that case there would be no owner document to actually do `--cnlPending`.

3. If we really want to do things perfectly reliably we should check atomically for the enqueued lines and and pending `checkNextLine` lines. So all access to the semaphore `mLinesAvailable` and `cnlPending` should be protected by a lock. Not a unsolvable problem just requires thinking because it is really easy to get it wrong and get a race or deadlock :-)

MeanSquaredError:
I tried all kinds of ideas that include counting but they either don't work or end up more complex than pause/unpause. In most cases both :-)

MeanSquaredError:
So maybe just increase the wait() and move the event processing in the loop?

sunderme:
sounds okay

MeanSquaredError:
OK, I changed the PR to use the simpler strategy :

1. Have an ever-increasing counter of enqueued lines `SyntaxCheck::mLinesEnqueuedTotal`
2. In `waitForQueueProcess` do the following in a loop:
2.1. Remember the current number of total enqueued lines.
2.2. Process events and then wait for 5 ms to give the worker thread time to enqueue the next line.
2.3. If the number of enqueued lines has changed retry the wait loop.

I tried overloading the computer while it was running the tests and the wait loop seems to work correctly. So hopefully it will not break again. Well, at least not too often.

MeanSquaredError:
Ironically the `SyntaxCheck` test failed on the server this time even though it works perfectly on my computer. OK, let me see how to fix that.

MeanSquaredError:
OK, after the update tests seem to pass on the server.

MeanSquaredError:
@sunderme
After a small update in the code the `QT=qt5NoPoppler` build failed with
`The command "${TRAVIS_BUILD_DIR}/travis-ci/get-dependencies.sh" failed and exited with 1 during .`
I suspect that it is an internal failure of the build system unrelated to the code.
Can you please restart the build?

sunderme:
I don't have a button to restart a pull-request run. But the fail can be ignored. I will merge the code tonight. Fri Feb 07 10:28:22 GMT+01:00 2020 texstudio-org/texstudio <reply@reply.github.com>: @sunderme After a small update in the code the QT=qt5NoPoppler build failed with The command "${TRAVIS_BUILD_DIR}/travis-ci/get-dependencies.sh" failed and exited with 1 during . I suspect that it is an internal failure of the build system unrelated to the code. Can you please restart the build? — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

MeanSquaredError:
Great. Thank you.

PR #934, opened 2020-02-13
add file extension "synctex.gz" to cleandialog

muzimuzhi:
`*.synctex.gz` files are the default aux file created by `synctex` program, when `*.tex` files are compiled with option `-synctex=1`, which is set by default in texstudio.

This pr adds `synctex.gz` to the list of file extensions shown in "Clean Auxiliary File" dialog.

### DIFF
--- a/src/cleandialog.cpp
+++ b/src/cleandialog.cpp
@@ -4,7 +4,7 @@
 #include "configmanager.h"
 #include "utilsUI.h"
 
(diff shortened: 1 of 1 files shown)


timhoffm:
We intentionally did not include synctex.gz. As opposed to other auxiliary files it is needed beyond the compilation process. As soon as you delete it, syncing between pdf and tex does not work anymore in TeXstudio. Therefore, I prefer to not delete this by default.

muzimuzhi:
Reasonable.

Issue #889, opened 2020-01-17
Secondary toolbar change scale problem


jcardenast:
### Environment
- TeXstudio: 2.12.20-win
- Qt:  QT5
- OS: Windows  8 and 10
- TeX distribution:  miktex 2.9

### Expected behavior
on Options -> Scale Gui -> secondary toolbar, the secondary bar size change as new configuration

### Actual behavior
on Options -> Scale Gui -> secondary toolbar, the secondary bar size change but when TexStudio restart, the bar continues with the same default size.

### How to reproduce
Change the Secondary toolbar size and apply, it works in this time, but if you close and open TexStudio this bar not apply the change. It refreshes on Entering options and click accept and it change the size.

A similar issue is on the preview document window, when all bars size looks ok, if you are watching a document and enter to Options menu and accept, in one second the icons on the preview menu resize to bigger one.

foreachthing:
Huh?! I cannot confirm in TXS 2.12.20b1 (Win 10 x64).

sunderme:
can't confirm either

jcardenast:
I updated and the problem still continue. I attached this images, the first show the size of the icons (I make smaller this items before but in the start they don't change). Next I only enter to Options and click Ok without changing anything. Finally, the window update the icons to the size that I modified before. This behaviour is the same all the time that I enter to Tx.

![T1](https://user-images.githubusercontent.com/12958198/72685456-3b35da80-3ab8-11ea-9cda-124bdc23d5a5.png)
![T2](https://user-images.githubusercontent.com/12958198/72685458-3d983480-3ab8-11ea-9dda-4afd6637190b.png)
![T3](https://user-images.githubusercontent.com/12958198/72685459-40932500-3ab8-11ea-87ad-18d7c200f3a9.png)

foreachthing:
Got two computers, same specs as you do ... cant reproduce your issue on neither machine.

Maybe completely uninstall TXS (also the config dir; where texstudio.ini is) an reinstall it. Make a backup or rename your config-dir.

MeanSquaredError:
Does TXS preserve the other settings when you change them and then exit TXS and start it again?
For example does it keep General / Font Size if you set a non-default value and then restart it?

sunderme:
no feedback

jcardenast:
> Does TXS preserve the other settings when you change them and then exit TXS and start it again?
> For example does it keep General / Font Size if you set a non-default value and then restart it?

[...]

PR #939, opened 2020-02-15
Do not set a minimum size of the secondary toolbar icons

MeanSquaredError:
This PR fixes https://github.com/texstudio-org/texstudio/issues/889
I could not reproduce the issue on Windows using the screen resolution and scaling on the user's computer. However removing the `qMax()` from the code fixed the issue for the user (he tried a custom version of TXS compiled without the toolbar size limitation).

Since we don't have that `qMax()` limitation in the other places where we update the secondary toolbar size, I guess removing it from the code that reads the initial settings shouldn't cause problems.


sunderme:
thanks

PR #941, opened 2020-02-15
Export the parser tokens to user scripts

MeanSquaredError:
This PR implements the C++ part of the TeX source formatter that we talked about a while ago.

More specifically makes the following changes:

1. Exports `QDocument::getLineTokens()` to QtScript. This function allows user scripts to see the parser tokens created by `latexDetermineContexts2` for a given line.
2. Exports the enum `Token::TokenType` to QtScript. The enum is exported as an object called `latexTokenType` where each property of that object corresponds to one token type.
3. Adds a test called `QDocumentTest::getLineTokens()` for the new method.
4. Documents the method `QDocument::getLineTokens()` and the enum object `latexTokenType` in the user manual.

I am still testing and fixing the QtScript counterpart of the TeX code formatter but it should be ready in a few days.

### DIFF
--- a/src/latexdocument.h
+++ b/src/latexdocument.h
@@ -90,6 +90,7 @@ public:
     friend class LatexEditorViewTest;
 	friend class LatexStructureMerger;
 	friend class LatexStructureMergerMerge;
+	friend class ScriptEngineTest;
 
 private:
 	static QStringList someItems(const QMultiHash<QDocumentLineHandle *, ReferencePair> &list);
--- a/src/latexparser/latextokens.h
+++ b/src/latexparser/latextokens.h
@@ -6,6 +6,36 @@
 class QDocumentLineHandle;
 class QDocument;
 
+/*
+ * \brief Wrapper for the TokenType enum.
+ * \details This class contains all the TokenType enums. It is has two purposes:
+ * 	1. Provide enums for Token class.
+ * 	2. Generate a MOC object that is used to provide these enums to the script engine.
+ * The enums should really be a namespace, but Q_NAMESPACE is only supported in Qt5.8+
+ */
+class EnumsTokenType
+{
+	Q_GADGET
+	Q_ENUMS(TokenType)
+
+public:
+	enum TokenType {
+		none = 0, word, command, braces, bracket,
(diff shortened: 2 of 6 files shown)


MeanSquaredError:
OK, the build failed. Seems some problem with the new Qt. Now to fix it...

sunderme:
I am not thrilled about getLineTokens being a part of QDocument as a QtScript special function.
I would rather see it in scriptengine.cpp, similarly implemented like "search"
`extendedQEditor.setProperty("search", engine->newFunction(&searchFunction), QScriptValue::ReadOnly | QScriptValue::Undeletable);`

You can bet that the function in QDocument will be forgotten in the next script refactoring ...

MeanSquaredError:
I made `getLineTokens` a part of `QDocument` since it gets the tokens for a given document/line and not just for the document in the editor.
Creating an extended script object for every `QDocument` (like we do for the editor) would complicate the code too much (we have only one editor but multiple documents and we can access them either through the editor or through the documents manager), so I just moved it to `QDocument`.

So the easiest way to have `getLineTokens` in `scriptengine.cpp` would be to move it to `extendedEditor`. It doesn't really affect the tex source formatter since it works with the active document only, but I wanted to have it accessible for all documents (and not just the active one) just in case someone needs it in the future.

But I will prepare the changes (moving `getLineTokens` to `extendedEditor`) either this evening or tomorrow morning.

sunderme:
I don‘t really care to which object the function is bound in the script objects, only that a similar approach is probably useful.

MeanSquaredError:
OK, I updated the PR as follows:

1. `getLineTokens` remains a property of `editor.document()`, but it is implemented as a part of `extendedQDocument` (similar to `extendedQEditor`) in `scriptengine.cpp`.
2. The tests for `getLineTokens()` are implemented as a part of the scripting engine tests. I could not find a way to get the script exit/return value from `scriptengine`, in order to check the returned token list values. So the test code just throws an error in the JavaScript code. So if the tokens are different it will just open an alert popup and the test will timeout. Not perfect but it works.

sunderme:
okay, looks reasonable.
We can see whether the test can be improved.

PR #943, opened 2020-02-16
cwl - update "minted.cwl" to minted v2.5

muzimuzhi:
Update `minted.cwl` so that it contains every user command and option provided by `minted` v2.5.


sunderme:
thanks

PR #946, opened 2020-02-19
partly fix #277

### LINKED ISSUE #277: Allow math commands in macro argument
I am currently writing a `.cwl`-file for a custom package in which I specify a `\unit` macro. In the `.cwl` I write `\unit[quantity]{unit}` and the auto-complete works fine with that. However the `<unit>`-part will be typeset in math mode and therefore constructs like `\unit[3]{m^{-1}}` are perfectly valid and won't produce any errors.

The problem is: TS complains about the `^` being used outside of math-mode. Is there some cwl-magic I can use to convince TS that the mandatory macro argument is going to be in math-mode?

SantiagoBautista:
In the cwl format, add a '%formula' suffix
to allow users to specify that a field with an
arbitrary name should be treated as a math field.


sunderme:
thanks

PR #948, opened 2020-02-19
Document cwl arg names "formula" and "%formula"

muzimuzhi:
In cwl file, 
 - support for argument name `formula` is added by https://github.com/texstudio-org/texstudio/commit/5dd68d835473083f9fce12b044fe335533ae03bc
 - support for argument name ends with `%formula` is recently added by pr #946

This pr documents the above change into the user manual, and makes the fix for issue #277 more complete.


sunderme:
Thanks

PR #950, opened 2020-02-20
Minor cleanup in latexparsing.cpp

MeanSquaredError:
While working on the .tex code formatter I came across a few minor issues in latexparsing.cpp

1. When we create a line cookie of type `LEXER_COMMENTSTART_COOKIE`, the comment type by default should be `Token::unknownComment`. Currently the code sets it to -1. It is getting away with that because we never directly check for `unknownComment` (or `-1` for that matter) anywhere else in the code. Still I think it is better to use a meaningful enum since it makes clearer the meaning of the variable.
2. A few typos in the C++ comments that I noticed.


sunderme:
thanks

PR #955, opened 2020-02-22
cwl - change "tikz.cwl"

muzimuzhi:
1. Allow cmds like `\draw` to be used both inside `tikzpicture` env and
   after `\tikz`
2. Unset syntax checking for `#keyvals` of `\usetikzlibrary`, since
   package `tikzmark` provides no sty file but a tikz library file.

### DIFF
--- a/completion/tikz.cwl
+++ b/completion/tikz.cwl
@@ -1,5 +1,5 @@
 # tikz package
-# muzimuzhi/12 Nov 2019
+# muzimuzhi/12 Nov 2019, 23 Feb 2020
 
 #include:pgfcore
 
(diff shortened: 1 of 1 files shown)


sunderme:
you are sure that draw etc. should not have any limitation on validity?

muzimuzhi:
`\draw` can be used both as
```tex
\begin{tikzpicture}
  \draw ... ;
\end{tikzpicture}
```
and
```tex
\tikz \draw ... ;
% or
\tikz { \draw ... }
```
The deletion of `#/tikzpicture` in `tikz.cwl` is for the second kind of usage.

sunderme:
is the second use case realistic/ commonly used  ?

muzimuzhi:
Yes, it is introduced in [documentation of tikz v3.1.5b](https://github.com/pgf-tikz/pgf/releases/download/3.1.5b/pgfmanual.pdf), sec. 2.3 and 12.2.2.

Most of the time, I use `tikzpicture` environment. But since there are many `\tikz` use examples throughout the doc of tikz, users may also use the `\tikz \draw ...` case.

sunderme:
okay then

thanks

muzimuzhi:
A simple `grep` on the `doc` sub-directory of repo https://github.com/pgf-tikz/pgf on [latest commit](https://github.com/pgf-tikz/pgf/tree/15c943b7a5774c58e45390acdddd36e9f4260607) shows that
```bash
$ grep -rni '\\tikz[^a-zA-Z|@]' . | wc -l 
    1169
$ grep -rni '\\begin{tikzpicture}' . | wc -l
    1112
```

So, in the doc of tikz, the use cases of `\tikz ...` are as common as `tikzpicture` environment.

PR #958, opened 2020-02-23
Preserve text indentation when pasting script code

MeanSquaredError:
TXS has a minor bug. When pasting script code into the `Macros/Edit Macros.../LaTeX Content` editor control it will strip all leading whitespace thus removing any script code indentation. This behavior is a problem when the script is large and indentation is important.

This PR changes the corresponding `QEditor` behavior so that it preserves indentation.


sunderme:
thanks

PR #961, opened 2020-02-26
Fix build warnings for obsolete Qt methods

MeanSquaredError:
This PR fixes some of the build warnings for Qt 5.11.0+ with GCC9
It replaces calls to the deprecated method `QFontMetrics::width()` with calls to a custom function
`UtilsUi::getFmWidth()`. The latter function calls:
1. `QFontMetrics::horizontalAdvance()` on Qt 5.11.0+
2. `QFontMetrics::width()` before Qt 5.11.0
This fixes about 40-50% of all the build warnings.

The PR also fixes a `-Wdeprecated-copy` warning when assigning to `BibTeXType`

There are still quite a few build warnings (mostly calling Qt obsolete methods) but I will provide separate PRs once I find clean ways to fix those warnings.

### DIFF
--- a/src/utilsUI.cpp
+++ b/src/utilsUI.cpp
@@ -114,7 +114,7 @@ QToolButton *createComboToolButton(QWidget *parent, const QStringList &list, con
 		QString text = list[i];
 		//QIcon icon = (i<icons.length()) ? icons[i] : QIcon();
 		QAction *mAction = mMenu->addAction(text, receiver, member);
-		max = qMax(max, fm.width(text + "        "));
+		max = qMax(max, getFmWidth(fm, text + "        "));
 		if (i == defaultIndex) {
 			combo->setDefaultAction(mAction);
 			defaultSet = true;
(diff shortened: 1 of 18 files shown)


src/utilsUI.cpp
```diff
@@ -359,4 +359,37 @@ void resizeInFontHeight(QWidget *w, int width, int height)
...
+
+/*!
+ * \brief Given font metrics return pixel size of a text string
+ * \param[in] fm Font metrics
+ * \param[in] text Text string
+ * \param[in] len Only calculate width of the first len characters of the string. If not specified,
+ * then -1 is assumed which means calculate width of the whole string.
+ * \returns Returns pixel size of the text string
+ */
+int getFmWidth(const QFontMetrics &fm, const QString &text, int len)
```
sunderme: len is unused ?

MeanSquaredError: It is my omission. `len` should be an argument to `fm.horizontalAdvance`/`fm.width`. I fixed that and force-pushed the new version.

Ironically I was fixing warnings and did not notice my own bug that introduced a new warning. I guess I shouldn't write code while I am falling asleep.

sunderme:
Okay, looks good.

Thanks

PR #962, opened 2020-02-27
Replace calls to obsolete screen geometry methods.

MeanSquaredError:
This PR provides the next set of fixes for build warnings. It does the following two changes:

1. It replaces the obsolete calls to `QDesktopWidget::screenGeometry()` and `QDesktopWidget::availableGeometry()` with calls to methods of `GuiApplication::primaryScreen()`
2. It adds a custom function `UtilsUi::getAvailableGeometryAt()` which is used to replace calls to
`QDesktopWidget::availableGeometry(const QPoint &)`


sunderme:
thanks

PR #967, opened 2020-03-01
Fix Qt5.13.2 build warnings (final batch)

MeanSquaredError:
This PR contains the last batch of fixes for the Qt5.13.2 build warnings. With it applied TXS+Qt5.13.2 build without any warnings.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1578,7 +1578,7 @@ bool BuildManager::runCommandInternal(const ExpandedCommands &expandedCommands, 
 		p->subCommandName = cur.parentCommand;
 		p->subCommandPrimary = expandedCommands.primaryCommand;
 		p->subCommandFlags = cur.flags;
-		connect(p, SIGNAL(finished(int)), SLOT(emitEndRunningSubCommandFromProcessX(int)));
+		connect(p, SIGNAL(finished(int, QProcess::ExitStatus)), SLOT(emitEndRunningSubCommandFromProcessX(int)));
 
 
 		p->setStdoutBuffer(buffer);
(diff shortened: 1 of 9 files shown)


src/buildmanager.cpp
```diff
@@ -2322,7 +2322,10 @@ void ProcessX::startCommand()
 	if (cmd.startsWith("txs:///")) {
 		onStarted();
 		emit startedX();
+#pragma GCC diagnostic push
```
sunderme: I guess, this only silences the warning.
I assume that the warning is there because it will be removed with Qt6, so just silencing the warning does not solve anything ...

MeanSquaredError: I guess, I could replace the one-argument form of `finished` with the two-argument form in all the places where signal slots are defined and also where the signal is emitted. Would that be an acceptable solution?

sunderme: i think that qt can handle connects from 2 arg signals to one arg slots, so removing the one offending emit is probably enough.

MeanSquaredError: If I understand correctly the Qt documentation here
https://doc.qt.io/qt-5/signalsandslots.html
```
The signals and slots mechanism is type safe: The signature of a signal must match the signature of the receiving slot. (In fact a slot may have a shorter signature than the signal it receives because it can ignore extra arguments.) Since the signatures are compatible, the compiler can help us detect type mismatches when using the function pointer-based syntax.
```
So it seems to me that a slot with fewer arguments can indeed connect to a signal with more arguments, but the version of `connect()` that we use does the coupling between signals and slots at compile time.

If I understand correctly the inheritance of signals in Qt 5, if someone uses our class `ProcessX` with Qt5 and connects to its `finished(int)` signal, then the caller of `connect()` will connect to the one-argument version of `finished(int)` which is inherited from `QProcess`. So if we just remove the `emit finished(0);` for `txs:///` commands, then the connected code will wait for the one-argument version of `finished` and will never get it because we will only send the two-argument signal.

The picture would be different under Qt 6 where the one argument slot would always connect to the two-argument signal simply because there will be no one-argument signal.

So I think that for the code to work correctly under Qt 5 we should :
1. Remove the `emit finished(0);`
2. Change all of our `finished()` slots so that they use the two-argument version of the signal.

I am missing something from the picture above?

sunderme: Something like
 `connect(procx,SIGNAL(finished(int,QProcess::ExitStatus)),target,SLOT(finished(int)))`

should do it.

MeanSquaredError: Well, I also meant changing the actual signature of the function e.g. from
`void finished(int code);` to `void finished(int code, QProcess::ExitStatus status);`
but your suggestion is simpler. I will update the PR shortly.

MeanSquaredError: OK, I force-pushed the updated PR

sunderme:
thanks

PR #985, opened 2020-03-19
Autoscrolling on selecting beyond viewport

axkibe:
Autoscrolls when selecting beyond the viewport.


sunderme:
looks good.
Thanks

PR #988, opened 2020-03-20
Raise maximum number of macros

Socob:
Currently, the maximum number of macros that can be created is hard-coded as 1000. This change introduces a central constant for this number and raises it to 5000. Perhaps this limit should also be documented somewhere?

In case you’re wondering what the use case for this could be, I’m using macros to enable easy input of Unicode characters (e.g. typing “\psi ” triggers a macro and gives “ψ”), mostly for the purposes of `unicode-math`. Since there are a lot of Unicode characters (see [`unicode-math-table.tex`](https://github.com/wspr/unicode-math/blob/master/unicode-math-table.tex)), it can be quite easy to reach the current limit of 1000 macros at one macro per character. Now, perhaps there is another way to do this which requires fewer macros, but the restriction to 1000 is arbitrary/undocumented, so this seems like the simplest way to accommodate users who try something similar as I did.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -336,6 +336,7 @@ bool ManagedProperty::readFromObject(const QObject *w)
 }
 #undef READ_FROM_OBJECT
 
+const int ConfigManager::MAX_NUM_MACROS;
 QTextCodec *ConfigManager::newFileEncoding = nullptr;
 QString ConfigManager::configDirOverride;
 bool ConfigManager::dontRestoreSession=false;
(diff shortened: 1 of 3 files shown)


timhoffm:
Thanks. It’s reasonable to put that into a constant. I don‘t think anybody should need more than 1000 macros, but there‘s no harm in slightly increasing that limit.

For your particular use case, you can do everything within a single macro and use a regular expression for triggering. See e.g. https://github.com/texstudio-org/texstudio/wiki/Scripts#replace-characters-by-their-latex-equivalent-while-typing.

Socob:
@timhoffm Thanks, that works – I suspected there had to be a more elegant way!

PR #993, opened 2020-03-22
Request for Comment: A Terminal for TeXStudio

axkibe:
Screenshot: https://i.imgur.com/4GSusOF.png

So I made a working draft to add qtermwidget5 to TeXStudio. So far the patch is surprisingly light weighted. It does create a dependency on the library providing the widget however (option in BUILD.sh). One could argue about copying over the code for further adaption, but then it would have to be maintained / kept in sync.

One thing that puzzles me somewhat, is how to handle key combinations that are already bound to QActions when in terminal, like for example function keys. right now I did nothing. On the hand pausing them seems hard to code, on the other hand I wonder if this is even wanted. Hitting F1 for example in the normal shell should be quick build anyway, but if you say are in Midnight Commander, you might want it's help, and it's pretty hard to guess in which state the shell is in.

Also I didn't add a new tab in the config dialog, where one may want to change the font in the terminal or the used color scheme.

I'm not 100% happy how the #ifdefs worked out, adding/removing the whole class as #ifdef get's the linker making hiccups due to some .moc magic when enabling/disabling the terminal via BUILD.sh

Also this is tested for Linux only, qtermwidget ought to work for BSD and OSX too, but not for Windows. However, I doubt that Windows users want a terminal anyway :)

### DIFF
--- a/BUILD.sh
+++ b/BUILD.sh
@@ -84,6 +84,11 @@ else
 	OPTION_PHONON=no
 fi
 
+if [ "$SYSTEM" = 1 ]; then
+	readoption "Do you want a terminal widget (requires the qtermwidget5 library)?" yes;
+	OPTION_TERMINAL=$NEWVALUE
+fi
+
 readswitchy "Do you want to build a debug or release version?" "debug release d r deb rel" debug;
 case "$NEWVALUE" in
 	d|deb|debug) OPTION_DEBUG_BUILD=yes;;
@@ -109,6 +114,7 @@ fi
 #pass parameters to qmake
 TXSCOMPILEOPTIONS=$*
 if [ "$OPTION_PDFVIEWER" = no ]; then TXSCOMPILEOPTIONS="$TXSCOMPILEOPTIONS NO_POPPLER_PREVIEW=true"; fi
+if [ "$OPTION_TERMINAL" = yes ]; then TXSCOMPILEOPTIONS="$TXSCOMPILEOPTIONS TERMINAL=true"; fi
 if [ "$OPTION_PHONON" = yes ]; then TXSCOMPILEOPTIONS="$TXSCOMPILEOPTIONS PHONON=true"; fi
 if [ "$OPTION_TESTS" = no ]; then TXSCOMPILEOPTIONS="$TXSCOMPILEOPTIONS NO_TESTS=true"; fi
 if [ "$OPTION_DEBUG_BUILD" = yes ]; then
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -2124,7 +2124,7 @@ QAction *ConfigManager::newManagedAction(QObject *rootMenu,QWidget *menu, const 
     if (iconFile.isEmpty()) act = new QAction(text, rootMenu);
     else act = new QAction(getRealIcon(iconFile), text, rootMenu);
 
-    act->setObjectName(completeId);
+	act->setObjectName(completeId);
     act->setShortcuts(shortCuts);
 #if (QT_VERSION <= 0x050700) && (defined(Q_OS_MAC))
     // workaround for osx not being able to use alt+key/esc as shortcut
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -336,6 +336,7 @@ bool ManagedProperty::readFromObject(const QObject *w)
 }
 #undef READ_FROM_OBJECT
 
+const int ConfigManager::MAX_NUM_MACROS;
 QTextCodec *ConfigManager::newFileEncoding = nullptr;
 QString ConfigManager::configDirOverride;
 bool ConfigManager::dontRestoreSession=false;
@@ -1008,7 +1009,7 @@ QSettings *ConfigManager::readSettings(bool reread)
         QDir dir(joinPath(configBaseDir,"macro"));
         if(dir.exists()){
             // use file based macros
-            for (int i = 0; i < 1000; i++) {
+            for (int i = 0; i < ConfigManager::MAX_NUM_MACROS; i++) {
                 QString fileName=joinPath(configBaseDir,"macro",QString("Macro_%1.txsMacro").arg(i));
                 if(QFile(fileName).exists()){
                     Macro macro;
@@ -1020,7 +1021,7 @@ QSettings *ConfigManager::readSettings(bool reread)
             }
         }else{
             if (config->value("Macros/0").isValid()) {
-                for (int i = 0; i < 1000; i++) {
+                for (int i = 0; i < ConfigManager::MAX_NUM_MACROS; i++) {
                     QStringList ls = config->value(QString("Macros/%1").arg(i)).toStringList();
                     if (ls.isEmpty()) break;
                     completerConfig->userMacros.append(Macro(ls));
--- a/src/configmanager.h
+++ b/src/configmanager.h
@@ -34,6 +34,8 @@ class ConfigManager: public QObject, public ConfigManagerInterface
 	Q_OBJECT
 
 public:
+    static const int MAX_NUM_MACROS = 5000;
+
     ConfigManager(QObject *parent = nullptr);
 	~ConfigManager();
 
--- a/src/editors.cpp
+++ b/src/editors.cpp
@@ -11,11 +11,11 @@
  *
  * It maintains the knowledge of the currentEditor() and the currentTabWidget().
  * Also, it sends appropriate signals to the editors and tabWidgets and provides
- * singals to hook to these changes.
+ * signals to hook to these changes.
  *
  * Currently this class serves two purposes:
  *
- * 1) It maintains and abstract order of editors. Editors can be grouped (currently
+ * 1) It maintains an abstract order of editors. Editors can be grouped (currently
  *    implemented as tabs in tabWidgets). The groups are ordered and the editors
  *    within a group are ordered, thus providing a complete order of all editors.
  *    The main purpose of this class it to provide an interface to this order, which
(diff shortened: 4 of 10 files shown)


MeanSquaredError:
A few notes (some of them not related to the actual terminal implementation):

1. It seems that you merged the master branch of TXS into your branch before creating the pull request. It is probably better to do the merge with:
`git pull --rebase upstream master`, where `upstream` is the remote master at `https://github.com/texstudio-org/texstudio.git`
This way your pull request will not contain unnecessary merge commits.

2. When you have multiple commits that implement some feature you can use
`git rebase -i HEAD~3`, where 3 is the number of commits that you want to process interactively. Then you can squash the extra commits into the first one. This way you get only one nice looking commit which is easier to review. I do separate the whitespace formatting changes into a separate commit, though, since nobody wants to review whitespace changes anyway.
If you decide to use `rebase -i` just make sure you read carefully the git documentation about it since you don't really want to mess up your tree (even if you mess it up it can still be fixed through the reflog but it is tedious work).

3. I am sure that at least some people would use terminal under Windows too, especially people who are used to Linux shell but also use Windows (like I do). However if adding Windows support is too much work, I don't mind having it just for Linux.

4. I think that having a single `#ifdef` around the `TerminalWidget` class declaration and then another `#ifdef` around the method implementations is a better approach than having multiple separate `#ifdef`s inside the bodies of the `TerminalWidget` methods. If the second approach makes the build system (gcc ,ld, and/or the MOC compiler) unhappy, then it is probably better to fix the makefile. However I wouldn't expect the makefile to be able to handle gracefully changing the build options like adding/removing the TERMINAL define. So if you need to re-run BUILD.sh, then you can run `make distclean` before that and then run BUILD.sh. That is how I handle changes in the build optons and I don't think it is a big deal.

Of course the above are just my opinions/tastes and the opinion/tastes of the TXS maintainer(s) may differ.

axkibe:
Ah yes, thanks for the git rebasing explanation, I will keep that in mind.

Windows is not supported by qtermwidget5. Windows on this level works completely different, so it ain't easy. I also suppose you wouldn't want a "dos box" but the powershell? I guess this may be doable with ActiveX? I don't know, but a completely different way. Also with the Cygwin shell you'd get a unix like terminal in windows. But would windows users want a unix like shell?

About the #ifdef's yes, if needing to "make distclean" when changing options is okay, then I can simplify this again. Personally I like to keep compiler switches to a minimum, it makes projects so hard to read and maintain if there are too much of them.

timhoffm:
Thanks for looking into this. A terminal would really add value.

Windows: IMHO any sort of terminal would be better than nothing (`command` / powershell / cygwin (if available) / linux subsystem). One of the main usages would probably be calling additional programs like git. Here the kind of terminal does not matter too much. Also, users of the terminal are likely experienced and I doubt that there are many users of `command` / powershell that would have problems doing some basic stuff in a linux shell. Since windows support needs additional work, it's fine to start with linux-only.

sunderme:
can you please test that the terminal handles ‚dark mode‘ gracefully ?

axkibe:
Ok, I will, I guess an additional tab in config dialog will be necessary.

dbitouze:
> Windows: IMHO any sort of terminal would be better than nothing (`command` / powershell / cygwin (if available) / linux subsystem).

Git bash?

axkibe:
@dbitouze I read qtermwidget is accepting PRs for windows support :)
https://github.com/lxqt/qtermwidget/issues/42
for me this is beyond my scope of this endeavor.

dbitouze:
> I read qtermwidget is accepting PRs for windows support :)
> [lxqt/qtermwidget#42](https://github.com/lxqt/qtermwidget/issues/42)
> for me this is beyond my scope of this endeavor.

For me as well :smile: 

> If qtermwidget supports windows, the changes to TexStudio should be minimal to almost none (just the BUILD would need to know it can support it)

Agree.

MeanSquaredError:
@axkibe 

Under Windows most people probably need just the plain cmd.exe (as someone mentioned mainly for issuing different git commands and other similar uses).

Embedding cmd.exe, which is a standalone gui application shouldn't really require ActiveX (and I think cmd.exe does not support ActiveX anyway). I would imagine that plain WinAPI should do the trick, probably something like run cmd.exe, make its main window a child of the corresponding TXS tab window, then modify the look of the cmd.exe window to hide the borders and the system area. That solution would skip entirely using qtermwidget5 when building under windows. Actually googling for it seems that someone has already done it for AutoIt
https://www.autoitscript.com/forum/topic/63322-embed-cmdexe-without-a-border/

Another approach would be to run cmd.exe completely hidden, intercept it when it outputs characters to its window then feed these characters to qtermwidget5 and then feed the user input from qtermwidget5 to cmd.exe. That solution would use qtermwidget5 under Windows too, but is probably too complex and not worth the effort.

BeerStud:
On Mon, 23 Mar 2020 05:20:15 -0700, MeanSquaredError stated:
>@axkibe 
>
>Under Windows most people probably need just the plain cmd.exe (as
>someone mentioned mainly for issuing different git commands and other
>similar uses).
>
>Embedding cmd.exe, which is a standalone gui application shouldn't
>really require ActiveX (and I think cmd.exe does not support ActiveX
>anyway). I would imagine that plain WinAPI should do the trick,
>probably something like run cmd.exe, make its main window a child of
>the corresponding TXS tab window, then modify the look of the cmd.exe
>window to hide the borders and the system area. That solution would
>skip entirely using qtermwidget5 when building under windows. Actually
>googling for it seems that someone has already done it for AutoIt
>https://eur05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.autoitscript.com%2Fforum%2Ftopic%2F63322-embed-cmdexe-without-a-border%2F&amp;data=02%7C01%7C%7Cb4ac3977111249f0e44f08d7cf248c1f%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637205628179086655&amp;sdata=UUOcaq%2Fzx2%2Fz2pPg0zGk9UAb0A8wNQLxybIeWgPqsLc%3D&amp;reserved=0
>
>Another approach would be to run cmd.exe completely hidden, intercept
>it when it outputs characters to its window then feed these characters
>to qtermwidget5 and then feed the user input from qtermwidget5 to
>cmd.exe. That solution would use qtermwidget5 under Windows too, but
>is probably too complex and not worth the effort.

Just out of morbid curiosity, has anyone tried contacting Microsoft
either directly or through one of its numerous forums. For instance,
<https://techcommunity.microsoft.com/>?

Just my 2₵.

-- 
Gerard

axkibe:
I would redirect all the windows comparability discussion to the qtermwidget project. They made it clear that they are welcoming for it, if someone takes it on. qtermwidget provide an API that should abstract this away, at least for linux/bsd/osx they do. Generally speaking terminals sound like an easy thing, but in fact suddendly get very complicated :) I'm very happy for this widget that take care of most of the hard stuff.

sunderme:
qtermwidget seems not to be part of ubuntu

sunderme:
seems to work.
Any reasons not to pull this ?

axkibe:
I'm not fully happy, yet. I wanted to get your opinions. (as they say for open source development, publish early, publish often).
If creating a new file, the hook is missing yet to change the directory.
I suppose there really should be an option tab (color scheme, configure font size and style).

axkibe:
And what I personally don't like Ctrl+L doesn't work, because the Action is bound elsewhere, dunno. Maybe some Actions should be suspended when the terminal has focus?

sunderme:
well, keep it simple. If you want a terminal, take a terminal. Within txs it is a quick way to start a program, nothing more, nothing less.

axkibe:
Haha, thanks, from my opinion you can pull it, it works after all, but I'll be preparing a follow PR to improve anyway (and if I wouldn't I suppose in a few weeks, you'd get enhancement issues opening anyway, asking for these)

axkibe:
I will over an improved version shortly

axkibe:
Thanks for you feedback!

MeanSquaredError:
There was no need to close the PR. You could have just updated your branch and then pushed (or force-pushed) to your upstream of the PR and GitHub would have picked it up.

axkibe:
I thought you want it rebased? :D Ok, then I repoen it. Sorry, I'm used to the job of the maintainer, not of sending PRs.

MeanSquaredError:
You can rebase it and then force-push it. GitHub is smart enough to pick up the changes, update the PR, rerun the tests etc. The benefit from reusing the same PR is that you keep the whole discussion about this feature in one place.

These are not strict rules - I am just giving ideas.

axkibe:
Thanks, they are welcome. As said I'm a) not used to be on this side. b) even on the other side, I ever only had little PRs.

sunderme:
I have pulled the PR, should be fine

charlesxu90:
How to use this functionality in user end?

sunderme:
grafik

charlesxu90:
@sunderme Oh I see. I'm using Mac OS. That's the problem. Thanks for the reply.

cocoaaa:
Hi, Is there any plan for adding this feature for MacOS?

sunderme:
txs would need the library qtermwidget for qt6 on osx. 
If someone can provide that, txs can build with terminal on osx.

axkibe:
I think qtermwidget does support OSX in principle, I just didn't care for it and not owning a recent Apple device had no way to test it too.

sunderme:
It should be possible. But someone has to set up the build chain on github.

Meron35Z:
Now that qtermwidget has been updated to qt6 in most repositories, is it possible to update this so that it builds against qtermwidget6 instead?

PR #998, opened 2020-03-24
Config tab for terminal

axkibe:
And here is the announced config tab for the terminal, allowing setting of color scheme, font family, font size and the shell to be used.

(I also combinded the #ifdef's in toolwidgets.cpp again)


sunderme:
Thanks

PR #1003, opened 2020-03-26
override ctrl+c shortcut snitching

axkibe:
Issue: Ctrl+C is something I considered should really work in the terminal, otherwise somebody may get really mad about some accidental started destructive operation they cant stop.

While at it added some other key combos, that don't make sense when focusing the terminal.

Also had forgotten a proper destructor, oops

### DIFF
--- a/src/toolwidgets.cpp
+++ b/src/toolwidgets.cpp
@@ -147,7 +147,7 @@ TerminalWidget::TerminalWidget(QWidget *parent): QWidget(parent)
 	layout->setSpacing(0);
 	layout->setMargin(0);
 	setLayout(layout);
-	this->installEventFilter(this);
+	installEventFilter(this);
 	initQTermWidget();
 }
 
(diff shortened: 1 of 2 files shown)


src/toolwidgets.cpp
```diff
@@ -147,10 +147,38 @@ TerminalWidget::TerminalWidget(QWidget *parent): QWidget(parent)
 	layout->setSpacing(0);
 	layout->setMargin(0);
 	setLayout(layout);
-
+	this->installEventFilter(this);
```
MeanSquaredError: `this->` is superfluous.

MeanSquaredError:
It might be better to intercept all keyboard events instead of separate combinations. I don't have a clear opinion on that, though.

src/toolwidgets.cpp
axkibe: your correct. I cleaned this up.

axkibe:
This is a design decision. Probably there are Shortcuts you want to keep, like F1 for quick build.

I suppose ideally it could be config options, all shortcuts to terminal when focus, or only selected ones (where not having it really doesn't make much sense).

I considered Ctrl+C to be quite important tough.

Issue #1008, opened 2020-03-26
TERMINAL does not build on travis


sunderme:
Terminal support does not build on travis-ci, see https://travis-ci.org/github/texstudio-org/texstudio/jobs/667383301?utm_medium=notification&utm_source=github_status

### Environment
- TeXstudio: HEAD
- Qt: 
- OS: ubuntu 16.04
- TeX distribution: N/A

axkibe:
Probably a tool old version of the qtermwidget library not supporting setTerminalSizeHint.
Which is absolutely not essential, the size hints are IMO just undesirable in the case of TexStudio
Unfortunately I don't a way to #ifdef for the library versions or even the existence of this member function.

sunderme:
there are ways to detect version with pkg-config (see has_poppler32).
However that seems not to work properly with qtermwidget ...

MeanSquaredError:
`qtermwidget` does not provide the corresponding `.pc` file, so `pkg-config` does not work with it indeed.

I think that the easiest workaround would be to drop the autodetection of `qtermwidget` and just expect the user to specify the `TERMINAL` build option if he needs the terminal in TXS.

Another more complex workaround would be to use `qtCompileTest()` as described here
https://stackoverflow.com/a/44011108

sunderme:
hmmm, on ubuntu: '/usr/lib/x86_64-linux-gnu/pkgconfig/qtermwidget5.pc'

MeanSquaredError:
on Fedora 31 it does not have the `.pc` file:
```
root@localhost:~# rpm -ql qtermwidget | grep -Ei '\.pc'
root@localhost:~# 
```
Do you mean that `pkg-config --print-provides qtermwidget5` does not work on the Travis-CI server but works on your development machine?

In the latter case maybe the Travis-CI server has a qtermwidget package without the `.pc` file?

sunderme:
Pkg-config -print-provides works fine (here and on travis)
--atleast-version seems not to work and qmakes existPackage (or similar) does not detect libqtermwidget...

MeanSquaredError:
The Qt documentation says that `packagesExist` uses `pkg-config`, so I cannot test it on my machine because on Fedora the `.pc` file is missing.

Can you check the `Name:` tag inside `/usr/lib/x86_64-linux-gnu/pkgconfig/qtermwidget5.pc` ?
Is the package `Name:` defined as `libqtermwidget` or `qtermwidget5` ?

[...]

PR #1014, opened 2020-03-28
Fix detection of qtermwidget5 in qmake project file

MeanSquaredError:
This PR fixes the detection of qtermwidget5 on Fedora 31 (probably on other Linux distributions too).
It is expected to fix https://github.com/texstudio-org/texstudio/issues/1008

### DIFF
--- a/texstudio.pro
+++ b/texstudio.pro
@@ -1,3 +1,43 @@
+#########################################################################################
+# pkgAtLeastVersion(widget_name, widget_version)
+# Custom test that checks if widget_name is installed and at least version widget_version
+#########################################################################################
+defineTest(pkgAtLeastVersion) {
+	# We do not use "pkg-config --atleast-version ..." because there is bug in
+	# Ubuntu 16.04 which causes the command to fail.
+	MOD_LINES = $$system($$pkgConfigExecutable() "--modversion " $$1, lines, exitCode)
+	if (!isEqual(exitCode, 0)) {
+		return (false)
+	}
+	if (isEmpty(MOD_LINES)) {
+		return (false)
+	}
+	MOD_VERSION = $$first(MOD_LINES)
+	# Compare found module version in MOD_VERSION and required version in argument 2
+	# versionAtLeast() is not supported before Qt 5.10, so we have a custom implementation
+	# of this function.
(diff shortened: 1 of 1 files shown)


sunderme:
the version detection does not work here.
pkg-config always exists with status 1.

MeanSquaredError:
I updated the code so that texstudio.pro will log a bit of extra information just before it tries to detect qtermwidget5. If qtermwidget5 .pc file is present it should print the variable names (without the values) from the .pc file. Otherwise it should print some error message.

Can you please run qmake and let me know what it actually prints?

sunderme:
`message(Begin terminal check)
system($$PKG_CONFIG_EXE "--errors-to-stdout --print-variables qtermwidget5")
message(End terminal check)
isEmpty(TERMINAL):system($$PKG_CONFIG_EXE "--atleast-version 0.9.0 qtermwidget5"){`

MeanSquaredError:
@sunderme 
The variables names look OK (same as on Fedora 31)

Can you please upload your file `qtermwidget5.pc`
IIRC you mentioned that it is in
`/usr/lib/x86_64-linux-gnu/pkgconfig/qtermwidget5.pc`

sunderme:
[pc.zip](https://github.com/texstudio-org/texstudio/files/4397363/pc.zip)

MeanSquaredError:
That is funny - it says `Version: 0.14.1` which is definitely sufficient.
What does `pkg-config --modversion qtermwidget5` print when you run it on your machine?

sunderme:
0.14.1

sunderme:
--atleast-version seems not to give the correct answer, whyever ...

MeanSquaredError:
That's really strange. I guess it is some kind of pk-config bug. Anyway, I will update the PR so that it will use `pkg-config --modversion` as a workaround.

MeanSquaredError:
@sunderme 
I implemented a workaround that uses `pkg-config --modversion` to check if there is a suitable version of `qtermwidget5`. It works correctly on Fedora 31. Please test it on your machine and let me know if it works correctly on your machine too.

sunderme:
that works
 I have pulled it, thanks.

PR #1017, opened 2020-03-29
Fix a bug that prevents mouse cursor from changing its look in the embedded PDF viewer

MeanSquaredError:
There is a bug in TXS which causes it to disable mouse tracking events for the PDF viewer widget which in turns prevents the widget from updating its internal state. For example this bug prevents the cursor from changing to pointing hand when hovering over links in the internal PDF viewer.

How to reproduce:

1. Start TXS.
2. Open "Options / Configure TeXstudio..." then go to "Internal PDF viewer" and disable "Auto-hide Toolbars in Embedded Mode", then save the settings by clicking on the [OK] button.
3. Create a sample .tex file that has a link, e.g.
```
\documentclass{amsart}
\usepackage[unicode=true,bookmarks=false,breaklinks=false,pdfborder={0 0 1},backref=section,colorlinks=true]{hyperref}
\hypersetup{linkcolor=blue,citecolor=blue,filecolor=blue,urlcolor=blue,pdftitle=,pdfauthor=,pdfsubject=,pdfkeywords=}
\begin{document}
\url{https://www.google.com/}
\end{document}
```
4. Compile the document and view it in the embedded PDF viewer. Hover the mouse over the link to www.google.com to make sure that the cursor turns to a pointing hand when hovering over the link.
5. Open "Options / Configure TeXstudio..." and then close the configuration popup by clicking on the [Cancel] button.
6. Hover the mouse cursor in the embedded PDF viewer over the link to www.google.com and see that it no longer turns into a pointing hand when the cursor is over the link.

The problem happends because the following call chain happens that disables the mouse tracking for the PDFWidget:

```
#0  QWidget::setMouseTracking (this=0x1967790, enable=false) at ../../../include/qt5/QtWidgets/qwidget.h:814
#1  0x0000000000b6a3f2 in PDFDocument::setAutoHideToolbars (this=0x2cf15c0, enabled=false) at src/pdfviewer/PDFDocument.cpp:4208
#2  0x0000000000b5d4ef in PDFDocument::reloadSettings (this=0x2cf15c0) at src/pdfviewer/PDFDocument.cpp:2575
#3  0x0000000000962ffa in Texstudio::generalOptions (this=0x1770eb0) at src/texstudio.cpp:6561
```

The proposed PR fixes this bug.

EDIT:

I just added a fix for a second bug that was present the the code for the PDF viewer auto-hide button. This second bug can be reproduced by:

1. Start TXS and display some PDF in the embedded (internal) PDF viewer.
2. Disable PDF viewer auto-hiding of the toolbar. This can be done either from "Options / Configure TeXstudio... / Internal PDF viewer" or from the auto-hide button on the embedded PDF viewer toolbar.
3. Exit TXS.
4. Start TXS. The previous PDF will be displayed in the embedded PDF viewer and auto-hiding of the toolbar will be disabled.
5. Enable auto-hiding of the PDF viewer toolbars by clicking on the auto-hide button on the toolbar.

Even though auto-hiding is enabled it will not work correctly, that is the toolbar will be visible as long as the mouse pointer is hovering over the PDF in the embedded PDF viewer.

This second bug is fixed by calling `reloadSettings()` in `PDFDocument::toggleAutoHideToolbars`.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -3231,6 +3231,7 @@ void PDFDocument::toggleAutoHideToolbars()
 	QAction *act = qobject_cast<QAction *>(sender());
 	if (act) {
 		globalConfig->autoHideToolbars = act->isChecked();
+		reloadSettings();
 	}
 }
 
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -539,6 +539,8 @@ PDFWidget::PDFWidget(bool embedded)
 	setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Ignored);
 	setFocusPolicy(embedded ? Qt::NoFocus : Qt::StrongFocus);
 	setScaledContents(true);
+	// Mouse tracking must be always enabled because we use the mouseMoveEvent to update the
+	// widget state (e.g. change the cursor image when hovering over links)
 	setMouseTracking(true);
 	grabGesture(Qt::PinchGesture);
 	grabGesture(Qt::TapGesture);
@@ -4200,13 +4202,14 @@ void PDFDocument::setAutoHideToolbars(bool enabled)
 {
 	setToolbarsVisible(!enabled);
 	actionAutoHideToolbars->setChecked(enabled);
-	// since we want to have the MouseMoveEvent down at the pdfWidget (internally e.g. for magnifier) up to
-	// the window (for toolbar hiding) all widgets in between seem to need MouseTracking enabled. Otherwise
-	// they will swallow the move event.
-	QWidget *w = pdfWidget;
-	while (w) {
-		w->setMouseTracking(enabled);
-		w = w->parentWidget();
+	// When autohiding the toolbars we need the mouseMoveEvent of pdfWidget in order to track user activity.
+	// We also enable/disable the mouse-tracking flag of all ancestors of the widget because otherwise the may
+	// swallow the mouseMoveEvent. However we do not change the mouse-tracking flag of the pdfWidget itself
+	// because it must be always enabled (the widget uses mouseMoveEvent to maintain its internal state).
+	if (pdfWidget) {
+		for (QWidget *w = pdfWidget->parentWidget(); w; w = w->parentWidget()) {
+			w->setMouseTracking(enabled);
+		}
 	}
 }
 


src/pdfviewer/PDFDocument.cpp
```diff
@@ -4200,13 +4202,14 @@ void PDFDocument::setAutoHideToolbars(bool enabled)
...
-	QWidget *w = pdfWidget;
-	while (w) {
-		w->setMouseTracking(enabled);
-		w = w->parentWidget();
+	// When autohiding the toolbars we need the mouseMoveEvent of pdfWidget in order to track user activity.
+	// We also enable/disable the mouse-tracking flag of all ancestors of the widget because otherwise the may
+	// swallow the mouseMoveEvent. However we do not change the mouse-tracking flag of the pdfWidget itself
+	// because it must be always enabled (the widget uses mouseMoveEvent to maintain its internal state).
+	if (pdfWidget) {
+		for (QWidget *w = pdfWidget->parentWidget(); w; w = w->parentWidget()) {
```
timhoffm: I'm not sure anymore that this is needed at all or makes sense with the modification. The original comment suggested that we had to enable tracking for all parents to make the pdfWidget tracking work. If we're not changing the tracking status of pdfWidget anymore, we should also not need to do it for the parents.

Can you please check if enabling tracking for parents is necessary at all? If not, this loop can be removed. If so, we should do it once in the PDFWidget constructor.

MeanSquaredError: I just tried removing this loop from `PDFDocument::setAutoHideToolbars` and it caused problems when auto-hiding of toolbars is enabled. Without the loop it always shows the toolbars when the mouse is over the pdf viewer. That is it does not auto-hide the toolbars. 

I tried moving the loop into the PDFWidget constructor and changing the parameter to `true`, that is 
```
	// Mouse tracking must be always enabled because we use the mouseMoveEvent to update the
	// widget state (e.g. change the cursor image when hovering over links)
	setMouseTracking(true);
	// When autohiding the toolbars we need the mouseMoveEvent of pdfWidget in order to track user activity.
	// We also enable/disable the mouse-tracking flag of all ancestors of the widget because otherwise the may
	// swallow the mouseMoveEvent. However we do not change the mouse-tracking flag of the pdfWidget itself
	// because it must be always enabled (the widget uses mouseMoveEvent to maintain its internal state).
	for (QWidget *w = parentWidget(); w; w = w->parentWidget()) {
		w->setMouseTracking(true);
	}
```
Then the same problem happened - it just does not auto-hide the toolbars. Thinking about it, this is expected because when we execute the PDFWidget constructor, it does not have any parent yet, so none of its future ancestors get its mouse tracking enabled.

One place where we could move the loop to is into `PDFDocument::init` just after `pdfWidget->setPDFDocument(this);`
because at this point the widget already has its ancestor chain setup and we can enable tracking for them.

However I am not really sure that we want `PDFDocument::mouseMoveEvent` to be called when auto-hide is disabled. After all that method checks if auto-hide is enabled and if it is disabled, the method does nothing.

So I my guess that the current code is OK. What do you think about it?

sunderme:
thanks

Issue #1019, opened 2020-03-31
texstudio crash after using option filter


fg-dev:
When I  try to filter option in the configure TeXstudio window, the app crashes.

### Environment
- TeXstudio: 3.00.00 (git 3.0.0alpha2-34-gad302ef)
- Qt: 5.12.1, compiled with Qt 5.12.1 R
- OS: Windows 10 64bit
- TeX distribution: miktex

### Expected behavior

Filter the option according to the typed letters.
### Actual behavior

crash and the app window close.

### How to reproduce
open configure TeXstudio and type a letter in option filter.

sunderme:
works fine on linux and windows here...
Have you tried reinstall ? removing texstudio.ini ?

fg-dev:
I deleted the texstudio.ini file however the problem still persists
for the installation, I am using the portable file.

MeanSquaredError:
I tested it on Windows 7 and it did crash indeed. I will try to prepare a fix shortly.

fg-dev:
> I tested it on Windows 7 and it did crash indeed. I will try to prepare a fix shortly.

thank you.

PR #1022, opened 2020-04-02
Fix a crash in ConfigDialog::metaFilterChanged()

MeanSquaredError:
This is a fix for https://github.com/texstudio-org/texstudio/issues/1019

TXS crashes if terminal support is not built and the user tries to use the option filter in the config dialog.

The crash happens in `ConfigDialog::metaFilterChanged` on the line `ui.contentsWidget->item(i)->setHidden(!on);` 
because `ui.pagesWidget->count()` is 16 (terminal page settings exist in `configdialog.ui`)
on the other hand `ui.contentsWidget->count()` is 15 (terminal page content item is not added to the list through a call to `ConfigDialog::createIcon()`).

The proposed PR fixes the issue by always adding the terminal content item to the contents list, but when terminal is disabled during build then the contents item is kept always hidden in the config dialog. This is a hack and ideally we should not add the terminal configuration page when terminal is disabled, however it seems that Qt does not support conditional compiling of widgets from the .ui files.

Maybe the proper solution would be to implement some kind of XML parser/builder that would remove the terminal settings from `configdialog.ui`, but it would require lots of custom code for the XML parser/builder in the project file, or dependency on an external XML tool.

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -509,20 +509,28 @@ ConfigDialog::ConfigDialog(QWidget *parent): QDialog(parent), checkboxInternalPD
 	createIcon(tr("Commands"), getRealIcon("config_commands"));
 	createIcon(tr("Build"), getRealIcon("config_quickbuild"));
 	createIcon(tr("Shortcuts"), getRealIcon("config_shortcut"));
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

PR #1023, opened 2020-04-02
Makes badwords detection case insensitive

mrclschstr:
I discovered that the detection of bad words and stop words is case sensitive at the moment. Therefore you need to add each word at least twice to make it work at the beginning of a sentence (e.g. `again` and `Again`).

**Important note:** I don't have a full developement setup on my computer right now. For that reason I couldn't compile and test my changes. I would really appreciate if you could check that for me 😃

### DIFF
--- a/src/grammarcheck.cpp
+++ b/src/grammarcheck.cpp
@@ -428,9 +428,10 @@ void GrammarCheck::backendChecked(uint crticket, int subticket, const QList<Gram
 	}
 	if (config.badWordCheck) {
 		for (int w = 0 ; w < words.size(); w++) {
-			if (ld.badWords.contains(words[w]))
+			QString normalized = words[w].toLower();
+			if (ld.badWords.contains(normalized))
 				cr.errors[tb.lines[w]] << GrammarError(tb.indices[w], tb.endindices[w] - tb.indices[w], GET_BAD_WORD, tr("Bad word"), QStringList() << "");
-			else if (words[w].length() > 1 && words[w].endsWith('.') && ld.badWords.contains(words[w].left(words[w].length() - 1)))
+			else if (normalized.length() > 1 && normalized.endsWith('.') && ld.badWords.contains(normalized.left(normalized.length() - 1)))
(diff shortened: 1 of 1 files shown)


sunderme:
I am not convinced that it should be case insensitive. Other language (e.g. German) distinguih the case within a sentence.

Furthermore you exchanged Qset<QString> with QStringlist. The reason for Qset is speed and even doing the comparison case insensitive, you don't need to switch to stringlist but just convert the test word to lower case...

MeanSquaredError:
Switching from `QSet` to `QStringList` might slow the grammar check for large documents and/or large lists of badwords/stopwords. It is probably a good idea to keep the lists as QSets.

MeanSquaredError:
@sunderme 
Sorry, I did not see your comment, it seems we commented simultaneously.

mrclschstr:
> I am not convinced that it should be case insensitive. Other languages (e.g. German) distinguish the case within a sentence.

I'm sorry, I don't get your point (btw I'm german). In my opinion the reason for bad words and stop words is the "word itself" regardless of the capitalization.

I get the point reagarding the performance of `QStringList`, but now I have to add each word at least twice to cover each capitalization possibillity. I don't think doubling the word lists is beneficial for performance as well, but this ist just a guess.

Do you have another suggestion? Maybe another option to activate case insensitivity?

sunderme:
concerning QStringList <-> Qset: it needs to use QSet.

sunderme:
we can pull the case insensitivity, but it it should use sets...

mrclschstr:
> we can pull the case insensitivity, but it it should use sets...

Alright, I'll fix this 😃

mrclschstr:
I had another clean look at the code and discovered, that the stop words are already normalized to lower case (see following lines as well): https://github.com/texstudio-org/texstudio/blob/c27a8c1438b0436bf06e6e292402026a51ccc456/src/grammarcheck.cpp#L403

Therefore I chose a similar approach in my new PR attempt 😄

sunderme:
thanks

PR #1027, opened 2020-04-04
Added 3 more themes to the Beamer wizard

JackHack96:
I've added 3 modern themes to the Quick Beamer wizard.


sunderme:
thanks

homocomputeris:
Is it possible to fetch the list of installed themes from the disk?

muzimuzhi:
@homocomputeris Which list, the list of installed latex beamer theme, or the list of themes in beamer wizard of texstudio? If it's the first list, then here, the repo of a latex editor, is not the right place to ask. 

PS: You could (or should) consider posting a new issue.

homocomputeris:
I mean, can't TXS get the list of beamer themes from the disk instead of relying on the hardcoded one? (It's not a reason for an issue, just curiosity)

PR #1030, opened 2020-04-06
Fix the description of the ?p{ext}: pathname selector in the manual.

MeanSquaredError:
While answering a user question I noticed that the description of the `?p{ext}:` pathname selector in the manual is incorrect in several places. In some places the trailing `:` is missing. In one place the `p` character is missing, etc.

The proposed PR fixes the description in the manual.


sunderme:
thanks

Issue #1044, opened 2020-04-18
TXS should not install files into /usr/share/appdata


jonasstein:
### Environment
- TeXstudio: 2.12.*
- OS: Linux

> 2.1.2. Filesystem locations
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

`/usr/share/appdata/texstudio.appdata.xml`
should be
`/usr/share/metainfo/texstudio.appdata.xml`

(Ticket on Gentoo Bug Tracker https://bugs.gentoo.org/709464)

PR #1045, opened 2020-04-19
store appdata to /usr/share/metainfo

jonasstein:
/usr/share/appdata/texstudio.appdata.xml
is deprecated and should be changed to
/usr/share/metainfo/texstudio.appdata.xml

See section

    2.1.2. Filesystem locations
    https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

(Ticket on Gentoo Bug Tracker https://bugs.gentoo.org/709464)
Closes: #1044


sunderme:
thanks

PR #1051, opened 2020-04-23
Simplify the debug logger code

MeanSquaredError:
After the removal of the support for Qt4, the two classes implementing the debug logger, `DebugLoggerMain` and `DebugLoggerQt` can be merged into a single class. This PR simplifies the debug logger code by merging the two classes.


sunderme:
thanks

PR #1053, opened 2020-04-24
cwl - add no-opt-arg lines to xcolor.cwl

muzimuzhi:
In #793, I used to think that no-optional-arg form of commands `\definecolor`, `\providecolor`, and `\colorlet` are rarely used. I am wrong so it comes the current pr.


sunderme:
thanks

PR #1055, opened 2020-04-25
Fix events processing in tests

MeanSquaredError:
This PR fixes several minor issues with the handling of events while running tests.

1. Currently we try to ignore UI events while running tests by processing all events except for UI and network events
( `QApplication::processEvents(QEventLoop::ExcludeUserInputEvents | QEventLoop::ExcludeSocketNotifiers)` )
However there are several places where either QTest or Qt core framework internally processes all events including UI events.
Most notably this happens in `QTest::qWait()` and in the different Qt functions that show the message popups during search&replace testing.

As a result it is possible to crash TXS while it is running the tests by providing keyboard or mouse input. For example by running on Linux:

`textstudio --execute-all-tests`
then while tests are running, pressing CTRL+W several times at different moments while tests are running.

This will close the main tab in which the tests are running, so naturally TXS will crash with SIGSEGV when the testing framework tries to access the main document tab.

Same type of crash will happen is the mouse is clicked on the [x] button for the main tab.

I tried several approached at fixing this issue and the best one seems to be installing a native events filter inside TestManager. This filter only filters events coming from the physical keyboard and mouse and the synthetic events from the QTest framework are not affected. So this kind of fix is used in this

2. Another possible issue taken care by this PR is that while processing events by `QApplication::processEvents(...)` our current code does not take care of the posted events non-system events as recommended by the documentation at

https://doc.qt.io/qt-5/qcoreapplication.html#processEvents

This is mostly a problem with deferred deletes (`QEvent::DeferredDelete`) as documented by the above link. There is at least one place where we use deferred deletes in MessageBoxCloser::closeNow(). I do not know if skipping processing of deferred deletes is a problem besides not freeing memory early on, but I checked the source code for `QTest::qWait()` and copied I copied their code that processes `QEvent::DeferredDelete` into our custom function `QTest::processPendingEvents()`.

3. Additionally this PR adds ifdefs to `SyntaxCheck::waitForQueueProcess()` so it is only defined when tests are enabled. The documentation for this method is slightly expanded and processing for deferred deletes are added to that method too.

### DIFF
--- a/src/syntaxcheck.cpp
+++ b/src/syntaxcheck.cpp
@@ -291,7 +291,8 @@ void SyntaxCheck::waitForQueueProcess(void)
 	linesBefore = mLinesEnqueuedCounter.fetchAndAddOrdered(0);
 	forever {
 		for (int i = 0; i < 2; ++i) {
-			QApplication::processEvents(QEventLoop::ExcludeUserInputEvents | QEventLoop::ExcludeSocketNotifiers, 1000); // Process queued checkNextLine events
+			QCoreApplication::processEvents(QEventLoop::AllEvents, 1000); 			// Process queued checkNextLine events
+			QCoreApplication::sendPostedEvents(Q_NULLPTR, QEvent::DeferredDelete);		// Deferred delete must be processed explicitly. Using 0 for event_type does not work.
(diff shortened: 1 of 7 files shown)


sunderme:
thanks

PR #1056, opened 2020-04-25
Fix a build warning by removing an unused variable

MeanSquaredError:
Fix a build warning by removing the unused variable `backendAvailable` in `src/grammarcheck.cpp`


sunderme:
thanks

Issue #1046, opened 2020-04-19
Open terminal does not work


jonasstein:
### Environment
```
TeXstudio 3.00.00 (git 3.0.0alpha3-15-g5147db65)
Using Qt Version 5.14.2, compiled with Qt 5.14.2 R
OS: Gentoo Linux
```

Compiled latest commit from git on Gentoo Linux to support https://github.com/texstudio-org/texstudio/issues/888 and tested, if https://github.com/texstudio-org/texstudio/pull/898 fixes the problem here too. 

**Clicking on `Open Terminal` does not do anything.** 
Not even a message on STDERR. Do we need specific settings?

https://github.com/texstudio-org/texstudio/issues/888#issuecomment-576333120

MeanSquaredError:
Does this problem happen only in TeXstudio 3 or does it also exist in TexStudio 2.12.x ?

KarlMagnusLarsson:
I can not repeat this on Linux Debian 10.3 (or OSX 10.13.6).

Texstudio 3.00.00 (git 3.0.0alpha3-15-g5147db65)
Using Qt Version 5.11.3, compiled with Qt 5.11.3 R.
OS: Debian Linux 10.3
Tools->Open terminal work for me and opens a terminal at working directory.

Building on OSX, 10.13.6 also work for me.
TeXstudio 3.00.00 (git 3.0.0alpha3-15-g5147db65)
Using Qt Version 5.14.1, compiled with Qt 5.14.1 R

jonasstein:
Yes, same for 2.12.22
```
TeXstudio 2.12.22 (git n/a)
Using Qt Version 5.14.2, compiled with Qt 5.14.2 R
```

Any idea, what I should test?
How do you call the terminal exactly?
I fear it is a bug in our package. Just opened a ticket on Gentoo (https://bugs.gentoo.org/719132)

MeanSquaredError:
FYI, it works correctly on Fedora 31, too.

The terminal command is invoked in `startTerminalEmulator()` defined in `src/utilssystem_unix.cpp`
This function in turn relies on `getTerminalCommand()` (again defined in `src/utilssystem_unix.cpp`) to get the executable pathname of the terminal program.

`getTerminalCommand()` tries a few things in order to get the pathname of the terminal executable:

1. First it tries to run `gsettings get org.gnome.desktop.default-applications.terminal exec`
If that succeeds, it strips the surrounding quotes and uses the result as the executable pathname.
[...]

PR #1061, opened 2020-04-29
Make configurable the external terminal command

MeanSquaredError:
This PR makes configurable the command for the external terminal as discussed in https://github.com/texstudio-org/texstudio/issues/1046

The PR provides the following changes:

1. It adds a new command (`txs:///terminal-external`) which can be configured through
`Options / Configure TeXstudio... / Commands / External Terminal`
By default the value for the external terminal is initialized to a value which mimics the executable autodetected by the old code in `startTerminalEmulator()` (different values for macOS, UNIX and Windows).
The command can use the regular command expansion symbols just like the other commands. The macOS version of the command makes use of that by specifying the default value of the external terminal commads as `open -a Terminal ?c:a"`
This allows us to get rid of the macOS specific code which had to specify the work directory explicitly when running the external terminal.

2. The old menu item `Tools / Open Terminal` is renamed to `Tools / Open External Terminal`. The updated menu item runs the external command through `BuildManager::runCommand()` and therefore it outputs information about the execution of the command into the log window, just like the other commands do. More specifically it will log any execution errors that might happen.

3. The builtin terminal is renamed to `Internal Terminal`. More specifically the settings are renamed from `Options / Configure TeXstudio... / Terminal` to `Options / Configure TeXstudio... / Internal Terminal`.
Also the qmake and C++ defines are renamed from `TERMINAL` to `INTERNAL_TERMINAL` in order to avoid confusion about the purpose of this define.


sunderme:
thanks

PR #1062, opened 2020-04-29
Cwl changes

muzimuzhi:
Main changes
 - add keyvals list to `pdfpages.cwl`
 - add opt-arg variants of `\parbox` to `latex-document.cwl`
 - minor fix to keyvals of `hlist.cwl`


sunderme:
thanks

PR #1077, opened 2020-05-06
Fix cwl

### LINKED ISSUE #1072: Autogenerated cwl is missing environments created with \specialcomment from comment package
### Environment
- TeXstudio: 2.12.22+debian-1build1
- Qt: 5.12.8
- OS: Ubuntu 20.04
- TeX distribution: texlive

### Expected Behavior
I use `\specialcomment{solution}...` from the comment package to create an environment that I can exclude for certain builds (e.g. solutions in an assignment sheet) using `\excludecomment`.  As for every other regular environment defined in my cls file (works fine), I expect that I get completion for `\begin{solution}`.

I noticed that there is a folder with auto generated cwl files. the cwl for my cls file is missing the solution environment. Probably because it does not take into account that `\specialcomment` also creates such an environment.

### Current Workaround
At the moment I fix this by adding the missing parts to a custom cwl file. So its not a huge problem.

### Side note
The command  `\excludecomment` from the comment package is also not recognized. But all other commands from that package are, as far as I can see.

muzimuzhi:
Main changes:
 - `comment.cwl`, fix typo in `\specialcoment`, should contain double `m`. This should fix #1072.
 - `latex-dev.cwl`, remove undefined `\lineskiplimits`. `\lineskiplimit` is a primitive glue register, and is recorded in `tex.cwl`.

### DIFF
--- a/completion/comment.cwl
+++ b/completion/comment.cwl
@@ -4,7 +4,7 @@
 \begin{comment}#V
 \end{comment}
 \includecomment{envname}#N
-\excludecoment{envname}#N
+\excludecomment{envname}#N
 \CommentCutFile#*
 \ProcessCutFile#*
 \generalcomment{envname}{begdef}{enddef}#*N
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

Issue #1069, opened 2020-05-02
Magnifying glass cursor not reverting back to normal after magnifying preview


paulenit:
### Environment
- TeXstudio: _2.12.22_
- Qt: _5.14.2, compiled with Qt 5.14.0 R_
- OS: _Manjaro with Xfce Desktop 4.14_
- TeX distribution: _texlive_

### Expected behavior
When clicking "Build & View (F5)" and hovering and clicking over the preview (internal pdf viewer) the cursor changes to the magnifying glass. When pressing ESC to close the preview panel and getting back to the source code panel the cursor should switch back to default (text editing vertical line).

### Actual behavior
The cursor does not revert and you still have the magnifying glass following you around your text. This does not change upon clicking and selecting text.
!!! Does only happen when clicking in the preview i.e. magnifying the output. !!!

### How to reproduce
Open valid latex document (simple 'article' template is sufficient) and press F5, then click in the preview panel, press ESC and return to the editing panel.

sunderme:
seems to be fixed for currentdev builds.
Can you try 3.0.0alpha4 ? (see releases) or build from source ?

paulenit:
Downloaded latest AppImage from bintray (texstudio-linux-3.00.00-202004291523-git_e20ec23-x86_64.AppImage) and the bug is still there.

MeanSquaredError:
I managed to reproduce the bug. I will try to provide a fix today.

MeanSquaredError:
Actually the bug only happens if the user does not move the mouse after releasing the mouse button that caused the zoom. That is:
1. Open the internal PDF viewer.
2. Hover the mouse over some text and press left mouse button (or whichever button causes the magnifying glass to open).
3. Release the mouse button.
4. Without moving the mouse press ESC to close the internal PDF viewer.

Now to find out what exactly is going on in the code...

MeanSquaredError:
After some digging into the TXS and Qt source code it seems that this time it is not a TXS bug, but a bug in the Qt itself.

Looks like in some cases when we close the `PDFDocument` widget, Qt does not detect that the widget below the cursor has changed, so it does not bother to refresh the look of the cursor. The bug only shows up when the mouse has not been moved between the mouse button release and ESC keypress. However it has nothing to do with our `PDFWidget::mouseMoveEvent` handler, because it behaves the same way even when I completely remove our `PDFWidget::mouseMoveEvent` handler.

I will try to prepare a workaround that is not too ugly :-)

[...]

PR #1087, opened 2020-05-13
Restore the cursor shape properly when closing the PDF viewer

MeanSquaredError:
This PR is a fix for https://github.com/texstudio-org/texstudio/issues/1069
Effectively it is a workaround for what seems to be a Qt bug.


MeanSquaredError:
It seems to me that the Travis-CI errors are cause by some kind of internal server error on github.

sunderme:
thanks

Issue #1094, opened 2020-05-15
open external terminal does not work


fg-dev:
### Environment
- TeXstudio: TeXstudio 3.00.00 (git 3.0.0alpha4-32-g577b520)
- Qt: Version 5.12.1, compiled with Qt 5.12.1 R
- OS: Windows 10 64bit 
- TeX distribution: miktex

### Expected behavior

an open window of cmd

### Actual behavior
I can see in the message windows this message "Process started: C:\WINDOWS\system32\cmd.exe"
but no window appears.

I checked in the process monitor, and I found the process is started, but no window appears.

### How to reproduce
go to Tools menu, then click on the open external terminal.

MeanSquaredError:
I can confirm the problem. I think it is due to the new code that changes the way the external terminal executable is run. I will provide a PR with the fix shortly.

PR #1096, opened 2020-05-17
Fix "Open External Terminal" on Windows

MeanSquaredError:
This PR fixes https://github.com/texstudio-org/texstudio/issues/1094

Basically the problem is that on Windows we have to pass `CREATE_NEW_CONSOLE` when creating the new process. Qt only supports that kind of Windows-specific flags since Qt 5.7 (`QProcess::setCreateProcessArgumentsModifier`). I assume that we need compatibility with Qt versions 5.7, so I switched to Windows-specific code similar to the one that we had before.

This PR also adds support for `ExecProgram::execDetached()` which is a wrapper for `QProcess::startDetached()`. It is used when opening a terminal under Linux.

Given that we no longer use `BuildManager::runCommand()`, we no longer can rely on its success/error reporting, so this PR also does some basic success/error reporting in the log window.

### DIFF
--- a/src/execprogram.cpp
+++ b/src/execprogram.cpp
@@ -2,18 +2,24 @@
 #include "utilsSystem.h"
 
 ExecProgram::ExecProgram(void) :
+#ifdef Q_OS_WIN
+	m_winProcModifier(nullptr),
+#endif
 	m_normalRun(false),
 	m_exitCode(-1)
 {
 }
 
-ExecProgram::ExecProgram(const QString &progNameAndArguments, const QString &additionalSearchPaths, const QString &workingDirectory) :
-	m_program(progNameAndArguments),
+ExecProgram::ExecProgram(const QString &shellCommandLine, const QString &additionalSearchPaths, const QString &workingDirectory) :
 	m_additionalSearchPaths(additionalSearchPaths),
 	m_workingDirectory(workingDirectory),
+#ifdef Q_OS_WIN
+	m_winProcModifier(nullptr),
+#endif
 	m_normalRun(false),
 	m_exitCode(-1)
 {
+	setProgramAndArguments(shellCommandLine);
 }
 
 ExecProgram::ExecProgram(const QString &progName, const QStringList &arguments, const QString &additionalSearchPaths, const QString &workingDirectory) :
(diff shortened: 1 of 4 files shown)


MeanSquaredError:
The Travis-CI build error looks like it is caused by a missing STL header include. Let me try to fix that.

MeanSquaredError:
`CreateProcessW` is not really related to the lambdas.

We use `CreateProcessW` because on Qt 5.0 - Qt 5.6, there is no way to specify the internal flag `CREATE_NEW_CONSOLE` when creating the `cmd.exe` process on Windows. Without that flag `cmd.exe` hands without opening a window. On newer Qt 5.7+ we can use `QProcess::setCreateProcessArgumentsModifier` to specify that flag, so we can skip using `CreateProcessW`. Howeve the question is whether we need compatibility with Qt 5.0 - Qt 5.6 ? If not, then we can avoid the low-level `CreateProcessW`.

The lambdas are used in `ExecProcess` for code reuse. I think I can avoid them, so I will rewrite the code without them.

sunderme:
the only platform with old qt is linux. So, for win, we can assume qt>5.7...

MeanSquaredError:
OK, I will rewrite the code so that it will use `QProcess` and `QProcess::setCreateProcessArgumentsModifier` instead of low-level Windows code.

MeanSquaredError:
OK, I updated the PR. A few notes:

1. I removed the lambda expressions from the internal code of `ExecProgram`. However when running the external terminal application we have to specify the native WinAPI flag `CREATE_NEW_CONSOLE` flag. In order to do that the Qt API relies on a callback and the documentation for `QProcess::setCreateProcessArgumentsModifier` gives an example with a lambda function. So I had to use a lambda function in `Texstudio::openExternalTerminal`. It is a small lambda function and hopefully not confusing.

2. In order to use the `CREATE_NEW_CONSOLE` flag with a detached process we use `QProcess::setCreateProcessArgumentsModifier()`, which in turn requires a `QProcess` object. Therefore we need to run the detached process through that object. Therefore for detached processes we have to use the non-static version of `QProcess::startDetached()`, which appears for the first time in Qt 5.10.0. Therefore for Windows we require Qt 5.10.0+. For other (non-Windows) platforms we still require Qt 5.0.0+.

3. Since `QProcess::startDetached()` requires a `QProcess` object, we have to start the program name and arguments for that `QProcess` object first. However the `QProcess` API only allows setting the arguments as a `QStringList` object. It does not allow setting the program name and arguments as a single `QString` object like we did with the older API.

That is why the constructor 
`ExecProgram::ExecProgram(const QString &shellCommandLine, const QString &additionalSearchPaths, const QString &workingDirectory)`
has to parse the command line and store the program name as a string and the arguments as a `QStringList`. That is why I wrote a simple command-line parser in `ExecProgram::setProgramAndArguments`. It is not perfect but I tested with the Linux/Windows builds and it seems to work fine.

Overall it turned out a bigger change than I expected for a simple terminal fix, but it seems to work OK :-)

sunderme:
okay, i am adding the pr.
tests for the parser may be nice though...
on a general note, lambdas are fine. lambdas in variables used somewhere else may be an issue.

MeanSquaredError:
OK, I will provide shortly a PR with tests for the `ExecProgram` parser.

PR #1098, opened 2020-05-19
Add tests for the ExecProgram command-line parser.

MeanSquaredError:
This PR adds tests for the ExecProgram command-line parser.


MeanSquaredError:
Looks like I forgot to add the actual .cpp and .h files. Now let's see if I still have a copy of them :-)

MeanSquaredError:
OK, I added the missing files, so now the tests should work correctly.

sunderme:
thanks

PR #1102, opened 2020-05-21
Improve debug logger

MeanSquaredError:
This PR improves the debug logger framework slightly by moving the start of the debug logger earlier in the TXS initialization process.

While investigating https://github.com/texstudio-org/texstudio/issues/1036 I used the debug logger framework and I came across a problem with it. At the moment the debug logger activates after the constructor `Texstudio::Texstudio()` is invoked. In that constructor the previous session is restored and the `SyntaxCheck` thread is started. Thus currently the debug logger is not suitable for logging events that happen early on, e.g. `SyntaxCheck::run` events that happen early on.

This PR moves the start of the debug logger earlier before the call to `Texstudio::Texstudio()`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -186,10 +186,10 @@ bool handleCommandLineOnly(const QStringList &cmdLine) {
 int main(int argc, char **argv)
 {
 	QString appId = generateAppId();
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

PR #1115, opened 2020-05-29
Attempt to fix issue #1036

MeanSquaredError:
This PR attempts to fix https://github.com/texstudio-org/texstudio/issues/1036
The user reported high CPU usage happening at random moments while editing a large file.
I (sort of) reproduced the problem on my computer, but in my case the high CPU usage appeared each time immediately after the .tex file was loaded by TXS.
Since the file(s) provided by the user seem to be copyrighted material and I don't know whether he autorizes its distribution, I have prepared a sample file that demonstrates the issue:

[slow_syntax_check.tex.txt](https://github.com/texstudio-org/texstudio/files/4698983/slow_syntax_check.tex.txt)

After opening the file in TXS on my computer the CPU load jumps to about 130% and stays so forever (or until you close the document).
Also if you open the file and scroll to the bottom of the file you can see how on the last line
```
\end{document}
```
The `document` environment name stays red forever, which means that the syntax checker never actually completes and never marks the environment name as valid.

After investigating the issue it seems that the problem is caused by a combination of the following two facts:

1. When syntax check is started for a group of lines, each line is queued separately in the syntax checker through `SyntaxCheck::putLine()`. Each queued line can force the next line to be queued if the environment stack has changed (in these cases the `checkNextLine` signal is sent).
2. When `SyntaxCheck::run()` checks if the next line should be checked through the `checkNextLine` signal, the check is made through using the previously stored environment stack from the previous line (`newLine.prevEnv`).
However this previously stored environment is not fetched at the moment when the check is made but instead it is taken from `newLine.prevEnv` which is the value taken at the moment when `SyntaxCheck::putLine()` was called.

The combination of 1 and 2 means that when we do a full re-check of a file which has `N` lines, this means that we have `N` waves of syntax checked lines that overlap and mess each other's environment stacks and never stop until they reach the end of the file, thus requiring approximately `(N^2)/2` lines to be syntax checked.

Imagine the following file

```
\documentclass{amsart}
\begin{document}
$
1+2
$
\end{document}
```

At step one we would start with empty env. stacks for the lines:

```
<empty>
<empty>
<empty>
<empty>
<empty>
<empty>
```

At step two, after each "wave" (total of 6 syntax checks) has moved we would have the following env. stacks

```
normal
normal
normal
normal
normal
normal
```

Then after each wave moves again (total of 5 syntax checks) we would have the following env. stacks

```
normal
normal/document
normal/math
normal
normal/math
normal
```

Then after one iteration (4 syntax checks) we would have

```
normal
normal/document
normal/document/math
normal/math
normal/math
normal/math
```

Then after 3 syntax checks we would have

```
normal
normal/document
normal/document/math
normal/document/math
normal
normal/math
```

Then after 2 syntax checks we would have

```
normal
normal/document
normal/document/math
normal/document/math
normal/document
normal/math
```

And finally after 1 syntax check we would have

```
normal
normal/document
normal/document/math
normal/document/math
normal/document
normal/document
```

It is possible that I got some of the values in the above example incorrect, but hopefully it demonstrates the problem and how exactly 1 and 2 combine to cause the problem.
For a file of about 65,000 lines, we get up to 2 billion syntax checks until the "waves" stop.

I think the problem can be resolved either by fixing 1 or by fixing 2.

Fixing 1 means that instead of starting N waves, we just remove the environment stack cookie for the updated lines and only queue the first line that has to be updated.
Fixing 2 would imply changing the code in `SyntaxCheck::run()` so that it would fetch the environment and tokens for the previous line on the spot instead of fetching them from `newLine`.

I decided to fix the problem by fixing 1.

The proposed patch seems to fix the problem (you can test it with the supplied sample file). However the issue seems complex enough, so I would appreciate if the other develope(s) review it and validate the solution.

### DIFF
--- a/src/latexdocument.cpp
+++ b/src/latexdocument.cpp
@@ -1246,47 +1246,47 @@ bool LatexDocument::patchStructure(int linenr, int count, bool recheck)
 		parent->addDocToLoad(fname);
 	}
 	//qDebug()<<"leave"<< QTime::currentTime().toString("HH:mm:ss:zzz");
-    if (reRunSuggested && !recheck){
+	if (reRunSuggested && !recheck){
 		patchStructure(0, -1, true); // expensive solution for handling changed packages (and hence command definitions)
-    }
-    if(!recheck){
-        // do syntax check after recheck !
-        bool initialRun=false;
-        if (syntaxChecking && languageIsLatexLike()) {
-            for(int i = lineNrStart; i < linenr + count; ++i){
-                QDocumentLineHandle *dlh = line(i).handle();
-                if (!dlh)
-                    continue; //non-existing line ...
-                StackEnvironment env;
-                getEnv(i, env);
-                QDocumentLineHandle *lastHandle = nullptr;
-                TokenStack oldRemainder;
-                if (i > 0) {
-                    lastHandle = line(i - 1).handle();
-                }else{
-                    // check whether syntax checker has ever run
-                    if(!dlh->hasCookie(QDocumentLine::STACK_ENVIRONMENT_COOKIE)){
-                        // initial run
-                        // it is sufficient to put first line
-                        SynChecker.putLine(dlh, env, oldRemainder, true);
-                        initialRun=true;
-                        break;
-                    }
-                }
-                if (lastHandle) {
-                    oldRemainder = lastHandle->getCookieLocked(QDocumentLine::LEXER_REMAINDER_COOKIE).value<TokenStack >();
-                }
+	}
+	if(!recheck){
+		// do syntax check after recheck !
+		bool initialRun=false;
+		if (syntaxChecking && languageIsLatexLike()) {
+			for(int i = lineNrStart; i < linenr + count; ++i){
+				QDocumentLineHandle *dlh = line(i).handle();
+				if (!dlh)
+					continue; //non-existing line ...
+				StackEnvironment env;
+				getEnv(i, env);
+				QDocumentLineHandle *lastHandle = nullptr;
+				TokenStack oldRemainder;
+				if (i > 0) {
+					lastHandle = line(i - 1).handle();
+				}else{
+					// check whether syntax checker has ever run
+					if(!dlh->hasCookie(QDocumentLine::STACK_ENVIRONMENT_COOKIE)){
+						// initial run
+						// it is sufficient to put first line
+						SynChecker.putLine(dlh, env, oldRemainder, true, i);
+						initialRun=true;
+						break;
+					}
+				}
+				if (lastHandle) {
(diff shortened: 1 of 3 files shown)


sunderme:
I see the issue with 2.12.22 but not with HEAD

MeanSquaredError:
I sent an email to the person who originally reported the problem asking him if it is OK for me to send you an archive with the .tex files that he sent me earlier. Once he confirms that it is OK, I will send you that archive. Hopefully it will display the problem on your computer too.

MeanSquaredError:
This is strange because the problem occurs for me with the master HEAD from github.

While waiting for the reply from  ap0305-thunder, one thought occurred to me, did you rename `slow_syntax_check.tex.txt` to `slow_syntax_check.tex` (effectively removing the `.txt` suffix from the filename) ? I only added the .txt suffix because github did not allow me to attach the `.tex` file.

I think that opening a .txt file (as opposed to .tex) will not activate the actual syntax checker.

sunderme:
yes, the file is renamed to ....tex

MeanSquaredError:
That is strange because the problem shows up on my computer under Linux with the master HEAD and the sample file. Could be some difference in the thread speed (main thread vs. syntax checker thread).

Let me try to create another more complex sample file (while waiting for the author to give permission to send the original files).

MeanSquaredError:
@sunderme 

I tested my example file with the debug and release builds of the latest head. The debug build shows the problem and the release build does not show the problem.

However I managed to build another (sort of) minimal example from the files that the user sent me and that example shows the problem both with the debug and release builds.

Do you have an email where I could send you the minimal example? If you don't want to post the email publicly for spam reasons, you can email me at meansquarederror@protonmail.com and I will send you back the archive with the minimal example.

sunderme:
you can send info to sunderme@gmx.de

MeanSquaredError:
I emailed you the test files for the syntax checker. On
my computer it triggers the problem. The email also explains how to open the files and trigger the problem (the way it works on my computer).

I have recorded a screencast showing the problem on my computer:

https://youtu.be/KFKPRq4lglc

(the test has been made with the latest TXS head built in Release mode)

sunderme:
i have not yet received any mail ...

MeanSquaredError:
It was sent from my personal domain so maybe it was filtered by some mail server along the way. It has a 400 KB attachment, so maybe this also caused the problem.

I just re-sent the email from my gmail account. Please let me know if it arrived in your mailbox.

sunderme:
okay, looks good.
Thanks

ap0305-thunder:
> 
> 
> I sent an email to the person who originally reported the problem asking him if it is OK for me to send you an archive with the .tex files that he sent me earlier. Once he confirms that it is OK, I will send you that archive. Hopefully it will display the problem on your computer too.

yes please, you can send what I sent to you. thanks

PR #1116, opened 2020-05-30
Log the current thread ID in the debug logger

MeanSquaredError:
This PR adds logging of the current thread ID to the debug logger. It is a minor improvement useful when investigating multi-threaded issues. I had to add it when investigating the syntax checker issue (because the syntax checker uses multiple threads).


sunderme:
thanks

Issue #593, opened 2019-05-07
Where is the latest manual hosted on?


muzimuzhi:
Currently, the links of manual of texstudio, both from the README of this repo and from texstudio.org, point to the same webpage hosted on sourceforge: http://texstudio.sourceforge.net/manual/current/usermanual_en.html .

However, the manual on sourceforge is outdated. For example, the explanation about supported `% !TeX program = <engine>`, added in this commit https://github.com/texstudio-org/texstudio/commit/7d5e836cb4ff60a1415d0433861423e0f41f1ae9, can not been found in the manual on sourceforge.

Hence the question:
 - Is the latest version of manual hosted?
 - If yes, where can I visit it, rather than open the html file from local clone?

manudroid19:
Looks like it is not hosted, good point. @sunderme should we consider moving this manual to the wiki, the official website or a github.io webpage?

sunderme:
Usually i would use the code git but if that is not feasible, I am open for suggestions...

manudroid19:
The regular user won't be looking in git, and for sure the won't look in the source code. The website looks like the most appropiate place. I could get to it and also improve the style of the webpage in one month aprox.

muzimuzhi:
I just found that menu `Help -> User Manual` will open local `usermanual_en.html`. This would be the best way for *regular user* to open the user manual, especially it corresponds to the locally installed version of texstudio.

At the same time, updating the hosted user manual to the latest version, or with multiple versions corresponding to main changes of user manual or stable releases of texstudio, still has its significance.

muzimuzhi:
From https://github.com/texstudio-org/texstudio/pull/1119#issuecomment-649598788, it seems that users of texstudio are more aware of online user manual hosted on [sourceforge](http://texstudio.sourceforge.net/manual/current/usermanual_en.html) than the one bundles with texstudio itself.

This raised the need to keep sourceforge manual up-to-date (to the latest stable release).

sunderme:
changed texstudio.org to link to github/master directly

muzimuzhi:
Thanks. Some more things:

1. [`htmlpreview`'s README](https://github.com/htmlpreview/htmlpreview.github.com#usage) suggests to append a`github.com` url to `https://htmlpreview.github.io/?`. Currently a `raw.githubusercontent.com` url is appended. Not sure if there is any preference between them.
[...]

PR #1119, opened 2020-06-02
More completions for siunitx command options.

ElieGouzien:
I added lot of command option for `siunitx.cwl`.
Some allowed options have not been included where they have no effects.

Full list of option has been set for \sisetup ; for the other commands
only the relevant subset has been kept.

Supported siunitx version : 2.8b

### DIFF
--- a/completion/siunitx.cwl
+++ b/completion/siunitx.cwl
@@ -4,17 +4,109 @@
 \A#*
 \ampere
 \amu#*
-\ang[%<options%>]{%<degrees%>}
-\ang[%<options%>]{%<decimal number%>}
 \ang{%<degrees%>}
 \ang{%<decimal number%>}
+\ang[%<options%>]{%<degrees%>}
+\ang[%<options%>]{%<decimal number%>}
 #keyvals:\ang
+# Detecting Fonts ; section 5.2 : ignored
+# Font settings ; section 5.3
+color=#%color
+mode=#math,text
+math-rm=#*
+math-sf=#*
+math-tt=#*
+text-rm=#*
+text-sf=#*
+text-tt=#*
+# Font settings ; section 5.3 ; with "number-"
+number-color=#%color
+number-mode=#math,text
+number-math-rm=#*
+number-math-sf=#*
+number-math-tt=#*
+number-text-rm=#*
+number-text-sf=#*
+number-text-tt=#*
+# Font settings ; section 5.3 ; with "unit-"
+unit-color=#%color
+unit-mode=#math,text
+unit-math-rm=#*
+unit-math-sf=#*
+unit-math-tt=#*
+unit-text-rm=#*
+unit-text-sf=#*
+unit-text-tt=#*
+# Parsing numbers ; section 5.4
+input-close-uncertainty=
+input-comparators=
+input-complex-roots=
+input-decimal-markers=
+input-digits=
+input-exponent-markers=
+input-ignore=
+input-open-uncertainty=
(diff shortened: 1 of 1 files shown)


MeanSquaredError:
Seems to mess up a couple of `SyntaxCheckTest::checkkeyval` tests. I guess these tests will need adjustments.

muzimuzhi:
BTW, cwl now supports syntax `#keyvals:\cmd[,\cmd]*`. For example, you can use `#keyvals:\ang,\num,\anotherCmd`. This makes cwl file more compact.

ElieGouzien:
> BTW, cwl now supports syntax `#keyvals:\cmd[,\cmd]*`. For example, you can use `#keyvals:\ang,\num,\anotherCmd`. This makes cwl file more compact.

Good to know! Can we define options for a single command in different `#keyvals` sections ? Typically here I would need to group options in `#keyvals` sections looking to something like `#keyval:\num,\SI`, `#keyval:\si,\SI`…

muzimuzhi:
> Can we define options for a single command in different #keyvals sections ?

Yes

sunderme:
I think the lines 'math-sf=#*' is probably unintentional incorrect.
In context of key/vals it means that only "*" is a valid option (that why one of the test fails).
I suppose it was meant to say "unusual" command, which is not defined in this context, i.e. for key/vals.

ElieGouzien:
It should be better now, and also easier to maintain; I grouped all the options by section of the documentation and never repeated anyone.
If you want I can squash last commit into the first one (to avoid +1000 lines in the history).

> I think the lines 'math-sf=#*' is probably unintentional incorrect.
> In context of key/vals it means that only "*" is a valid option (that is why one of the test fails).

Exact and corrected, except that what made failing the test was rather to have comments between `keyvals` and `endkeyvals` (what might not be really convenient).

sunderme:
Thanks 👍

muzimuzhi:
> Exact and corrected, except that what made failing the test was rather to have comments between `keyvals` and `endkeyvals` (what might not be really convenient).

@ElieGouzien After #1143, I think using comments between `#keyvals` and `#endkeyvals` is now supported (so there is no need to split keyvals into groups). Welcome to test, if you could [compile from source](https://github.com/texstudio-org/texstudio/wiki/Compiling) or after the next alpha/beta release of 3.0.0.

ElieGouzien:
Nice that you fixed this! A 10 s check indicates that it's now OK.

I don't think I will modify this pull request, since the split is also needed because each options section has a specific set of commands for which they have  non-trivial effect, but anyway it's a valuable improvement.

Maybe documenting the several command options (`#keyvals:\cmd[,\cmd]*`) would also be really useful.

muzimuzhi:
> Maybe documenting the several command options (`#keyvals:\cmd[,\cmd]*`) would also be really useful.

The user manual released with texstudio v2.12.18 or newer already contains
> `#keyvals:<command[,command,...]>` (at start of line)

https://github.com/texstudio-org/texstudio/blob/6ff4aa3bb043155edad8114a36e36fd3b804fde6/utilities/manual/usermanual_en.html#L1579
which is added in https://github.com/texstudio-org/texstudio/commit/18ed1a2fbc07f07f600a715a131b96ee8b420b64 (Aug 2019). 

The problem is that the online user manual hosted on [sourceforge](http://texstudio.sourceforge.net/manual/current/usermanual_en.html) is outdated. This has been reported in #593 (May 2019).

PR #1121, opened 2020-06-03
Add write locking to fix crashes when accessing document line cookies

MeanSquaredError:
This PR adds `QDocumentLine` write locks in `LatexDocument::reCheckSyntax()`

I changed that method while working on the fix for the high CPU load but forgot to add the write locking (although was planning to do so).

Noticed today that `texstudio --execute-all-tests` crashes if the tests run while the computer is heavily loaded with other tasks too. Investigation showed that crashes seem to be caused by the lack of locking in `LatexDocument::reCheckSyntax()`

Adding the locks fixes the crashes for me, but I guess we will have to keep an eye on the new code for a while.


sunderme:
Thanks

PR #1122, opened 2020-06-05
Some cwl changes

muzimuzhi:
Main changes:
 - Record more macros in `ulem.cwl`
 - Fix semantics of argument name "class" in and remove duplicate lines from `titlesec.cwl`
 - Make minor changes to `etoolbox.cwl`
 - Proofread `latex-document.cwl`, remove useless lines and move lines into separate (new) cwl files
    - [reverted] Remove line `%<%:TEXSTUDIO-GENERIC-ENVIRONMENT-TEMPLATE%>`
    - [reverted] Remove environment forms of font size changing commands (`\Huge`, `\huge`, `\LARGE`, `\Large`, `\large`, `\normalsize`, `\small`, `\footnotesize`, `\scriptsize`, `\tiny`)
    - Remove environment `theorem`. It is not defined by LaTeX2e.
    - Remove math environments `Vmatrix` and `vmatrix`. They are defined by `amsmath` package, not by LaTeX2e.
    - Add line `# combinations` to mark the start of combinations like `\left\lbrace`


completion/latex-document.cwl
```diff
@@ -34,17 +32,13 @@
 \begin{figure*}
 \begin{figure*}[placement]
 \begin{filecontents}
+\begin{filecontents}[options%keyvals]
 \begin{filecontents*}
+\begin{filecontents*}[options%keyvals]
 \begin{flushleft}
 \begin{flushright}
-\begin{footnotesize}
```
sunderme, line 40: any reason why this construct is removed ?
from what I read , this is valid and not defined somewhere else ?

completion/latex-document.cwl
```diff
@@ -14,9 +14,7 @@
 \appendixname#*
 \arabic{counter}#*
 \author{names}
-%<%:TEXSTUDIO-GENERIC-ENVIRONMENT-TEMPLATE%>
```
sunderme, line 17: this line generates the general "\begin{*environment*}"
The auto test fails because this line missing.
Please don't remove stuff just because it is not clear what it is doing ...

completion/latex-document.cwl
dbitouze: Characters size changing by the way of environments should be discouraged. See https://sourceforge.net/p/texstudio/feature-requests/130/.

sunderme: It is not the editor's task to enforce a style of writing. 
If a valid construct is suddenly marked as syntax error in a new txs version, people will complain.

dbitouze: IMO, it could be, and even it should be, the task of a TeX editor, designed by people who are aware of good practices, to discourage bad practices. Maybe some users would complain but others would be grateful to learn something.

I agree it would be nice in such a case to display (e.g. with a tooltip) why it is now marked as syntax error ("Leads to spurious spaces" should be enough) and what is the replacing good practice (E.g. "It is advised to replace `\begin{footnotesize}...\end{footnotesize}` by {\footnotesize ...}, see LaTeX → Font Sizes menu").

The same should happen for `{\bf ...}` versus `\textbf{...}`, `eqnarray` versus `align`, etc. (See https://www.ctan.org/search?phrase=l2tabu).

muzimuzhi: Reverted. 

Although the environment forms of font size changing commands are not mentioned in _LaTeX: A Document Preparation System_ by Lamport and _The LaTeX Companion_, 2nd edition, other good LaTeX documents _do_ mention those usages. For example, _LaTeX2e: An unofficial reference manual, [sec. 4.2](https://latexref.xyz/Font-styles.html#Font-styles) and _LaTeX for Complete Novices_ (Dickimaw LaTeX Series), [sec. 4.5.2](https://www.dickimaw-books.com/latex/novices/html/fontsize.html). Also, those environment forms are not marked obsolete in documentation [`l2tabu`](https://ctan.org/pkg/l2tabu).

Furthermore, it seems that using an environment to restrict the scope of font size changing is reasonable (perhaps also straight-forward), therefore I have reverted those changes.

Thanks.

completion/latex-document.cwl
muzimuzhi: Reverted. Sorry.

completion/latex-document.cwl
muzimuzhi: Some arguments about "valid construct" (https://github.com/texstudio-org/texstudio/pull/1122#discussion_r436275627) and "bad practices" (https://github.com/texstudio-org/texstudio/pull/1122#discussion_r436276987):

* Inside `\end{<envname>}`, `\csname end#1\endcsname` is used to construct command `\end<envname>`, and because of the `\csname`, `\end<envname>` is let to `\relax` if it is not defined. This means `\end{<envname>}` will never complains about `undefined control sequence `\end<envname>`. 
* So in general one can think every command have a corresponding environment form, no matter the `\end<envname>` is defined or not. 
* Therefore it is hard to say whether `\begin{small} ... \end{small}` is a valid construct or not, but IMO (and from the references mentioned in my previous comment) it is not a bad practice, at least not too bad.

muzimuzhi:
I have appended a detailed changelog of `latex-document.cwl` in https://github.com/texstudio-org/texstudio/pull/1122#issue-428810977.

sunderme:
thanks

completion/latex-document.cwl
dbitouze: I guess many users would use `small`, `footnotesize`, etc. environments the following way:
```latex
\begin{tiny}
Foo bar.
\end{tiny}
```
which could lead to spurious spaces (see MWE below). These spurious spaces are avoided only if these environments would be used the following way:
```latex
\begin{tiny}Foo bar.\end{tiny}
```
but I guess only a few users would be aware of this subtlety. And, even if TXS would insert such environments on a single lines, how many users would add line breaks in order to come back to the usual form:
```latex
\begin{tiny}
Foo bar.
\end{tiny}
```
**MWE**
```latex
\documentclass{article}
\begin{document}
Foo bar.
\begin{tiny}
Foo bar.
\end{tiny}
Foo bar.

Foo bar. \begin{tiny}
Foo bar.
\end{tiny} Foo bar.

Foo bar. \begin{tiny} Foo bar. \end{tiny} Foo bar.

Foo bar. \begin{tiny}Foo bar.\end{tiny} Foo bar.

Foo bar. {\tiny Foo bar.} Foo bar.

Foo bar.
\begin{tiny}Foo bar.\end{tiny}
Foo bar.

Foo bar.
{\tiny Foo bar.}
Foo bar.
\end{document}
```
test

muzimuzhi: @dbitouze Your example shows that using env-forms inline requires special care. On the other hand, env-forms might be helpful in changing font size for entire paragraph(s). But since these days I seldom write tex documents apart from small examples, my opinion about good or bad practices may not be representative.

Those env-forms are added by [jsundermeyer's commit r1848](https://sourceforge.net/p/texstudio/code/1848/) in April 2011, a commit changed/added only those lines. So I guess there was indeed a need to add them. To change a long-standing but controversial thing often requires strong reason and careful considerations about forward compatibility. Therefore, for now I am OK with either of the decisions (keep or remove).

dbitouze: @muzimuzhi For entire paragraph(s), one can always use the TXS feature: select it (or them) and go to LaTeX → Font Sizes menu: that will surround the paragraph(s) with a pair of curly braces, the opening one followed by the desired font size declaration command (e.g. `{\footnotesize ...}`).

dbitouze: > Also, those environment forms are not marked obsolete in documentation [`l2tabu`](https://ctan.org/pkg/l2tabu)

Indeed but they are raised in the [`nag`](https://ctan.org/pkg/nag) documentation package:

>**3 Switch vs. Environment**
People often use switches as environments and vice versa. This is dangerous in
because it tends to almost work. (Consider font size commands in particular, but also \centering vs. center environment.) As usual, “it’s not an error if
you know what you’re doing”. [...]

Issue #1127, opened 2020-06-13
Crash Report: Tikz ing of background


ccaprani:
TexStudio crashes instantly upon editing line 25 of the attached MWE and no backtrace is produced.

### Environment
- TeXstudio: 2.12.22
- Qt: 5.12.1
- OS: Windows 10
- TeX distribution: MikTex

### Expected behavior
Should not crash, compiles the changes and renders.

### Actual behavior
TexStudio instantly vanishes and closes without any saving of open state or backtrace as usually produced.

### How to reproduce
Use attached file. Change extension to `*.tex`. Edit the tile size on Line 25 of the attached. Re-compile once or twice. TexStudio insta-crashes.

[barcode_test_crash.txt](https://github.com/texstudio-org/texstudio/files/4774180/barcode_test_crash.txt)

muzimuzhi:
Not re-producible on macOS 10.15.5.

PS: I change line 25 from `(-2cm,0) rectangle (\textwidth,4cm);` to `(-2cm,0) rectangle (.5\textwidth,4cm);` or `(-2cm,0) rectangle (.5\textwidth,6cm);`.

PS 2: Have you tried [texstudio 3.0.0alpha6](https://github.com/texstudio-org/texstudio/releases/tag/3.0.0alpha6)?

sunderme:
with 2.12.22 I see a problem with forward sync
It is slow on 3.0.0 but it seems to work.

ccaprani:
Tried TexStudio3, changed:
> `(-2cm,0) rectangle (\textwidth,4cm);` to `(-2cm,0) rectangle (\textwidth,5cm);`
Recompile twice and again insta-crash.

FYI, `(-2cm,0) rectangle (\textwidth,8cm);` gives the TeX memory capacity exceeded error. Perhaps it's a very edge case LaTeX code, but I guess TS should still exit gracefully.

muzimuzhi:
> FYI, `(-2cm,0) rectangle (\textwidth,8cm);` gives the TeX memory capacity exceeded error. Perhaps it's a very edge case LaTeX code, but I guess TS should still exit gracefully.

`(-2cm,0) rectangle (\textwidth,8cm);` works well on my local texlive 2020 on macOS, and on overleaf.com which runs a texlive 2019 by default. In each case I test each of pdftex, xetex, and luatex engine.

So, which engine are you using? And what is the version of that engine? You can find these info in the first line of `.log` file, for example:
```tex
[...]

PR #1133, opened 2020-06-19
Remove recursion in _synctex_free_handle

MeanSquaredError:
This PR updates the code _synctex_free_handle() to the new version provided by jlaurens from the synctex development team.

It greatly alleviates the crash issue reported in https://github.com/texstudio-org/texstudio/issues/1127

There still seems to remain some kind of bug which causes crashes but it is much rarer and difficult to reproduce on my Windows 7 machine. Before this PR I was able to reproduce the crash reliably after each time in the .tex source. After the PR it takes me more than half an hour or random changes and rebuild&view (F5) from the TXS UI before it actually crashes (and sometimes it takes more than an hour).

Since I am unable to reproduce the remaining crashes reliably, I think they represent some other bug which I am unable to catch yet and I suspect that the other bug is not related to synctex.

### DIFF
--- a/src/pdfviewer/synctex/synctex_parser.c
+++ b/src/pdfviewer/synctex/synctex_parser.c
@@ -1,15 +1,15 @@
 /*
  Copyright (c) 2008-2017 jerome DOT laurens AT u-bourgogne DOT fr
- 
+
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #1137, opened 2020-06-21
Some cwl changes, new file titlesec.cwl and others

muzimuzhi:
Main changes:
 - Delete lines from `titlesec.cwl` which belongs to pacakge `titletoc`.
  Note that `titlesec` does not load `titletoc`.
 - Add new file `titletoc.cwl`. more variants with or without optional arguments are added based on the lines from `titlesec.cwl`.
 - Add comment to `%<%:TEXSTUDIO-GENERIC-ENVIRONMENT-TEMPLATE%>`, related to #1122.
 - Remove duplicate lines from `xepersian.cwl`, which are already recorded in `latex-document.cwl`.
 - Unify argument name used in `tcolorbox.cwl` to that in `latex-document.cwl`. 
  For example, now `args` is used uniformly in `\newtcolorbox{envname}[args][default]{options%keyvals}` and `\newcommand{cmd}[args][default]{def}`.


sunderme:
thanks

PR #1143, opened 2020-06-25
cwl: ignore comment line in keyvals list

muzimuzhi:
Currently, the following cwl input causes empty autocomplete pop-up for `\cmd{<cursor pos>}`:
```
\cmd{options%keyvals}
#keyvals
# comment
key1
key2
#endkeyvals
```

This problem is reported, for example in https://github.com/texstudio-org/texstudio/pull/1119#issuecomment-638473385. This pr tries to fix it.

The only cwl keyword that is treated below the treatment for keyvals list is `#repl:`. Maybe it should be moved above?


sunderme:
thanks

PR #1155, opened 2020-07-02
arm 32 and 64 bit support for Linux and NetBSD ( fixes #739 )

### LINKED ISSUE #739: fails to compile for ARM
### Environment
- TeXstudio: 2.12.16
- Qt: 5
- OS: Raspbian 10
- TeX distribution: N.A.

### Expected behavior
Compilation of source code for arm7.

### Actual behavior
compile error in debughelper.cpp/SimulatedCPU::set_all

### How to reproduce
see https://build.opensuse.org/package/live_build_log/home:jsundermeyer/texstudio-2.12.16/Raspbian_10/armv7l

matthewberryman:
Tested on Ubuntu 20.04 aarch64 build, and Raspberry Pi OS Buster (arm 32 bit).
I've not had much luck testing on NetBSD—the AMI image I was using had standard processes dying for no apparent reason, however once I get my monitor I will test. In the mean time I have double checked it against the definition of mcontext used in NetBSD per https://github.com/NetBSD/src/blob/bf21c9368813f0484e699e8472967c8dc2ca4f68/sys/arch/arm/include/mcontext.h#L108 and fixed a typo in one case and double checked it through a code review so I'm pretty confident it will work.

### DIFF
--- a/src/debug/debughelper.cpp
+++ b/src/debug/debughelper.cpp
@@ -58,7 +58,7 @@ struct SimulatedCPU {
 	char *frame;  //e.g. ebp, r11
 	char *stack;  //e.g. esp, r13
 #if defined(CPU_IS_ARM) || defined(CPU_IS_ARM64) || defined(CPU_IS_MIPS) || defined(CPU_IS_PPC) || defined(CPU_IS_SPARC32)
-	char *returnTo;  //lr/r14
+	char *returnTo;  //lr/r14(arm)/x30(arm64)
 #endif
 
 	inline void push(char *value)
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

MeanSquaredError:
@matthewberryman 
Thank you for the PR. Overall it looks good to me.

I had some doubts whether we could re-use the same
```
void SimulatedCPU::call(char *value)
```
for arm and aarch64 (on line 1179).
My doubts were mostly because the inline assembler uses the `fp` and `lr` registers and these register names are not really used on aarch64. However after some digging it seems that for aarch64 many (probably most?) assemblers alias `fp` and `lr` to `x29` and `x30` accordingly. So given that it compiles correctly, I guess the gnu assembler does that kind of aliasing too, so we should be OK using the same assembler code.

Might break on some exotic non-gcc compilers, but realistically the TXS codebase as a whole supports only gcc anyway :-)

matthewberryman:
Yes it all seems fine on gcc from my understanding and also compiling and testing. I understand from the BUILD.sh that clang+LLVM is used on macOS—I think that should also be fine but will test on macOS x86_64 and aarch64 in due course... I have applied for test Mac aarch64 kit to hopefully speed that side up.
________________________________

sunderme:
it builds on aarch64

PR #1158, opened 2020-07-02
Spurious double space fixed

dbitouze:
This double space is visible not only in `tags/asymptote_tags.xml` (changed in this PR) but also in `src/additionaltranslations.cpp` but, strangely enough, the latter file is not considered as changed by Git when I edit it.


sunderme:
Thanks

PR #1175, opened 2020-07-14
[cwl] fix wrong order of classifiers

muzimuzhi:
From user manual, sec. 4.14.3, if multiple classifiers are used, `*` should be the first one. For exampleg `\LoadClass{class}#u*` should be `\LoadClass{class}#*u`.

For users, this PR fixes the problem that commands like `\LoadClass` are marked unrecognized.


sunderme:
thanks

PR #1192, opened 2020-07-30
Some minor cwl changes

muzimuzhi:
For more detail, see the commit messages.

### DIFF
--- a/completion/class-beamer.cwl
+++ b/completion/class-beamer.cwl
@@ -336,19 +336,13 @@
 \keywords{text}
 \section<mode specification>[short title]{title}
 \section<mode specification>{title}
-\section<mode specification>*[short title]{title}
 \section<mode specification>*{title}
-\section*[short title]{title}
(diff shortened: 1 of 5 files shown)


muzimuzhi:
> why is section* removed ?

Since starred sections doesn't accept optional short title, see `texdoc beamer`, sec 10.2. This rule is inherited from standard document classes and `\section*[...]{..}` is not recorded in `latex-document.cwl` as well. 
https://github.com/texstudio-org/texstudio/blob/a0c5707c17d6c78fb6ff5daba3c0b7727b872dae/completion/latex-document.cwl#L361-L364

Also see the following example.
```tex
\documentclass{beamer}

\begin{document}
% the following line produces the first frame containing text "short title]title", 
% since beamer treats "[" as mandatory argument of \section*
\section*[short title]{title}

\begin{frame}
  content
\end{frame}
\end{document}
```

sunderme:
Thanks

PR #783, opened 2019-11-03
Make "tex.cwl" contain full list of tex primitives and plain tex commands

muzimuzhi:
This pr follows the discussion in #751 .

Main changes:
 - Manually review `tex.cwl`, make it contain _full list_ of tex primitives and plain tex commands.
 - Remove duplicates in other cwl files, resulting the `latex-mathsymbols.cwl` nearly empty.
 - Delete `latex-mathsymbols.cwl` and clear codes using it.

When reviewing, I take the list in [_TeX for the Impatient_](https://www.gnu.org/software/teximpatient/), Chap. 13 as the main reference.

Issue #751, opened 2019-10-14
Need some boundaries between engine and format related cwl files


muzimuzhi:
Currently, we have had several cwl files recording engine primitives and format macros. For engine primitives,
 - original tex primitives are in `tex.cwl`
 - etex [1] add-ons are missing (Adding them is an easy task when boundaries are drawn.)
 - pdftex add-ons are in `latex-dev.cwl`
 - xetex and luatex add-ons are in corresponding `xetex.cwl` and `luatex.cwl`

For format macros,
 - plain tex macros are mainly in `tex.cwl`, and the rest are in `latex-dev.cwl`
 - latex2e macros are in `latex-dev.cwl`, `latex-mathsymbols.cwl` and `latex-document.cwl`
 - ConTeXt macros are missing (As I guess the number of ConTeXt users is rather small.)

And the cwl files loaded by default are (as far as I understood)
https://github.com/texstudio-org/texstudio/blob/3e9181d8421bbdf5daf023c79161e2af9d95025e/src/configmanager.cpp#L905-L906

My suggestion:
 - Primitives
   - Add etex add-ons as `etex.cwl`.
   - Move pdftex add-ons from `latex-dev.cwl` to new file `pdftex.cwl`.
   - If texstudio can be enhanced to load different engine cwl, for example, according to the current default compiler (`txs:///compiler`), load the corresponding cwl file. If not (yet), load `pdftex.cwl` by default.
   - Each of `pdftex.cwl`, `xetex.cwl` and `luatex.cwl` depends on `etex.cwl`. (see the note [1])
 - Format macros
   - latex2e supports most of the plain tex macros. Move these macros to `latex-dev.cwl`; for those familiar to normal latex users (rather than package authors), move them to `latex-document.cwl` or `latex-mathsymbols.cwl`.
   - If we do need a plain tex cwl file, or in the future texstudio can auto-detect the format, add full plain tex macros as `plaintex.cwl`. PS: some (less than ten) of the plain tex macros are not supported by latex2e, see [`texdoc source2e`](http://mirrors.ctan.org/macros/latex/base/source2e.pdf), sec. 9.
    - For ConTeXt, add `context.cwl` if the similar need or auto-detection function is ready.

This issue is mainly for discussion. Once the boundaries are drawn and decisions are made, my contributions are ready to go :).
------
[1] Intro to etex engine
[...]

PR #1200, opened 2020-08-07
[cwl] Five new files, multimedia.cwl updated, and more

muzimuzhi:
Main changes:
 - Added five new files
    - `incgraph.cwl`
    - `pgffor.cwl` (just for finer dependency of tikz/pgf bundle, see below)
    - `plaintex.cwl` (for storing commands which are available in plain TeX format only)
    - `thmtools.cwl`
    - `xmpmulti.cwl` (one of `beamer`'s subpackages)
 - Reorganized dependency of cwl of packages in tikz/pgf bundle.
 - Updated `multimedia.cwl`, another subpackage of `beamer`.

PS: some of the lines in `xmpmulti.cwl` added and `multimedia.cwl` updated are duplicate with some of the lines in `beamer.cwl`, and this will be fixed when I complete the update for `beamer.cwl` (maybe in next pr).


completion/plaintex.cwl
```diff
@@ -0,0 +1,4 @@
+\ttfam#*
```
sunderme, line 1: why is this moved to a separate file ?

completion/pgffor.cwl
```diff
@@ -0,0 +1,9 @@
+# pgffor package
+# muzimuzhi, 8 Aug 2020, pgffor in tikz v3.1.5b
```
sunderme, line 2: why separate file ?
Is there any scenario where pgffor is separated from pgfcore ?

completion/plaintex.cwl
muzimuzhi: These lines are added to `tex.cwl` in pr #783 opened by myself.

1. All four `\xxfam` commands are macros, not primitive, and are never defined in LaTeX format but only in plainTeX format. 
1. Moreover, I find having them in autocompletion makes the autocompletion less convenient. For example when I type `\ttf` and hit TAB, what I expect is to have `\ttfamily` autocompleted, but with `\ttfam` there I must type more characters to distinguish `\ttfam` from `\ttfamily. (I always set autocompletion "all".)
1. There might be more plainTeX macros currently recorded in `tex.cwl`, but these four are the most annoying ones hence I removed them. 

In most cases, texstudio is used to edit a tex file written in LaTeX format (starting with `\documentclass`), so this change is transparent to most users. PlainTeX users can check `plaintex.cwl` in configuration to add support for plainTeX, hence this change would not remove any utilities.

So in short, remove from `tex.cwl` for LaTeX users, and move to separate `plaintex.cwl` for plainTeX users (and a bit of backward compatibility).

completion/pgffor.cwl
muzimuzhi: Sometimes, yes. As the `pgfmanual` says in sec. 88,

> This package (`pgffor`) can be used independently of PGF. 

Personally I sometimes load `pgffor` without PGF/tikz, mostly when preparing small examples as an answer to online LaTeX questions. In the future I might add more lines to `pgffor.cwl` if possible. Will deeper dependency of cwl files have obvious impact on the user experience?

muzimuzhi:
PS: the CI fails when building on macOS ([link](https://travis-ci.org/github/texstudio-org/texstudio/jobs/715945782#L151))

```
==> cmake . -DBUILD_GTK_TESTS=OFF -DENABLE_CMS=lcms2 -DENABLE_GLIB=ON -DENABLE_QT5=ON -DENABLE_UNSTABLE_API_ABI_HEADERS=ON -DWITH_GObjectIntrospection=ON
==> make install
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

completion/pgffor.cwl
sunderme: Okay, this is fine

completion/plaintex.cwl
sunderme: I am a bit confused why it should be moved from a cwl which is already named tex.cwl.

sunderme: Any particular reason not use "typical" completion ?

muzimuzhi: (Sorry it takes me too long to compose and adjust words and sentences to express my intension.)

> I am a bit confused why it should be moved from a cwl which is already named tex.cwl.

Not sure if I understand your confusion, but I guess this is similar to #751.

Concerning the 3 auto-loaded cwl files (`tex.cwl`, `latex-document.cwl` and `latex-dev.cwl`), what they should provide (and provide only) can be split into two parts:
 - the primitives provided by a tex engine, and
 - the commands (including environments) provided by a tex format.

The most common combination is pdftex engine + latex format, and this is what the most lines in those three cwl files do.
 - To add support for engines other than pdftex, separate files [`xetex.cwl`](https://github.com/texstudio-org/texstudio/blob/master/completion/xetex.cwl) and [`luatex.cwl`](https://github.com/texstudio-org/texstudio/blob/master/completion/luatex.cwl) were added years ago. And these two cwl files are not auto-loaded.
 - To add support for formats other than latex,  my decision in pr #783 was to add lines for plainTeX format into `tex.cwl`. But I found this inconvenient, hence in this pr I started to move plainTeX-only lines to separate file `plaintex.cwl`.

This change makes handling for different formats the same for the handling for different engines.

BTW, in practice it is very very rare that one `.tex` file is deliberately made compilable in different formats (compilable both by executable `pdftex` and `pdflatex`), but it is not such rare that one `.tex` file is compilable in different engines. Therefore recording commands for different formats in different files dose, in most cases, no harm.

> Any particular reason not use "typical" completion ?

I need to type in macros marked `#*` in cwl (used for completion only in with the "all" tab) too often. Perhaps I am too far away from typical latex users.

sunderme: Are you aware that txs counts completion usage and that rare commands appear in typical if they are completed a few times ?

I see that you yourself added these commands ... Are you sure that not even more can be moved ? Default should contain stuff which most users need ...

muzimuzhi: > Are you aware that txs counts completion usage and that rare commands appear in typical if they are completed a few times ?

Thanks. I will have a try.

> I see that you yourself added these commands ... Are you sure that not even more can be moved ?

There are definitely more can be moved, though currently this pr just moves the most annoying ones. I've found a recent doc ([`tex-nutshell.pdf`](http://mirrors.ctan.org/info/tex-nutshell/tex-nutshell.pdf)) which has a summary of plainTeX macros in sec. 14.

If we both agree that plainTeX-only macros can be moved from `tex.cwl` to `plaintex.cwl`, I can do the rest moving things in one or two days. This way related changes are kept in single pr.

> Default should contain stuff which most users need ...

Most users don't use plainTeX.

sunderme: Agreed

muzimuzhi: > [...] I can do the rest moving things in one or two days. This way related changes are kept in single pr.

The moving work has been done in #1203, so I'll mark this conversation as resolved.

PR #1203, opened 2020-08-09
[cwl] move all plaintex-only macros to plaintex.cwl

muzimuzhi:
This pr follows discussions in https://github.com/texstudio-org/texstudio/pull/1200#discussion_r467325762. That is, move all the plaintex-only macros from `tex.cwl` to `plaintex.cwl`.

 - The 1st and 3rd commits only moved lines between files, hence the total lines of insertions and deletions are equal.
 - The 2nd commit did some misc changes.
    - `\textindent` was recorded in `tex.cwl` twice prior to this pr, and it was moved to `plaintex.cwl` in 1st commit. Hence the second record can be safely removed.

PS: plain TeX format is defined in file `plain.tex`, which is locally located in path returned by runing `kpsewhich plain.tex` and in [`base` package](https://ctan.org/pkg/base) of CTAN.

### DIFF
--- a/completion/plaintex.cwl
+++ b/completion/plaintex.cwl
@@ -1,4 +1,52 @@
-\ttfam#*
-\bffam#*
-\itfam#*m
-\slfam#*
+\advancepageno
+\beginsection
+\bffam
+\bye
+\cleartabs
+\endinsert
+\eqalignno{%<line%> \cr %<line%> \cr}#m
+\eqalign{%<line%> \cr %<line%> \cr}#m
+\fivebf
+\fivei
+\fiverm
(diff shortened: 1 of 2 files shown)


muzimuzhi:
(Rebased by force pushing.)

sunderme:
thanks

lvjr:
I think the first comment line in `tex.cwl` need to be adapted to the change:
```
# tex/latex mode: tex primitives and plain tex macros
```

muzimuzhi:
@lvjr LaTeX and plain TeX share some macros, so the comment still partially holds.

PR #1204, opened 2020-08-10
[cwl] add alias for class "scrreport", update tcolorbox.cwl

muzimuzhi:
1. add alias for `scrreport`
According to [README of koma-script](https://ctan.org/tex-archive/macros/latex/contrib/koma-script) and from the source file `scrreport.cls`, `scrreport` is a very simple wrapper class to `scrreprt` (note there is no character `o` in `scrreprt`).
1. update `tcolorbox.cwl`
Motivation: I find in usage `\newtcolorbox{mybox}[1]{options}`, `[1]` is not recognized. Hence I started to updated corresponding cwl file, and this leads to the following notable changes:
    1. Record every variant of commands like
     `\newtcolorbox[init options]{envname}[args][default]{options%keyvals}`
    2. Comment option list accepted by arg "init options". When texstudio supports multiple keyvals fields in one command in the future, uncomment it.
    3. (following ii.) Remove suffix "%keyvals" from "init options%keyvals".
    4. Replace `\name` with `arg`.

### DIFF
--- a/completion/tcolorbox.cwl
+++ b/completion/tcolorbox.cwl
@@ -128,28 +128,12 @@ all
 \tcbsetforeverylayer{%<options%>}
 \tcbox[%<options%>]{%<box content%>}
 \newtcolorbox{envname}{options%keyvals}#N
-\newtcolorbox{envname}[args]{options%keyvals}#N
-\newtcolorbox{envname}[args][default]{options%keyvals}#N
-\newtcolorbox[init options]{envname}{options%keyvals}#N
-\newtcolorbox[init options]{envname}[args]{options%keyvals}#N
 \newtcolorbox[init options]{envname}[args][default]{options%keyvals}#N
 \renewtcolorbox{envname}{options%keyvals}
-\renewtcolorbox{envname}[args]{options%keyvals}
-\renewtcolorbox{envname}[args][default]{options%keyvals}
-\renewtcolorbox[init options]{envname}{options%keyvals}
-\renewtcolorbox[init options]{envname}[args]{options%keyvals}
 \renewtcolorbox[init options]{envname}[args][default]{options%keyvals}
 \newtcbox{cmd}{options%keyvals}
-\newtcbox{cmd}[args]{options%keyvals}
-\newtcbox{cmd}[args][default]{options%keyvals}
-\newtcbox[init options]{cmd}{options%keyvals}
-\newtcbox[init options]{cmd}[args]{options%keyvals}
 \newtcbox[init options]{cmd}[args][default]{options%keyvals}
 \renewtcbox{cmd}{options%keyvals}
-\renewtcbox{cmd}[args]{options%keyvals}
-\renewtcbox{cmd}[args][default]{options%keyvals}
-\renewtcbox[init options]{cmd}{options%keyvals}
-\renewtcbox[init options]{cmd}[args]{options%keyvals}
 \renewtcbox[init options]{cmd}[args][default]{options%keyvals}
 \tcolorboxenvironment{envname}{options%keyvals}#N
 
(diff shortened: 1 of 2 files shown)


completion/tcolorbox.cwl
```diff
@@ -321,81 +357,89 @@ all
...
-\ProvideTCBox[init options%keyvals]{\name}{specification}{options%keyvals}#*
-\DeclareTotalTCBox{\name}{specification}{options%keyvals}{content}
-\DeclareTotalTCBox[init options%keyvals]{\name}{specification}{options%keyvals}{content}#*
-\NewTotalTCBox{\name}{specification}{options%keyvals}{content}
-\NewTotalTCBox[init options%keyvals]{\name}{specification}{options%keyvals}{content}#*
-\RenewTotalTCBox{\name}{specification}{options%keyvals}{content}
-\RenewTotalTCBox[init options%keyvals]{\name}{specification}{options%keyvals}{content}#*
-\ProvideTotalTCBox{\name}{specification}{options%keyvals}{content}
-\ProvideTotalTCBox[init options%keyvals]{\name}{specification}{options%keyvals}{content}#*
+\ProvideTColorBox[init options]{envname}{specification}{options%keyvals}#*N
```
sunderme, line 384: Personally i find adding all variants to the completer exaggerated. It clutters the completion list without any real benefit. None of the commands are used regularly. Completing the general one and adapting it to your needs seems more appropriate (see the reason why now there is a plaintex.cwl ...)

sunderme: Furthermore there has not been any complaint in years that variants are missing ...

muzimuzhi: > Personally i find adding all variants to the completer exaggerated.

Yes. I suggest to only add all variants for `\newtcolorbox` and `\newtcbox`, and removing others. Is this acceptable?

> [...] (see the reason why now there is a plaintex.cwl ...)

That's different. Formats are generally exclusive with each other, and are more basic than packages. It's common that a tex document uses several packages, but one cannot use both formats without a lot of tricks. That's the reason I suggested to move plaintex only macros from auto-loaded `tex.cwl` to manually loaded `plaintex.cwl`. Under ideal conditions, switching between formats should contain both adding and removing operations on completion info, but loading packages almost contains adding operation only.

Given another example, if someday users of ConTeXt format want TeXstudio to add support, there could be a new `context.cwl` which can be manually loaded, instead of adding ConTeXt-only macros to `tex.cwl`.

sunderme: My point is rather clutter which makes it more difficult to select the wanted command, not so much the concrete moving of commands to plaintex.cwl

I would suggest that you really look at it from a practical point of view, so the less variants the better. 
I will accept your suggestion, but maybe you try it out in practice to see how it works ?

muzimuzhi: My motivation is to have texstudio recognizing legal use of commands. Although all variants are added, many of them are suffixed with `#*` hence will show in "all" tag only. And more variants can be marked with `#*`, to leave only the no-optional-arg and full-opt-arg variants in the list. Does this still seems to clutter the completion list, from your point of view?

completion/tcolorbox.cwl
```diff
@@ -128,13 +128,29 @@ all
 \tcbsetforeverylayer{%<options%>}
 \tcbox[%<options%>]{%<box content%>}
 \newtcolorbox{envname}{options%keyvals}#N
-\newtcolorbox[init options%keyvals]{envname}[args][default]{options%keyvals}#*N
+\newtcolorbox{envname}[args]{options%keyvals}#N
+\newtcolorbox{envname}[args][default]{options%keyvals}#N
+\newtcolorbox[init options]{envname}{options%keyvals}#N
+\newtcolorbox[init options]{envname}[args]{options%keyvals}#N
+\newtcolorbox[init options]{envname}[args][default]{options%keyvals}#N
```
muzimuzhi, line 135: For example, mark all but
```tex
\newtcolorbox{envname}{options%keyvals}#N
\newtcolorbox[init options]{envname}[args][default]{options%keyvals}#N
```
as `#*N`. This will show only two lines in completion list for `\newtcolorbox`, and recognize all six variant usages.

completion/tcolorbox.cwl
sunderme: Adding all variants to the cwl does not help the syntax checker. Only the longest variant is internally taken for syntax checking. In case of tcolorbox syntax checking does not work reliably because the commands violate the basic syntax grammar (optional args then mandatory ones) In order to handle the package's grammar, internal storage of commands would need to be changed ...

So, the commands with all arguments need to be there, possibly with #*, and maybe a more usual variant. Putting in all variants is not helpful... (clutter, even with #*)

sunderme: Txs understands optional arguments and how they can be used, though not after mandatory args ...

completion/tcolorbox.cwl
sunderme: correct

completion/tcolorbox.cwl
muzimuzhi: > Txs understands optional arguments and how they can be used, though not after mandatory args ...

Understood.

muzimuzhi:
I tested "optional args after mandatory args", it seems to work. Did I misunderstand something? I haven't checked the corresponding source code.

Having
```
$ cat cwl-test.sty
\usepackage{tcolorbox}
```

```
$ cat cwl-test.cwl
\newtcolorbox{envname}{options%keyvals}#N
\newtcolorbox{envname}[args]{options%keyvals}#N
\newtcolorbox{envname}[args][default]{options%keyvals}#N
\newtcolorbox[init options]{envname}{options%keyvals}#N
\newtcolorbox[init options]{envname}[args]{options%keyvals}#N
\newtcolorbox[init options]{envname}[args][default]{options%keyvals}#N
```
and 
```
$ cat txs-cwl-test.tex
\documentclass{article}
\usepackage{cwl-test}

\newtcolorbox{mybox}[1]{title=#1}
\newtcolorbox{mybox2}[1][default]{title=#1}

\begin{document}
\begin{mybox}{Title}
  content
\end{mybox}

\begin{mybox2}
  content
\end{mybox2}
\end{document}
```
I get the right syntax checking:
image

sunderme:
the arguments are probably just ignored, not 100% sure though

sunderme:
so, how do we proceed ?

muzimuzhi:
I've added a new commit to reduce variants in `tcolorbox.cwl`. Only shortest and longest are kept.

sunderme:
Thanks

PR #1208, opened 2020-08-13
Fix #1207

### LINKED ISSUE #1207: PDF spliter ask for existing destination file
### Environment
- TeXstudio: 2.12.22
- Qt: 5.12.1
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior
When using the split and merge tool, from the PDV viewer > File > Split and merge, I expect creating a new output file and therefore, when using the browse button for the destination file, I do expect being able to input a new file name.

### Actual behavior
The file selector ask for a file to *open* and therefore will only accept an existing PDF file.
The only way to work around this is to change the file name by hand in the "Destination File" input field, after selecting a dummy PDF file.

### How to reproduce
Open the split and merge tool, browse for a destination file, try to input a non-existing filename and validate.

suzizecat:
Using getSaveFilename should allow for a non-existant file

### DIFF
--- a/src/pdfsplittool.cpp
+++ b/src/pdfsplittool.cpp
@@ -40,7 +40,7 @@ PDFSplitMergeTool::PDFSplitMergeTool(QWidget *parent, const QString &infile) :
 
 void PDFSplitMergeTool::outputFileDialog()
 {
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #1211, opened 2020-08-19
workaround for old GNU assemblers on arm 32 and 64 bit

matthewberryman:
Refers to registers by direct names rather than aliases. Built and tested this fix on Ubuntu 16.04 (where the problem was [reported](https://github.com/texstudio-org/texstudio/issues/739#issuecomment-674361389)) and 20.04 (to make sure it didn't break) for 32 and 64 bit arm.


sunderme:
thanks

MeanSquaredError:
@matthewberryman 
Thank you for implementing and testing the fix!

Issue #1169, opened 2020-07-08
User command for knitr does not send filename correctly


vam103:
### Environment
- TeXstudio: 3.00.00 git
- Qt: 5.9.5
- OS: Windows 10 and Linux Mint 19.3
- TeX distribution: texlive
- R distribution: 4.0.2
- knitr version: 1.29/1.3 

### Expected behavior
Setting user command for knitr send file of %.Rnw to RScript and first "knits" document before running through LaTeX build process with a pdf with embedded R code produced.

### Actual behavior
When executing the user command R throws an error of unknown object as the filename has had the quotes removed.

### How to reproduce
Using the latest git version of TeXStudio and R with the knitr package installed
Create a file as
```
\documentclass{article}

\begin{document}
\maketitle
\end{document}
```
Save as testing.Rnw

Set up user command
```
Rscript.exe -e "library(knitr);knit('%.Rnw')" | txs:///quick
```
and run the user command.

Interestingly it does work on TeXStudio 2.12.22

sunderme:
might be a problem with escaping quotation marks inside quotation marks

sunderme:
probably, you need to replace % by ?am or ?m (to avoid the enclosing double quotes)

sunderme:
no feedback

sbtseiji:
I tested the suggestions given above and no success.
To narrow down the cause of the problem, I also tested with the simplest user command:

> Rscript -e 'print("hello")'

and again, I got an error.

> Error in print(hello) : object 'hello' not found Execution halted

The command below runs successfully

> Rscript -e 'print("hello")'

sunderme:
@MeanSquaredError the setProgramAndArguments swallows characters.
example: sed -e 's/ab/"gh"/' text.txt  -> ' and " are removed ...

MeanSquaredError:
OK, let me check that...

sunderme:
qprocess apparently sets quotes around arguments which goes south if there are quotes within
Also ' collides at times ...

MeanSquaredError:
I checked the code and indeed `ExecProgram::setProgramAndArguments()` intentionally removes single and double quotes even when they are enclosed in other quotes. I guess at the time I did not take into account that someone might run external programs like sed that need single and double quotes as parts of their arguments.

I will prepare a fix shortly.

sunderme:
QProcess also manipuilates quotes, so please check that QProcess actually works as intended.
[...]

PR #1231, opened 2020-08-27
Allow quotes in executed program names and arguments

MeanSquaredError:
This PR fixes the issue mentioned in https://github.com/texstudio-org/texstudio/issues/1169#issuecomment-680991788 where it was reported that `ExecProgram::setProgramAndArguments()` swallows single and double quotes in the program name and arguments.

After the fix the following program names and arguments are supported:

1. A string without any whitespace, without any single quotes, without any double-quotes, e.g.
```
/my/program arg1 arg2 arg2
```

2. A single-quoted string that can contain whitespace, single quotes or double quotes. Whitespace and double-quotes are supported without any escaping. Single quotes should be backslash-escaped (i.e. \' )
For example:
```
/my/program 'arg2="name"' 'arg3=\'other\''
```

3. A double-quoted string that can contain whitespace or single quotes. Whitespace and single-quotes don't need any escaping. Double quotes inside double quotes are not supported (even when escaped) for compatibility with previous behavior.
For example:
```
sed -i 's/ab/"gh"/' text.txt
```

I tested the last example with `sed` by replacing the external PDF viewer with it and it worked correctly on the test file `text.txt`

The reason why double quotes cannot be supported when escaped inside double quotes is that inside `BuildManager::parseExtendedCommandLine` we have the following code:

```
QStringList BuildManager::parseExtendedCommandLine(QString str, const QFileInfo &mainFile, const QFileInfo &currentFile, int currentline)
{
	...
	// need to reformat literal quotes before the file insertion logic, because ?a"
	// might be extended to "C:\somepath\" which would then be misinterpreted as an
	// literal quote at the end.
	if (config->getOption("Tools/SupportShellStyleLiteralQuotes", true).toBool()) {
		str = ProcessX::reformatShellLiteralQuotes(str);
	}
	...
}
```
The config option `Tools/SupportShellStyleLiteralQuotes` seems to be always enabled so effectively, before `BuildManager::parseExtendedCommandLine` passes the command-line to `ExecProgram`, it always replaces `\"` (backslash-escaped double-quote) with `"""` (triple double-quote). So `ExecProgram::setProgramAndArguments()` never sees the backslash-escaped double-quote.

I did not want to get rid of the code in `BuildManager::parseExtendedCommandLine` that replaces the backslash-escaped double-quote with triple double-quote because I did not want to make too big changes.
This does not allow us to use double quotes inside double quotes, but it is mostly OK, since we can use double quotes inside single quotes (like in the sed example. If we need both double and single quotes in an argument we can use something like `program arg='value\'value"value'`, that is enclose the argument in single quotes, use double quotes inside and use backslash-escaped single quotes inside.

Any comments or suggestions for this PR are welcome.

### DIFF
--- a/src/execprogram.cpp
+++ b/src/execprogram.cpp
@@ -39,7 +39,13 @@ ExecProgram::ExecProgram(const QString &progName, const QStringList &arguments, 
  * \brief Basic command-line parser
  * \details Basic command-line parser that splits a command-line into program name and arguments list.
  * The parsed program name and arguments are stored in class member variables.
- * TODO: Improve the parser. We need to add proper quotes processing based on the OS type (UNIX/Win).
+ * The program name and each argument can be in one of the folloowing forms:
+ *   A string without any whitespace, single quotes or double quotes, e.g. C:\Users\Steven\MyProgram.exe
+ *   A double-quoted string. May contain whitespace, single quotes or backslash-escaped double quotes. For example "arg1='value1'"
+ *   A single-quoted string. May contain whitespace, double quotes, backslash-escaped single quotes or. For example 'arg1="value1"' or 'arg1=\'value\''
+ * As of now backslash escaped double-quotes are unusable because the option Tools/SupportShellStyleLiteralQuotes causes BuildManager::parseExtendedCommandLine() to
+ * replace \" (backslash-escaped double-quotes) with """ (triple double-quotes)
+ * TODO: Maybe improve the parser. We need to add proper quotes processing based on the OS type (UNIX/Win).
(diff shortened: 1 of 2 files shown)


sunderme:
seems to work
thanks

sunderme:
I have stumbled across another issue:
```
git  -C /tmp/TeXstudio-AGoIyC/ log --pretty='%h %s@@@'  content.txt
```

`--pretty='%h %s@@@'` is separated into two parts and git does not like that ...

MeanSquaredError:
Indeed the current code splits arguments across quote boundaries. I will prepare tomorrow morning a PR that fixes that issue.

Issue #1251, opened 2020-09-05
Closing the internal terminal causes a crash


MeanSquaredError:
@axkibe 

Closing the internal terminal in TXS causes TXS to crash.
How to reproduce:
1. Build TXS with support for the internal terminal.
2. Start TXS.
3. Click on the [Terminal] tab on the bottom panel.
4. Close the terminal in some way, e.g. by pressing `CTRL+D` or by typing `exit<Enter>`
5. TXS will activate the crash handler.

Disabling `CTRL+D` is not a proper fix, since there are various other ways to close the terminal e.g. by typing `exit<Enter>` and there are probably other ways too.

I guess one possible solution would be add some code that will detect when the terminal closes and to restart the terminal in these cases.

axkibe:
I can take a look at it in the upcoming days. As far as I remember it is designed to restart the terminal in case the process ends and I did test exit, Ctrl+D and kill situations, must have been gotten broken somewhere.

MeanSquaredError:
This is the backtrace + frame 1 where the actual crash happens
```
(gdb) bt
#0  0x000055555702a010 in ?? ()
#1  0x00005555558a55ed in TerminalWidget::initQTermWidget (this=0x555556db15e0) at ../src/toolwidgets.cpp:196
#2  0x00007ffff689ffb0 in doActivate<false> (sender=0x55555702a010, signal_index=7, argv=0x7fffffffcc10) at ../../include/QtCore/../../src/corelib/kernel/qobjectdefs_impl.h:395
#3  0x00007ffff689ffb0 in doActivate<false> (sender=0x55555787c3a0, signal_index=4, argv=0x7fffffffccf0, argv@entry=0x0) at ../../include/QtCore/../../src/corelib/kernel/qobjectdefs_impl.h:395
#4  0x00007ffff689a7f8 in QMetaObject::activate (sender=sender@entry=0x55555787c3a0, m=m@entry=0x7ffff7f94600 <Konsole::Session::staticMetaObject>, local_signal_index=local_signal_index@entry=1, argv=argv@entry=0x0)
    at kernel/qobject.cpp:3930
#5  0x00007ffff7f78527 in Konsole::Session::finished (this=this@entry=0x55555787c3a0) at /usr/src/debug/qtermwidget-0.14.1-4.fc32.x86_64/x86_64-redhat-linux-gnu/lib/moc_Session.cpp:433
[...]

PR #1252, opened 2020-09-05
Fix a crash when the internal terminal widget is closed

MeanSquaredError:
This PR fixes a crash that occurs when the internal terminal widget is closed.
It resolves issue https://github.com/texstudio-org/texstudio/issues/1251


sunderme:
thanks

Issue #1241, opened 2020-08-31
[Feature Request] Spell check with Multiple dictionaries


jthois:
It is very common for more than one languages to be present in a document. Therefore, It would be appreciated if TeXstudio included spell checking with multiple dictionaries, so we dont have to switch the default dictionary every time. 

Kind regards

Issue #462, opened 2019-01-06
Feature Request: Full Hunspell integration


jonasstein:
Usually there are 3 Hunspell dictionaries per language required

1. system wide dict
2. user dict
3. technical terms dict

Also the language may switch in the document.
This is not reflected by the current configuration in TeXstudio 2.12.14:

![image](https://user-images.githubusercontent.com/298682/50738065-45987580-11d0-11e9-9e51-0567c53184c5.png)

sunderme:
not so sure about the "usually" part, never seen a program to offer that

timhoffm:
1. @jonasstein is that documented somewhere? E.g. searching for „hunspell technical terms dict“ doesn‘t yield a specific result on multiple dictionaries.

2. Dictionaries are currently per file settings. It‘s not impossible to do otherwise, but that‘s how it‘s implemented right now.

jonasstein:
When I select `Add to dictionary` I would expect the words to be added to the users Hunspell dictionary.
`man hunspell` describes `~/.hunspell_default` as default folder for users dicts.
`hunspell -D` lists the system dictionaries. The folders are typically `/usr/share/myspell/` or `/usr/share/hunspell/`. I think instead of `/usr/share/texstudio` the dictionary folder should be one of these on Linux.
The package manager of your distribution manages also the dicts and keeps them up to date.
I am not sure, if it the same for all distributions. So the packager will have to adjust it.

Linux distributions usually prefer that dicts are shared so most files of 
https://github.com/texstudio-org/texstudio/tree/master/utilities/dictionaries
should not be installed by TXS. 
The stop words are not available, so they should be shipped by TXS.

Regarding the technical terms dict, I think about something like the [med_dict](https://github.com/glutanimate/hunspell-en-med-glut).

PR #1254, opened 2020-09-05
+84 dictionaries for easy use in TexStudio with Licenses

ZurMaD:
Adding more dictionaries
Dictionaries from Hunkspell
Files(261MB) taken from:
https://github.com/wooorm/dictionaries
Licences:
https://github.com/wooorm/dictionaries#dictionaries-1
GPT-2 
GPT-3 
Apache 2 
MIT

Related:
#1241 #462


sunderme:
hmm, as enticing as it is, it increases the size of the binary/installer probably quite significantly
how large are the files ?

sunderme:
klingon ?

ZurMaD:
All files around 261MB in total, the idea is to select the most used, not like Klingon haha.

sunderme:
so, any suggestions on how to proceed ?

ZurMaD:
Uhmm I think, first to select the most spoken languages.
image

https://en.wikipedia.org/wiki/List_of_languages_by_number_of_native_speakers

So If I select, the proposal list will be:

pt_PT: Portuguese
pt_PT2: Portuguese (Portugal)
ru_RU: Russian
tr_TR: Turkish
ko_KO: Korean
German variations: de_DE + de_AT de_CH
Spanish variations: es_ES + es_MX es_CO es_ES es_AR es_PE es_VE es_CL es_EC es_GT ...
Vietnamese: vi_VI

In case of spanish, IEEE is doing a good work trying to move scientists & researchers from MS Word to Latex. And the list of variations (!='es_ES') is sorted by population per country. 

image
https://en.wikipedia.org/wiki/List_of_countries_where_Spanish_is_an_official_language

sunderme:
I think that I would prefer a mechanism to download dictionaries from the internet, internet then being the our github repository.
alternatively the windows installer could be extended to allow customization.
Download size is not that much of an issue as it used to be but still ...
Do you want to look into that ?

sunderme:
so , I pull the request, but then we have to see what to do about the installers

ZurMaD:
I think it's a good idea `a mechanism to download dictionaries from the internet` but how we can deal with it? in which language will be written? or how to implement inside TexStudio or can be executed like an extension

sunderme:
the win installer is now 140 MB large instead of 90 MB. I think that is acceptable.
However the internal language selector needs to get more usable to handle that number of languages ...

Windows installer is nsi (nullsoft installer), otherwise things need to be integrated directly into txs with c++-

MeanSquaredError:
The Nullsoft installer uses their own scripting language. The language itself does not seem to support downloading external files, but it supports plugins and Nullsoft has a plugin that allows downloading external files:
https://nsis.sourceforge.io/Builtin_NSISdl_plug-in

However it is a solution just for Windows. I am not sure how the various Linux distros will handle the issue (short of just bundling everything into one big package).

For example the .rpm based ones, might offer the package for the base program and multiple additional packages for the different languages or groups of languages. E.g.
```
texstudio-3.x.x.rpm
texstudio-3.x.x-English.rpm
texstudio-3.x.x-French.rpm
texstudio-3.x.x-German.rpm
texstudio-3.x.x-Italian.rpm
...
```
That's the approach they usually take with large ecosystems that would be too big to ship as one package (e.g. Python, PHP, etc.)

sunderme:
for linux, i would not bother. If possible, system hunspell should be used and then the dictionaries need to come from the system
at the moment, only osx dmg needs to be adapted (it seems)

PR #1286, opened 2020-09-26
Add support for `interval` package in completion.

ElieGouzien:
Hello,

Here is a proposition of cwl file for the package `interval`.


sunderme:
thanks

PR #1293, opened 2020-10-05
minor cwl changes

muzimuzhi:
Main changes:
 - Use arg-spec `key%plain` for arguments with name `key` that are not label/ref keys.
 - Add missing `\DefineVerbatimEnvironment` for `fancyvrb.cwl`. It was wrongly recorded as `\RecustomVerbatimEnvironment`.


sunderme:
thanks

PR #1299, opened 2020-10-09
Change the default highlighting colors in dark mode

schtandard:
With the current default, the highlighted characters are made pretty undiscernible.
image
It is very hard to see if one typed braces, brackets or parentheses. A dark highlighting background makes this much easier.
image

Since the same highlighting is used for environment `\begin`nings and `\end`ings, the color has to work there, too. These are highlighted in a purple and dark blue shade, which makes dark blue unsuited as a background color. This is why I switched to green.
image


schtandard:
Hmm, don't know where those whitespace changes come from. Feel free to ignore them.

sunderme:
I have looked at the result.
Can't say that I like the green very much. I would probably rather go for something darker , blue maybe ?

schtandard:
That's true, I don't like it very much either. Blue doesn't work though; the `foo` in the screenshot above is not too easy to read as it is. It's basically impossible with a blue background.

This is due to the dark color of the `foo` of course. If you are open for it, I could just go through all the colors and make sure that every foreground color is light enough and every background color is dark enough to work together. That would probably mean changing a lot of colors, though.

(Since posting this, I noticed that there are several other background colors that are very light, so checking and adjusting every color may not be the worst idea, anyway.)

If you are open to this: Is there an existing test file showing all (or most of) the highlighting modes used by TeXstudio? That would be really helpful.

sunderme:
I am open to changes of the dark mode colors. I just put them together quickly following some style from the internet.
I have set it to blue so far, but if you come up with a comprehensive set, you can hand it in.

Unfortunately there is no prepared test to show all the colors ...

schtandard:
I was able to write test files for most of the highlighting colors, except the following:
- `preedit`
- `temporaryCodeCompletion`
- `grammarMistake`
- `grammarMistakeSpecial1` through `grammarMistakeSpecial4`
- `dtx:verbatim`
- `dtx:specialchar`

Could you provide some guidance on what I need to do to see those highlighting modes?

sunderme:
dtx is special syntax highlighting in dtx files (a latex package format which allows writing docu)
grammarMistake is triggered by grammar mistake (word repetition, "taboo" words and possibly grammar faults picked up by language tool)
preedit is present on OSX. It allows to put a accent (i.e. " ` " in the line which is then changed to è )
I have to look what temporaryCodeCompletion is.

schtandard:
> dtx is special syntax highlighting in dtx files (a latex package format which allows writing docu)

I am aware, but I can't seem to guess how to trigger `dtx:verbatim` mode. (I found `dtx:specialchar` in the meantime.) The usual verbatim commands use the default highlighting (which makes sense).

> grammarMistake is triggered by grammar mistake (word repetition, "taboo" words and possibly grammar faults picked up by language tool)

Do you have an example? Word repetition and bad words have their own highlighting modes. Do I need an additional tool to trigger `grammarMistake`, etc.?

> preedit is present on OSX. It allows to put a accent (i.e. " ` " in the line which is then changed to è )

Ok, I'm on Windows. Do I understand this correctly, that you type a `` ` ``, which then appears on the line, and if you type an `e` afterwards, it is converted to `è`? So `preedit` only highlights the `` ` `` in that intermediate state?

> I have to look what temporaryCodeCompletion is.

Ok, thanks.

sunderme:
> Do I understand this correctly, that you type a `, which then appears on the line, and if you type an e afterwards, it is converted to è? So preedit only highlights the ` in that intermediate state?

yes, on OSX it is usually represented as a yellow box with the accent, I have not checked what it now looks like in dark mode.

From the syntax file, it looks like dtx:verbatim is just like latex verbatim (\begin{verbatim}....)
dtx:specialChar regex is `\^\^[JKLM]` , so possibly ^^J ? (I have not checked)

schtandard:
Hmm, after looking at the syntax file (should have done that in the first place), `dtx:verbatim` seems to be started by `%<<[A-Za-z0-9]+` and ended by `%[A-Za-z0-9]+`. Have never seen this syntax and can't find any documentation on it. Can give it a color regardless, of course..

schtandard:
> dtx:specialChar regex is `\^\^[JKLM]` , so possibly ^^J ? (I have not checked)

Right, thanks!

sunderme:
it looks like grammar...special* is probably not used, grammarMistake is triggered by languagetool. It is a java program which can run in parallel to txs and checks text delivered by txs.

schtandard:
Ok, here's a proposal. Somme comments:
- There are formats in `defaultFormatsDark.qxf` that are not displayed in the settings dialog: `txs-test`, `txs-test-summary`, `txs-test-pass`, `txs-test-fail`, `txs-test-warn` and `txs-test-qdebug`. Do those do anything? (I just chose similar colors to the existing ones for now.)
- The `foreground` color for `text` was set to `#ffFFFFF` (which is an invalid color) in the file. Since it seems to get its color from elsewhere, I just removed that line.
- The formats `dtx:guar` and `dtx:macro` are located _inside_ the format `txs-test-qdebug`. Is that on purpose?

I wrote test files that display most of the formats. Do you want me to push them anywhere for later use?

sunderme:
I would assume taht the txs-test is used for self-test output, but I have not checked.
dtx:* should probably not be inside txs-test-qdebug

sunderme:
you can place the test file in utilities

the braceMatch color is too dark ... is barely visible with a dark background (adwaita Dark)

sunderme:
something like #0c58a5 ?

schtandard:
> 
> 
> you can place the test file in utilities
> 
> the braceMatch color is too dark ... is barely visible with a dark background (adwaita Dark)

Really? I thought it was pretty ok.

image

But subtlety is in the eye of the beholder, of course. I wouldn't go further than `#0b4c8d`, tough, I think.

image

sunderme:
the second is a bit better.
If you look at the colors in light mode, brace match color is not subtle ...

sunderme:
by the way, maybe your monitor is set way brighter than mine ... still it should be striking enough

schtandard:
> You want to pick out where the closing bracket is without effort !

Yeah, but once it's so bright that you can't actually read what's highlighted or what's around it, that isn't optimal either. I agree now that this brighter blue does the job better, though.

schtandard:
By the way, where are the red and blue box that mark matching environment names when one is selected set?

image

It isn't a huge issue, but I think a bit lighter colors may be better here.

sunderme:
not sure that the color is taken from the syntax colors ... I will take a look

sunderme:
that color is hard coded, not sure if it is worth the effort to make it changeable ...

schtandard:
Agreed. I'm sure there's more pressing issues you need to tend to.

sunderme:
I assume that it is finished

sunderme:
thanks

schtandard:
It is, thanks.

PR #1316, opened 2020-10-24
new cwl file for tdsfrmath.sty

kitanokuma:
new cwl file for tdsfrmath.sty
[tdsfrmath.cwl.zip](https://github.com/texstudio-org/texstudio/files/5433337/tdsfrmath.cwl.zip)


sunderme:
apparently you have taken a wrong base to generate the request
it needs to be based against current (master,HEAD) development branch

sunderme:
and your cwl is missing completely 
it should be place in ./completion and add to git

sunderme:
I have added the cwl on my side because it looks good enough and mybe @dbitouze wants to take another look

PR #1321, opened 2020-10-29
Add glossaries-extra.cwl

judober:
Using `\newabbreviation` from the `glossaries-extra` package, I did not see the labels appearing in texstudio. Therefore, I added a `glossaries-extra.cwl` file. It has only slight modifications form the autogenerated version in order to provide the label for `\newabbreviation` similar to `newacronym` of `glossaries`.


sunderme:
the file looks like the auto generated cwl

please remove all code which the user will not use/see , basically everything with #S
replace "args" with meaningful denominators
make use of txs features where applicable (usually better classification with #... , see manual and other cwls)

judober:
Alright, I removed everything generated by the autogeneration. I just used glossaries.cwl as model which seems to be an modified version of the autogenerated file (Also with #S lines). However, I actually only needed two lines for my use case.

sunderme:
you need to add `#include:glossaries` as the package loads the glossaries package.
Some more commands would be nice but it is a valid start

judober:
I plan to add further stuff in case I need it.

sunderme:
Thanks

Issue #1341, opened 2020-11-05
[enhancement] Put caption above table


homocomputeris:
### Environment
any

### Actual behavior
_Latex > Environments > \begin{table}_ produces this template:
```tex
\begin{table}
	% cursor is here
	\caption{}
\end{table}

```

### Enhancement
CMOS, Oxford, and other style guides recommend putting the caption above the table.
I suggest saving a few clicks:
```tex
\begin{table}
	\caption{} % put the caption right after begin...
	% ... and put the cursor here
\end{table}

```

dbitouze:
You could rather use the `floatrow` package and its `\floatsetup[table]{capposition=top}` setting:
```latex
\documentclass{article}
\usepackage{lipsum}
\usepackage{graphicx}
\usepackage{floatrow}

\floatsetup[table]{capposition=top}

\begin{document}
\lipsum[1]
\begin{table}[ht]
  \centering
  \begin{tabular}{|c|}
    \hline
    Foo\\\hline
    bar\\\hline
  \end{tabular}
  \caption{Foo bar}
\end{table}
\lipsum[2]
\begin{figure}[ht]
  \centering
  \includegraphics[width=.25\linewidth]{example-image-a}%
  \caption{Bar foo}
\end{figure}
\lipsum[3]
\end{document}
```
![image](https://user-images.githubusercontent.com/1032633/98261508-d420de00-1f84-11eb-8af3-2c7a7b6c5643.png)

homocomputeris:
That makes sense. The package is 11yo, though.

sunderme:
I also would not want to change default set-up for **every** user because of this request.
I would assume that the majority uses the set-up as is and does not complain.
(and the package solution is the superior one as caption position should be part of the document style)

homocomputeris:
Then would be great to able to customize these environments somehow without creating new macros.

>  (and the package solution is the superior one as caption position should be part of the document style)

A lot of things should, but many are configured via low-level commands. In Latex3 maybe...

homocomputeris:
You're right.
Also, the maintained `caption` package can do this too.

muzimuzhi:
@homocomputeris `caption` package _cannot_ do this. From the doc of `caption`, option `position`,

> Please note that `position=top` does _NOT_ mean that the caption is actually placed at the top of the figure or table. Instead the caption is usually placed where you place the `\caption` command.

[...]

PR #1343, opened 2020-11-05
Moved cursor after \caption in \begin{table}, #1341

muzimuzhi:


sunderme:
this needs to be discussed a bit more

muzimuzhi:
I've seen the discussions under #1341, and will close this PR.

Issue #1101, opened 2020-05-21
[feature request]  Footnote highlighting


wujastyk:
### Environment
- TeXstudio: TeXstudio 3.00.00 (git n/a) 
Updated today.

- Qt: Using Qt Version 5.9.5, compiled with Qt 5.9.5 R 
- OS: Linux Mint 19.3 Cinnamon 
- TeX distribution: TeXlive 2020 updated today

### Yearned-for behavior

Like many TeX users, I'm an academic commonly working with heavily-footnoted text.
It would be _wonderful_ if one could have footnotes highlighted, including across line breaks.

Thus: 
![image](https://user-images.githubusercontent.com/762246/82518710-2eb0a580-9add-11ea-957a-dab0b79beddf.png)

(with choice of colour)
### Actual behavior

```
\footnote{A very learned comment on 
   this that and the other.}
```

I've seen discussions at [Stackexchange](https://tex.stackexchange.com/questions/328144/highlight-footnote-in-the-original-tex-file) etc. but nobody has suggested a solution to this particular issue.  I understand that there are powerful syntax-highlighting features in TeXstudio, and with cwl files, but I cannot work out to make this happen.  Nor have I found a solution on the internet, after some hard searching.

Many thanks for a wonderful editor.

sunderme:
you can take a look here: http://texstudio.sourceforge.net/manual/current/usermanual_en.html#LANGUAGEDEF

wujastyk:
It is beyond my capability at present to turn these instructions into what
I need.  I am not a programmer.  I spent two hours yesterday editing a qnfa
file, with no success.

Best,
Dominik
--
Professor Dominik Wujastyk
<https://www.ualberta.ca/arts/about/people-collection/dominik-wujastyk>
,

Singhmar Chair in Classical Indian Society and Polity
,

Department of History and Classics <http://historyandclassics.ualberta.ca/>
,
University of Alberta, Canada
.


South Asia at the U of A:

sas.ualberta.ca

cktai17:
I also really want this feature to be implemented, as it is vital for a good workflow in humanities research. 

In the meantime, I spend hours in the past few months trying to understand qnfa to code my own solution, and I finally got it working. It is quite "hacky" and not as nice-looking as I had hoped, but it does the trick.

What you need to do is copy the `tex.qnfa` file here in github and add it in the language folder of your config directory. Change the second line of the code in the file to:
```
[...]

PR #1352, opened 2020-11-13
Add footnote highlighting to tex.qnfa

wujastyk:
Following the contribution of @cktai17 (https://github.com/texstudio-org/texstudio/issues/1101) This edit adds the possibility of highlighting the text of \footnote{} strings.

Installation

1. Place tex.qnfa in the TeXstudio local configuration area, under subdirectory "../languages".  For Linux, this is $HOME/.config/texstudio/languages 
	 
2. In TeXstudio / Options / Configure / Syntax Highlighting, choose your preferred colours at asymtote:block and asymtote:keyword.

That's it, folks!

### DIFF
--- a/utilities/qxs/tex.qnfa
+++ b/utilities/qxs/tex.qnfa
@@ -136,6 +136,29 @@ environments in smallUsefulFunctions.cpp)
 	-->	
 	</list>
 	
+	<!-- ========================= FOOTNOTES ====================== -->
+	
+	<!-- Footnote highlighting code contributed by @cktai17 (https://github.com/texstudio-org/texstudio/issues/1101) -->
+	<!-- In TeXstudio / Options / Configure / Syntax Highlighting, choose your preferred colours at asymtote:block and asymtote:keyword-->
+	
+	<context id="footnote" format="asymptote:block" transparency="true">
+    		<start parenthesis="footnote:open" parenthesisWeight="20" format="keyword">\\footnote{</start>
+		<stop parenthesis="footnote:close" parenthesisWeight="20" format="keyword">}</stop>
+		<context id="substring" format="asymptote:block">
+			<start parenthesis="curly2:open">{</start>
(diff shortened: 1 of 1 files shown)


sunderme:
not sure what the intention is here.
The PR changes the default for all users ? Which contradicts the remarks given above ...

wujastyk:
The intention is to add a valuable option to the base set.  It is very hard for most end-users to discover how to do this for themselves, but as many people need this as need highlighted headings etc.

sunderme:
> The intention is to add a valuable option to the base set. It is very hard for most end-users to discover how to do this for themselves, but as many people need this as need highlighted headings etc.
> 
> Which "remarks given above" are you referring to?

„place tex.qfna intexstudio local config area“ ...
Intention is still unclear ( valuable option ...)

Do you want to change highlighting for all users ?

cktai17:
The ideal solution would to be to add the new language under a different name, e.g. "(La)TeX with footnote highlighting". 

I found, however, that changing the name to anything other than "(La)TeX" breaks the highlighting of bibids and references within the document. If this behaviour could be extended to the new renamed language, then end-users can choose whether or not they want to highlight footnotes.

sunderme:
no, there will not be any alternative "Latex" highlighting. 
However the proposed tex.qnfa does not qualify as txs standard (using asymptote:? as color, a limited list of commands which are treated specially instead of a general approach)
I strongly recommend to put this code into the wiki instead. The user instruction on how to install are easy enough ...

wujastyk:
Okay by me.  I did the git pull because I thought this was a syntax-highlighting feature that would be universally desired.  (It was my idea to suggest this for the main branch, not  @cktai17 's.)  Wiki is fine, if you don't want to make this something general for everyone.  But honestly, I do think this is worthy of inclusion for everyone.

sunderme:
it may be, but that needs more work

PR #1367, opened 2020-11-26
AppData: Fill in newer release info

picsel2:
This commit adds release tags up to version 3.0.1. It also adds the
corresponding changelogs from 2.2.18 upwards.

The appdata file was missing the release info for some newer releases.
This results for example in wrong displayed version numbers in flatpak
since it just uses the latest release tag.


sunderme:
thanks

PR #1379, opened 2020-12-06
Fix classification for commands in mathtools.cwl

rishi1999:
The classifications for `\DeclarePairedDelimiterX` and `\DeclarePairedDelimiterXPP` have been changed from `md` to `nd`, as they are not meant to be used in math mode; they are used to define a command to subsequently be used in math mode.


sunderme:
thanks

PR #1393, opened 2020-12-14
cwl - improve mathtools.cwl slightly

muzimuzhi:
This PR fixes a small problem reported in [this TeX-SX question](https://tex.stackexchange.com/q/574293/79060), that in
```tex
\usepackage{mathtools}
\DeclarePairedDelimiter{\abc}{\lvert}{\rvert}
````
the second and third argument of `\DeclarePairedDelimiter` is not treated in math mode.

PS: This is not an integral improvement, which means there might be similar problems remained in `mathtools.cwl`.


sunderme:
thanks

Issue #1435, opened 2021-01-18
Crash on Windows


OgreTransporter:
### Environment
- TeXstudio: 57b9fb33ed11b7559ab9c1347e630dc88b958f40
- Qt: 5.15.2
- OS: Windows 10
- TeX distribution: miktex
- Compiler: Microsoft Visual Studio 2019

### Description
When I open a LaTeX document and click on a location in the document TXS crashes. Error message: `Expression: cannot increment end list iterator`. I was able to trace the error via debugger to the following line:
https://github.com/texstudio-org/texstudio/blob/57b9fb33ed11b7559ab9c1347e630dc88b958f40/src/cursorhistory.cpp#L61
During initialization the list `history` is empty (start = end), therefore the iterator cannot be incremented beyond the end. You can solve the problem as follows:
```
	if (history.size() > 0) {
		++currentEntry;
		history.insert(currentEntry, pos);
	}
	else {
		history.insert(history.begin(), pos);
	}
```

PR #1436, opened 2021-01-18
Bugfix for issue #1435

OgreTransporter:
Checks if the history list is empty before incrementing the iterator, otherwise a memory access error occurs.

See issue #1435.


sunderme:
thanks

PR #1437, opened 2021-01-18
[Improvement proposal] table of contents

OgreTransporter:
This PR adds a new QTreeView with the table of contents next to the structure.


OgreTransporter:
Apparently, there is a demand for such a window:
- https://latex.org/forum/viewtopic.php?t=20856
- https://tex.stackexchange.com/questions/272660/how-can-i-improve-navigation-within-texstudio
- https://tex.stackexchange.com/questions/269028/document-table-of-contents-tree-structure-on-the-left

OgreTransporter:
Why I implemented my own fast search? So that closed files are also covered. As long as the main document is known and open, all included files will be searched as well. If TXS offers the possibility to parse closed files, I would prefer this way. There should be the possibility to open the corresponding file from the tree view to work in the chapter. That is the reason for this feature. Otherwise you could also use the existing "structure" view.

1. If a file does not contain headers it will not be listed. I tried this with a TikZ drawing and it worked for me.
2. As I said, if there is a way to have closed files in memory as well, I like to use TXS. I have first created my own model to avoid crashes and to keep the interventions in TXS as small as possible.

The PR is intentionally still marked as draft to discuss the feature and implementation.

sunderme:
document->baseStructure contains already all structure information of one document and txs automatically loads included documents (unless that is turned off)
As it does not make too much sense to have structureView and a separate TOC, the target should be to replace structureview. Hence all includes should show up, one way or other.

OgreTransporter:
I would keep both views. The per-document structure with includes and other details, and a cross-document table of contents with just the chapters for quick navigation.

OgreTransporter:
I now use the internal data. Only the files for which no data is available are searched separately. Closed files are also not searched every time, but only once when they appear for the first time or when the structure is created. If the model of the structure is changed, the model for the TOC is also changed.

src/toctreeview.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+struct QDataComparer
+{
+	bool operator()(const QList<QVariant>& a, const QList<QVariant>& b) const
+	{
+		return a[1].toInt() < b[1].toInt();
+	}
+};
+
+QList<TocItem> parseFile(QString filePath)
```
sunderme, line 61: there should not be a file parser. txs loads included files for complete analysis and that should do it.
The file parser is way less capable (not verbatim/comment detection, no adaptability for packages, etc.etc) and it will give a noticeable difference between document toc and project toc.

src/texstudio.cpp
```diff
@@ -10959,4 +10969,9 @@ void Texstudio::openBugsAndFeatures() {
 	QDesktopServices::openUrl(QUrl("https://github.com/texstudio-org/texstudio/issues/"));
 }
 
+void Texstudio::tocItemDoubleClicked(QString section, QString file, int line)
```
sunderme, line 10972: txs uses single click for jumping, it makes no sense to have different use mechanisms

src/tocitemdata.h
```diff
@@ -0,0 +1,37 @@
...
+	int getLevel();
+
+private:
+	QList<QTocItemData*> m_childItems;
+	QList<QVariant> m_itemData;
+	QTocItemData* m_parentItem;
+
+	QString m_text;
+	QString m_file;
+	int m_line;
```
sunderme, line 33: there is a reason why txs does not rely on line numbers. The information is outdated as soon as the line is shifted.

src/toctreeview.cpp
```diff
@@ -0,0 +1,182 @@
...
+		while (!stream.atEnd())
+			text.append(stream.readLine() + "\n");
+	}
+	file.close();
+	QList<QString> sf; sf << "input"; sf << "section"; sf << "subsection"; sf << "subsubsection";
+	QList<QList<QVariant>> il;
+	for (int i = 0; i < sf.count(); i++) {
+		int pos = text.indexOf("\\" + sf[i]);
+		while (pos >= 0) {
+			QList<QVariant> item;
```
sunderme, line 78: this is very unreadable code
using a variant as a replacement for a struct (text,lineNr) which inhibit static code checks.

src/texstudio.cpp
```diff
@@ -4413,6 +4422,7 @@ void Texstudio::updateStructure(bool initial, LatexDocument *doc, bool hidden)
 	}
 
 	//structureTreeView->reset();
+	tocTreeView->updateTOC();
```
sunderme, line 4425: if I remember correctly, this is called with every change of structure.
It updates **all** documents from start to finish (rereading files). This does not scale well ...

sunderme:
I have implemented a first version in 3.1.0 which takes some inspiration from this code. The code in 3.1.0 needs further extension and tests.

OgreTransporter:
Thank you very much! I'll take a look and be happy to help. Unfortunately, I have not had time in the last few weeks.

Issue #1313, opened 2020-10-21
Add an entry in Help menu for bugs reports/feature requests


dbitouze:
### Environment
- TeXstudio: 3.0.2

It would be nice for the "Help" menu to contain an entry for bugs reports/feature requests (that would be an hyperlink to https://github.com/texstudio-org/texstudio/issues/new). In addition, it might avoid some of the repeatitive messages such as "[please report on github.com/texstudio-org/texstudio](https://sourceforge.net/p/texstudio/bugs/2476/#3b77)".

sunderme:
seems simple enough that someone can provide a patch/pull request

Issue #1109, opened 2020-05-24
Feature Request: Provide path to portable Miktex


stdedos:
### Environment
TeXstudio 2.12.22 (git 2.12.22)
Using Qt Version 5.12.1, compiled with Qt 5.12.1 R
Windows 10
Miktex

_Would you please also add that template in About window, so that it is pre-filled instead?_

## Expected result
Instead of giving me this warning:

![image](https://user-images.githubusercontent.com/133706/82757942-2792db80-9dec-11ea-8a72-d547c2606d0f.png)

Can you add a prompt to give this path (e.g. `E:\PortableApplications\LaTeX Tools\MiKTeXv2\texmfs\install\miktex\bin\x64`, `E:\PortableApplications\LaTeX Tools\MiKTeX\miktex\bin`) to a portable Miktex installation instead, before this fails?

(And then add it also in Advanced Settings > Additional Path configuration)

OgreTransporter:
In PR #1438 I added a command line option `--miktex` which allows you to specify a path to search for MiKTeX. I also use the portable versions of MiKTeX and TXS because I can distribute those to colleagues without admin privileges.

PR #1438, opened 2021-01-18
Small fixes

OgreTransporter:
Update for german bad words file and a bugfix for https://github.com/texstudio-org/texstudio/blob/fc25b3760c8e21e005c679c998f37649ac867160/src/adwaita-qt/style/adwaitahelper.cpp#L1589
When changing the style to adwaita, TXS crashes with the following error message:
```
Run-Time Check Failure #3 - The variable 'adjustment' is being used without being initialized.
```
If `!outline.isValid()` and `renderFrame` then no value is assigned to `adjustment`. That means the value is just the one in memory. This can be e.g. a very large negative value, where the calculations in line 1589 do not work.


sunderme (commented):
I would call this list too generous. Some words are used in typical technical terms.

OgreTransporter:
The list contains many words like e.g. _eigentlich_, _besser_ or _wahrscheinlich_, which I find almost daily in thesis. The words are only marked and not deleted. Besides, the option can be switched off. It works anyway only if you also select a word list. Everybody can decide for himself if he wants to use the word or not. So the effect should be very small.

OgreTransporter:
I have added two more small changes:

1. as requested in #1313 I added an entry to the help menu for bugs and features.
2. it is not a dialog but only a command line option but I added the possibility to specify a MiKTeX directory (#1109).

That's it for now.

sunderme:
I would very much prefer to have separate pull requests for separate issues, so that discussion and acceptance can be split easily.
That said, bugfix for invalid number gets accepted.
bug reports in help is also okay.

Is there any particular reason not to set ```static QString miktexBinPath = "<search>" ``` with the command line option ? It may also be an idea to generalize that option to work with miktex/texlive on linux/OSX as well. A pointer to the tex binaries should work anywhere.

As stated above, the word list is too extensive. If the argument is that the user does not use it or he can ignore it, that basically means they don't need the list (and we are talking about default installation for *everyone*) From a pedagogic point of view, unhelpful remarks which basically always need to be ignored lead to ignoring the markers altogether, making it irrelevant for the time it might be useful. 
The examples for bad words are already in the current word list. It may be helpful to make it more accessible. (Most interesting only German has "bad" words...)

OgreTransporter:
Ok, I created separate pull requests.

PR #1441, opened 2021-01-19
Adds a --miktex command line option

OgreTransporter:
The --miktex option can be used to specify a path to search for the MiKTeX binaries. Helpful for Issue #1109.

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -9,6 +9,7 @@
 #include "userquickdialog.h"
 
 #ifdef Q_OS_WIN32
+#include "configmanager.h"
 #include "windows.h"
 #endif
 
@@ -800,6 +801,7 @@ QString getMiKTeXBinPathInternal()
 		static const QStringList candidates = QStringList() << "C:\\miktex\\miktex\\bin"
 															<< "C:\\tex\\texmf\\miktex\\bin"
 															<< "C:\\miktex\\bin"
+															<< ConfigManager::miktexSearchDir
 															<< QString(qgetenv("LOCALAPPDATA")) + "\\Programs\\MiKTeX 2.9\\miktex\\bin";
 		foreach (const QString &path, candidates)
 			if (QDir(path).exists()) {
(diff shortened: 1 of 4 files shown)


sunderme:
I will take a closer look tomorrow

sunderme:
Is this approach only meant to mitigate #1109 or has it other applications.
#1109 is probably better fixed by adding a textstudio.ini with the appropriate search path set.

OgreTransporter:
This parameter is used when TXS portable is started for the first time. It cannot find MiKTeX portable:

1. there are no entries in the system configuration
2. there is no default path where the program is located

Of course, once TXS is started, it writes the found paths to the TXS configuration.

This is useful when, for example, in a Windows network, you automatically distribute TXS and MiKTeX portable as one program package to all computers. An unattended installation does not work properly, but you can easily have the portable versions copied automatically. Why should one do this? Because users are too lazy to install updates regularly. This way you can easily roll out updates.

I adjust the parameter so that it can also be used to find TeX Live.

sunderme:
I don't see any benefit of this switch.
In case you use the portable version, txs writes config in appDir/config. There you set whatever you want in additional search path.
Alternatively you can start txs from cli with ```PATH=xyz:$PATH;texstudio``` which basically also points to the tex installation.

Having a command line switch demands having a batch to start and you don't get any benefit from it.
I don't see any user making use f this switch ...

OgreTransporter:
I don't want a batch file or a cli call. It's all much too cumbersome. I copy TXS and MiKTeX to the target computers and create a shortcut in the start menu to TXS with the path to MiKTeX. I can set this on Windows via group policy.

Also, if you have copied TXS portable and MiKTeX/TeXLive portable to your computer, you can set the configuration directly on the first startup. Just start TXS with the LaTeX path using the key combination Windows+R - done!

sunderme:
so, there is no application for this PR ?

OgreTransporter:
> so, there is no application for this PR ?

Sorry, but I do not understand your question.

sunderme:
> > so, there is no application for this PR ?
> 
> Sorry, but I do not understand your question.

Will anybody use this command line option ? My guess is, no ...

OgreTransporter:
> Will anybody use this command line option ? My guess is, no ...

I will use it and all who work with portable versions of MiKTeX or TeXlive like in #1109.

sunderme:
I have implemented a more general approach (--texpath)

Issue #465, opened 2019-01-07
Copy file name to clipboard


tmelorc:
It would be useful to be able to copy to clipboard the name of an opened file from the structure tree, maybe using mouse right button.

foreachthing:
Use "show in Explorer", then you can copy the file name.

jonasstein:
![image](https://user-images.githubusercontent.com/298682/51275808-3cc75100-19d3-11e9-8a8b-78c9ac6db135.png)
something like this?

tmelorc:
This is my idea, in addition of seeing the file in folder, simply copy its name.

kbauer:
There is *File → Fixfix → Copy filename to clipboard". The functionality is however not available from the context menu of the structure view.

PR #1445, opened 2021-01-20
Copy file name and file path to clipboard

OgreTransporter:
Adds two entries for copying filename and file path to the context menu of "Structure" as in issue #465.


sunderme:
thanks

Issue #1025, opened 2020-04-04
Update GB speller


marcoagpinto:
Hello!

I don't know which spellers the software uses (I didn't find information about it), but just in case, could you update the British one to my forked version on GitHub?:
https://github.com/marcoagpinto/aoo-mozilla-en-dict

Also, if the US speller is outdated, you can also find the most recent version (Kevin Atkinson) on my GitHub.

Thanks!

Kind regards,
     >Marco A.G.Pinto
       -----------------

PR #1446, opened 2021-01-20
Update some dictionaries from libreoffice

OgreTransporter:
Update some dictionaries from libreoffice (https://cgit.freedesktop.org/libreoffice/dictionaries/), see #1025.


sunderme:
the text encoding seems to have changed from utf8 to ISO8859-1 ... I believe utf8 makes things easier.

OgreTransporter:
> the text encoding seems to have changed from utf8 to ISO8859-1

https://github.com/texstudio-org/texstudio/blob/761564d78d48c0a707bb7b4cd26c6530139675ad/utilities/dictionaries/de_DE.aff#L13

sunderme:
thanks

PR #1448, opened 2021-01-20
Create multiple issue templates to seperate features from bugs

OgreTransporter:
I have created different templates for bugs and new features. This makes it easier to keep the requests apart.

I would then revise #1442 again and create two entries or a submenu with two entries, which then pre-selects the respective template. In addition, it is perhaps possible per query variable already pass the system information.


sunderme:
are the templates automatically offered ?

OgreTransporter:
Yes, a selection will then appear. You can have a look at this e.g. at Tensorflow repository.

sunderme:
thanks

PR #1455, opened 2021-01-27
Add completion files for ctex-kit

zepinglee:
These are completion files for several packages from [ctex-kit](https://github.com/CTeX-org/ctex-kit). They are adapted from [hubutui/texstudio-cwl](https://github.com/hubutui/texstudio-cwl) are merged to texstudio with the author's permission (https://github.com/hubutui/texstudio-cwl/issues/1#issuecomment-718297957).

> 你们谁有兴趣合并就合并呗，我已经很久不用了这个了．目前也没有兴趣和精力去搞这个．

Translation: “You can merge if you like. I've not use this for a long time and I have no interest to deal with it”.


muzimuzhi:
`texstudio-cwl` uses MIT License, is its maintainer aware of that TeXstudio uses GPL-3.0 License?

completion/ctex.cwl
```diff
@@ -0,0 +1,278 @@
...
+
+\ctexset{keyvals}
+#keyvals:\ctexset
+fontset=#adobe,fandol,founder,mac,macnew,macold,ubuntu,windows,none
+#
+# Sec. 5.3
+punct=#quanjiao,banjiao,kaiming,CCT,plain
+space=#true,false,auto
+autoindent=
+linestretch=#*
```
muzimuzhi, line 25:
```suggestion
linestretch=
```
There is no `#*` for keys.

completion/ctex.cwl
```diff
@@ -0,0 +1,278 @@
...
+part=
+chapter=
+section=
+subsection=
+subsubsection=
+paragraph=
+subparagraph=
+#
+# Sec. 7.1
+part/numbering=#true,false
```
muzimuzhi, line 55:
```suggestion
part/numbering#true,false
```
This will allow input `part/numbering`, which makes use of the default value (usually `true`). See https://github.com/texstudio-org/texstudio/pull/687#issuecomment-520205897.

completion/xeCJK.cwl
```diff
@@ -0,0 +1,208 @@
...
+#modified zepinglee 26 Jan 2021
+
+#include:fontspec
+
+## 3.1
+\xeCJKsetup{keyvals}
+#keyvals:\xeCJKsetup
+xeCJKactive=#true,false
+#
+CJKspace=#true,false
```
muzimuzhi, line 12:
```suggestion
CJKspace#true,false
```

completion/xeCJKfntef.cwl
```diff
@@ -0,0 +1,55 @@
+# xeCJKfntef package
+# Darcy Hu <hot123tea123@gmail.com> 2016
+#modified zepinglee 26 Jan 2021
+
+\CJKunderline{内容}
```
muzimuzhi, line 5: Argument names in English?

completion/zhnumber.cwl
```diff
@@ -0,0 +1,380 @@
...
+tue=
+wed=
+thu=
+fri=
+sat=
+sun=
+#endkeyvals
+
+\zhnum[keyvals]{counter}
+#keyvals:\zhnum
```
muzimuzhi, line 78: Use the extended syntax `#keyvals:\cmdA,\cmdB` could save some lines.

completion/ctex.cwl
```diff
@@ -0,0 +1,278 @@
+# ctex package
+# Darcy Hu <hot123tea123@gmail.com> 2016
+#modified zepinglee 26 Jan 2021
+
+#include:zhnumber
```
muzimuzhi, line 5: Currently TeXstudio cannot load cwl by current engine. Since `ctex` loads `xeCJK` when xetex is used and `ctex` + xetex is popularly used, perhaps adding `#include:xeCJK` here is useful.

completion/xeCJKfntef.cwl
zepinglee: The arguments are taken from the original documentation of `xeCJK` in Chinese.

completion/ctex.cwl
zepinglee: The `xeCJk` is not loaded in when compiling with pdflatex or lualatex. I suggest explicit `\usepackage{xeCJK}` if the completion for it is needed.

completion/xeCJKfntef.cwl
muzimuzhi: Arg-names of other commands are already changed, not only by translating (e.g. Chinese to English) but also by renaming. For example, `\ctexset` is documented as `\ctexset{⟨键值列表⟩}` in `texdoc ctex`. But it is added as `\ctexset{keyvals}` in this PR, instead of either `\ctexset{键值列表%keyvals}` or `\ctexset{options%keyvals}`.

completion/ctex.cwl
muzimuzhi: I don't have a strong opinion here. After all, cwl cannot be as accurate as tex.

completion/ctex.cwl
zepinglee: The explicit “`=true/false`” pattern is used in the documentation thus i suppose it should be encouraged.

completion/xeCJKfntef.cwl
zepinglee: I'll fix it to keep it consistent with the documentation.

muzimuzhi: Though I don't think consistency with doc is a must ...

sunderme:
is this PR improved on ?

zepinglee:
> is this PR improved on ?

Yes, it's working in progress.

zepinglee:
I've just update the contents according to @muzimuzhi 's review.

sunderme:
good to be pulled ?

zepinglee:
> good to be pulled ?

Yes, please.

sunderme:
thanks

PR #1456, opened 2021-01-27
Corrected minor typos in section 1.4.1 of user manual

ahrendsen:
Just reading through the manual and came across them, so I thought I'd submit the corrections.


sunderme:
thanks

Issue #1409, opened 2020-12-30
Frequent crashes with `Assertion 'numsize == 1 && n < d' failed` in libpoppler 


astoeckel:
### Environment
- TeXstudio: 3.0.4 (git 3.0.3-5-g1b62ecb3)
- Qt: Using Qt Version 5.15.2, compiled with Qt 5.15.2 R
- OS: Fedora 33
- TeX distribution: texlive

### Expected behavior
TexStudio should not crash.

### Actual behavior
Since ~an update some time during the Fedora 32 cycle~ upgrading to my AMD Threadripper workstation due to WFH  (I think about five to six months ago), TeXstudio started crashing frequently (about every five to fifteen minutes) with SIGABRT due to `Assertion 'numsize == 1 && n < d' failed`. Crashes are intermittent whenever an updated PDF is loaded in the integrated document viewer (e.g., after recompiling the document).

The crash is caused by passing an empty string into `atof` in line 349 of `fofi/FoFiType1.cc` in Poppler (`fontMatrix[j] = atof(p);`).

**This is an upstream bug in Poppler.**
* **Upstream issue:**  
   https://gitlab.freedesktop.org/poppler/poppler/-/issues/1050
* **Minimal test case:**
   https://github.com/astoeckel/texstudio_poppler_crash_bug/blob/main/main_simple.cpp

~I guess that this is most likely an upstream bug, maybe caused by Poppler parsing an incomplete document. While this should be fixed upstream, maybe TeXstudio can work around this by only loading updated files into Poppler once the generated PDF file has been finalised (i.e., only once the build process has finished).~

### How to reproduce
~Unclear.~ Cannot reproduce on my ancient laptop running Fedora 33. Cannot reproduce when running TeXstudio via X11 forwarding over LAN (which is how I predominantly use TeXstudio). In other words, the slower the system, the less likely this bug is being triggered. This suggests that this is a timing related bug. Crashes seem to be more frequent in more complex documents that include at least one matplotlib-exported PDF.

Could be the same bug as reported in TeXstudio SourceForge bug 2387 (https://sourceforge.net/p/texstudio/bugs/2387/)

**Update:**  
See https://github.com/texstudio-org/texstudio/issues/1409#issuecomment-753247183 for a more reliable method for reproducing the bug. Reproducing the bug may require a high core count CPU.

**Update 2:**
See https://github.com/astoeckel/texstudio_poppler_crash_bug/ for test cases.

[...]

PR #1463, opened 2021-02-01
Hotfix: limit PDF render threads to 1. Prevents frequent crashes for some users.

Myridium:
HOTFIX 
Git diff contributed by @astoeckel in #1409 . Thank you for your contribution.

This commit limits the render threads to 1 in PDF generation. This prevents frequent crashing for users with high core counts, where the (undiagnosed; suspected) threading issue appears to be worse. 

For me, `texstudio` was crashing ~30% of the time the PDF was generated. This commit has eliminated the crashing completely. **I do not notice any difference in the render time**, so I wonder whether the multithreaded PDF generation is needed or working anyway. I haven't looked into it. All I know is that a working `texstudio` is better than one that crashes every third compile.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -190,9 +190,6 @@ int main(int argc, char **argv)
 #if QT_VERSION >= QT_VERSION_CHECK(5,6,0)
 	QApplication::setAttribute(Qt::AA_DisableHighDpiScaling);
 #endif
-
-	QThreadPool::globalInstance()->setMaxThreadCount(1);
-
 	// This is a dummy constructor so that the programs loads fast.
 	TexstudioApp a(appId, argc, argv);
 	bool startAlways = false;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -190,6 +190,9 @@ int main(int argc, char **argv)
 #if QT_VERSION >= QT_VERSION_CHECK(5,6,0)
 	QApplication::setAttribute(Qt::AA_DisableHighDpiScaling);
 #endif
+
+	QThreadPool::globalInstance()->setMaxThreadCount(1);
+
 	// This is a dummy constructor so that the programs loads fast.
 	TexstudioApp a(appId, argc, argv);
 	bool startAlways = false;
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -3086,7 +3086,7 @@ retryNow:
         renderManager = nullptr;
 	}
 
-	renderManager = new PDFRenderManager(this);
+	renderManager = new PDFRenderManager(this, 1);
 	renderManager->setCacheSize(globalConfig->cacheSizeMB);
 	renderManager->setLoadStrategy(int(globalConfig->loadStrategy));
 	PDFRenderManager::Error error = PDFRenderManager::NoError;


sunderme:
as the vast majority of users does not see that crash, I don‘t see the point in limiting performance for everyone.
So, a setting in texstudio.ini should be used and communicated. I thought that there already is one, but I am not sure.

Myridium:
@sunderme like I said I notice no performance difference. I am typesetting a 100 page document with several images. I am not sure if the multithreading was even worknig properly to begin with.

Obviously it is a bug and should be fixed. I'm disappointed that you can't see that.

Myridium:
And you should expect this to start affecting more users. I doubt you know how many users it is affecting. Most users who try a program that crashes will move on to another. With the latest generation of AMD processors we can expect this to come up much more frequently.

astoeckel:
Two thoughts. First, I think that only the changes to `src/pdfviewer/PDFDocument.cpp` are required, the patch limiting the global Qt thread pool size has no effect.

Second, I agree with Myridium's assessment that a) the perceived performance impact of single thread PDF rendering is negligible (and I have used TeX Studio with a single rendering thread on large documents for more than a month now) and b) that users having the same issue are unlikely to report this on GitHub; they may just move to another editor. I was also moving to another editor for a while after TeX Studio started crashing frequently, and only came back because TeX Studio crashing every ten minutes was still better than the other editors.

To summarize, we currently have the following situation:

Low thread count CPU -> Intrinsically small improvement in PDF rendering performance due to multi threading
High thread count CPU -> Potentially faster PDF rendering, but frequent crashes due to multi threading not being implemented correctly

I'd propose limiting the thread count to one per default, and telling users about an ini setting to re-enable multithreaded rendering at their own risk.

Myridium:
Thanks @astoeckel, I am currently recompiling texstudio with that extraneous line removed. I'll test it a bit then add the commit.

sunderme:
use the latest (dev) version of txs
There is a texstudio.ini only configuration which can limit the number of pdf threads to one. `Preview/LimitRenderQueues=1`
default is min(number of cores,8)

Myridium:
So rather than fix a known bug, or even hotfix it, you're going to tell users to implement a hotfix themself? What?

sunderme:
You are aware that it crashes in poppler which is (usually) thread-safe and the "fix" is basically a work-around which disables multi-threaded access to said library ?

astoeckel:
First of all, thank you @sunderme  for looking into this!

Poppler is only partially thread safe. For example, there can only be one thread accessing a single Page object at a time, see [1]. My gut feeling is that there is a race condition in the TeXstudio rendering code that, under rare circumstances, let's different threads access the same page at the same time. In other words, I wouldn't blame upstream for this before this issue is not more thoroughly investigated. I posted some code in #1409 that allows you to trigger the bug quite reproducibly, even within a debugging environment.

Limiting the number of TeXstudio render threads to eight per default may be a reasonable compromise, but I still don't feel good about this.

[1] https://gitlab.freedesktop.org/poppler/poppler/-/issues/666#note_76932

sunderme:
each render engine gets its own page, renders it and then deletes it. The code is in use for several years.
I don't see crashes with 16 threads and crashes are usually reported.
So, something is special why it crashes on your side. Unless that is cleared up, there is not much I can do.
I **have** provided your work-around in a more general way, but I don't see the necessity to activate it for everyone when they don't need it.

astoeckel:
Just to be sure, you are also not able to reproduce the bug when you disable the page cache as I described in #1409?

sunderme:
okay, if I disable caching and reduce size , so that several pages are visible, I can reproduce the crash.
It is still a poppler crash, apparently when parsing for font information. The debugger shows that 8 distinct pages are rendered.
Next step would be a test program which renders and crashes, though not sure if it would do so automatically.

astoeckel:
Good to know! By test program, do you mean a headless CLI program that hammers PDFRenderEngine to evoke the crash? If you'd like to, I can take a stab at that this weekend.

sunderme:
yes, that is what I mean.

PR #1498, opened 2021-02-21
Add PDFRenderEngineStressTest

astoeckel:
Implement a stress test for PDFRenderEngine evoking #1409.

*Do not merge!* This PR is not meant to be merged into `master`; but to share tools that facilitate hunting down the root cause of #1409.

See the discussion in #1463 for more detail.

**How to compile and run:**
```sh
# Clone this branch
git clone git@github.com:astoeckel/texstudio texstudio_pdfrenderengine_stress_test -b pdfrenderengine_stress_test
cd texstudio_pdfrenderengine_stress_test

# Download the test document
wget http://compneuro.uwaterloo.ca/files/publications/stoeckel.2020c.pdf

# Compile with unit tests enabled
qmake-qt5 CONFIG+=debug
make

# Run the unit tests
./texstudio --auto-tests --no-session
```

This should output something along the lines of
```
"/proc/139359/root"
Iteration 2334/100000texstudio: strtod_l.c:1497: ____strtod_l_internal: Assertion `numsize == 1 && n < d' failed.
Iteration 2335/100000Aborted (core dumped)
```

### DIFF
--- a/src/tests/pdfrenderengine_stress_t.cpp
+++ b/src/tests/pdfrenderengine_stress_t.cpp
@@ -1,0 +1,53 @@
+/***************************************************************************
+ *   copyright       : (C) 2021 by Andreas Stöckel                         *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef QT_NO_DEBUG
+
+#include <pdfviewer/pdfrendermanager.h>
+
+#include <QThread>
+#include <QtTest/QtTest>
+#include <iostream>
+#include <random>
+
+#include "pdfrenderengine_stress_t.h"
+
+void PDFRenderEngineStressTest::test_multithreading()
+{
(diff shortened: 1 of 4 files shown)


sunderme:
I would prefer a stand-alone program which just contains the render manager. If we want to hand it in to poppler, it should be as simple as possible to understand/debug.

astoeckel:
That's what I originally intended to do, but this wasn't compatible with the way TeXstudio tests are structured. I have no idea where do place a stand-alone integration test in the TeXstudio repository, and how to adapt the build system to compile a separate executable. I was really hoping to find unit and integration tests being compiled into separate executables.

Any suggestions on how to proceed?

sunderme:
the test case is for forwarding it to poppler, a separate project is totally fine. Ideally it just contains a .pro file plus main.cpp plus the rendermanager/engine. Mind you, the poppler designer needs to build it with as little effort as possible.

astoeckel:
Alright, this was actually much easier to do than I feared; here you go:

https://github.com/astoeckel/texstudio_poppler_crash_bug

Note that I'm using `meson` instead of `qmake` as a build system; I'm much more familiar with the former.

I'll also try to strip down the code a little more; most of `PDFRenderManager` can be thrown away.

astoeckel:
Ok, I've stripped the bug down to the bare minimum; for some reasons crashes are far less frequent now (mostly after I've made sure that PDFRenderManager waits for all threads to be finished after it has been destroyed, otherwise my test ran out of file handles).

PR #1518, opened 2021-03-03
CD: install poppler with qt5 on macOS

muzimuzhi:
This PR (re)enables the build of poppler-qt5 on homebrew, macOS. The key to build poppler-qt5, as I guess, is to use cmake args `-DENABLE_QT5=ON -DENABLE_QT6=OFF`, see https://github.com/Homebrew/homebrew-core/commit/eeb333044a1d1522e457ffd328a3f86e7b0e0157.

The `brew extract` approach proposed in current PR makes best use of homebrew, hence I think is better than other two approaches.
 - Compared to self downloading then building poppler in https://github.com/texstudio-org/texstudio/commit/f62e185a8cac5ef269d5c0f7970b31cf785f92e1, it makes use of configs in homebrew formular hence simplifies the `cd.yml`.
 - Compared to adding a static `poppler@version.rb` formular file as tried in https://github.com/texstudio-org/texstudio/discussions/1517#discussioncomment-427050, it is easier to upgrade poppler version once needed.

Related homebrew docs:
 - [`brew extract`](https://docs.brew.sh/Manpage#extract---version---force-formula-tap)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -70,14 +70,14 @@ jobs:
         echo "::set-output name=TXS_VERSION::${TXS_VERSION}"
         echo "::set-output name=GIT_VERSION::${GIT_VERSION}"
         
-    - name: Deploy zip to Bintray
+    - name: Upload zip to GitHub Artifacts
       if: github.event_name == 'push'
       uses: actions/upload-artifact@v2
       with:
         name: texstudio-win-zip
         path: ./package-zip/texstudio-win-${{ steps.package.outputs.VERSION_NAME }}.zip
         
-    - name: Deploy as Artifact
+    - name: Upload to GitHub Artifacts
       if: github.event_name == 'push'
       uses: actions/upload-artifact@v2
       with:
@@ -150,8 +150,8 @@ jobs:
      
 ############################
       
-  macosx-bintray:
-    name: Mac OS X (Bintray)
+  macosx:
+    name: Mac OS X
     runs-on: macos-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


muzimuzhi:
If in the future texstudio depends on `poppler@a.b.c` and still uses qt5, but that version of poppler in homebrew history has been configured to build with qt6, then texstudio has to maintain a modified `poppler@a.b.c.rb`.

muzimuzhi:
Hmm, the whole story seems to be a homebrew issue hence this PR (as well as https://github.com/texstudio-org/texstudio/commit/7ae1313db2b6ee8c136f78ee0c1430fcb535b99c and https://github.com/texstudio-org/texstudio/commit/f62e185a8cac5ef269d5c0f7970b31cf785f92e1) is totally unnecessary/avoidable. From the history of [`poppler.rb`](https://github.com/Homebrew/homebrew-core/commits/1946d1fed04eaeb521a3b227caad5a66872ac19f/Formula/poppler.rb), poppler was firstly changed to depend on qt6, then back to qt5. The previous CD failures happened just between that two changes.

muzimuzhi:
> _Edit:_ I've added [2814aa1](https://github.com/texstudio-org/texstudio/commit/2814aa12014dd66b3f006cc215c407fad9c1a114) to verify the above words.

Simply `brew install poppler` now passes the "Configure" step too, see https://github.com/muzimuzhi/texstudio/runs/2026226249

muzimuzhi:
So to fix the CD for now, a minimal change could be to revert https://github.com/texstudio-org/texstudio/commit/f62e185a8cac5ef269d5c0f7970b31cf785f92e1.

sunderme:
Thanks

PR #1519, opened 2021-03-04
CD: run on pull requests

muzimuzhi:
Run `cd.yml` on both pushes and pull requests, but upload to artifacts only on pushes.

Currently, pull requests won't trigger `cd.yml`, hence will not be tested by building/compiling. If a PR passes CI but breaks CD, then the problem will not be found until the PR is merged. This PR changed this behavior, hence roughly brought the conditions back to the ones before https://github.com/texstudio-org/texstudio/commit/f9c304620ec2e7c7858e7dc0f0d1d1de47f71fae.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,6 +1,7 @@
 name: CD
 on:
   - push
+  - pull_request
 
 jobs:
   build-linux-win:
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #1523, opened 2021-03-07
Extend cwl of siunitx

judober:
I noticed a few abbreviation units from the manual were missing in the cwl of siunitx e.g. `\uJ`.


sunderme:
Thanks

PR #1530, opened 2021-03-11
math highlighting for \begin{eqnarray*}

bogger33:
There's highlighting for \begin{eqnarray}, but not the version with the asterisk, which leaves off the equation numbers. this fixes that.

### DIFF
--- a/completion/latex-document.cwl
+++ b/completion/latex-document.cwl
@@ -29,6 +29,7 @@
 \begin{enumerate}\item
 \begin{equation}#\math
 \begin{eqnarray}#\math,array
+\begin{eqnarray*}#\math,array
 \begin{figure}
 \begin{figure}[placement]
 \begin{figure*}
@@ -91,6 +92,7 @@
 \end{enumerate}
 \end{equation}
 \end{eqnarray}
+\end{eqnarray*}
 \end{figure}
 \end{figure*}
 \end{filecontents}


muzimuzhi:
Both `eqnarray` and `eqnarray*` environments are specially recorded in `latex-l2tabu.cwl`, to suggest to not use them. But unfortunately `eqnarray` is contained in `latex-document.cwl` again, hence is shown in autocompletion pop-up list ...
https://github.com/texstudio-org/texstudio/blob/b3ebc39607d1599dc9842a4a11fcffd6e5978d0d/completion/latex-l2tabu.cwl#L2-L6

bogger33:
what are we supposed to use instead? I've seen align suggested, but it doesn't allow to align the relation operator seperately, so it makes it impossible to properly align e.g. \widehat{=} (at least from what I've tried, but I'm not an expert or anything), that's why I prefer eqnarray

e.g. 
```
\begin{eqnarray*}
	G 
	&=& F \\
	&\sim& P \\
	&\widehat{=}& P 
\end{eqnarray*}
```

will align things properly, whereas this doesn't:

```
\begin{align*}
	G 
	& =  F \\
	& \sim  P \\
	 & \widehat{=}  P 
\end{align*}

if i use two ampersands with align, it'll still not align it properly and cause a huge space
```

sunderme:
align is recommended
grafik

bogger33:
i presume it's also deliberate that \underline isn't recognized by default?

sunderme:
underline is part of  tex.cwl which should be activated by default (option/completion: checked tex.cwl)

bogger33:
I guess it must just be my distro's TexStudio package being out of date. I can also see a few others in there that aren't showing up for me (e.g. \bowtie and \sqcap)

PR #1545, opened 2021-03-17
add cwl for package 'changes'

fkastner:
Hi,
I adapted the autogenerated cwl file for the 'changes' package to recognize author id's and the available keyvals.
I have left the long list that txs originally autogenerated at the end. Should that be removed instead?
Is anything else needed?


completion/changes.cwl
```diff
@@ -0,0 +1,108 @@
...
+\setcommentmarkup{arg1}#S
+\setdeletedmarkup{arg1}#S
+\sethighlightmarkup{arg1}#S
+\setsocextension{arg1}#S
+\setsummarytowidth{arg1}#S
+\setsummarywidth{arg1}#S
+\settruncatewidth{arg1}#S
+\summaryofchangesname#S
+\theauthorcommentcount
+\\added#S
```
sunderme, line 78: I would prefer if the useless stuff below is removed (artifacts from the scan process)

completion/changes.cwl
```diff
@@ -0,0 +1,108 @@
...
+\changesnoloc#S
+\changesnosoc#S
+\changesreplacename#S
+\cleaders#S
+\compactsummaryofchangesname#S
+\csname#S
+\ifIsEmpty{arg1}{arg2}{arg3}#S
+\listofchangesname#S
+\protect#S
+\setaddedmarkup{arg1}#S
```
sunderme, line 65: these commands( \set...)  are mentioned in the manual, at least #S should be removed

fkastner:
I updated the file according to your review.

sunderme:
looks good
thanks

PR #1552, opened 2021-03-23
yathesis' completion file updated

dbitouze:


completion/yathesis.cwl
```diff
@@ -123,13 +125,13 @@ datetime
...
 \laboratory{%<nom du laboratoire%>}{%<adresse du laboratoire%>}#n
 #
-\pres[précision%keyvals]{nom du PRES}#n
-\comue[précision%keyvals]{nom de la ComUE}#n
-\institute[précision%keyvals]{nom de l'institut%text}#n
-\coinstitute[précision%keyvals]{nom de l'institut de cotutelle%text}#n
-\company[précision%keyvals]{nom de l'entreprise}#n
-\doctoralschool[précision%keyvals]{nom de l'école doctorale%text}#n
-\laboratory[précision%keyvals]{nom du laboratoire}{adresse du laboratoire}#n
+\pres[%<précision(s)%>]{%<nom du PRES%>}#n
```
sunderme, line 128: have you noticed that keyval completion does not work if you don't define a keyval part ?

sunderme: not sure why %<...%> is preferred here.

muzimuzhi:
curious about why there were so many merging commits ...

completion/yathesis.cwl
dbitouze: > have you noticed that keyval completion does not work if you don't define a keyval part ?

Am I missing something or there are keyval parts:

https://github.com/texstudio-org/texstudio/blob/59eada4a638b06c4647f16e8cc3c594248abf0b5/completion/yathesis.cwl#L134-L153

sunderme: you need to define which argument in the acrual command contains the keyvals
i.e. \pres[...%keyvals]...

dbitouze: > not sure why %<...%> is preferred here.

Not sure either: probably some reminiscences from the early stage of the `.cwl` syntax. 

But, anyway, I'm lost. With, say:
```
\supervisor[%<options%>]{%<prénom%>}{%<nom%>}#n
\cosupervisor[options%keyvals]{%<prénom%>}{%<nom%>}#n
\comonitor[%<options%>%keyvals]{%<prénom%>}{%<nom%>}#n
\referee[précisions%keyvals]{%<prénom%>}{%<nom%>}#n
\examiner[%<précisions%>%keyvals]{%<prénom%>}{%<nom%>}#n
#keyvals:\supervisor,\cosupervisor,\comonitor,\referee,\examiner
professor
seniorresearcher
#endkeyvals
```
I get:

| Command         | Completion | Placeholder |
|-----------------|------------|-------------|
| `\supervisor`   | ✔          | ✔           |
| `\cosupervisor` | ✔          | ✖           |
| `\comonitor`    | ✖          | ✔           |
| `\referee`      | ✔          | ✖           |
| `\examiner`     | ✖          | ✔           |

that is:

| Syntax                   | Completion | Placeholder |
|--------------------------|------------|-------------|
| `%<options%>`            | ✔          | ✔           |
| `options%keyvals`        | ✔          | ✖           |
| `%<options%>%keyvals`    | ✖          | ✔           |
| `précisions%keyvals`     | ✔          | ✖           |
| `%<précisions%>%keyvals` | ✖          | ✔           |

Hence, in order to take advantage of the placeholders for customized text, the `%<...%>` syntax seems to be necessary but, in such a case, no completion. Isn't it possible to take advantage of both?

dbitouze:
> curious about why there were so many merging commits ...

So am I! It looks like the old commits are not purged, but I don't understand why.

completion/yathesis.cwl
sunderme: %<...%> are explicit placeholder markers but pratically disable syntax description. However <%options%> is explicitely recognized as optional/keyval argument (see manual)
I assume, that you did your test with the second argument still using %<...%>. As seen, that does not work. Use the same style as the first argument. It works well in the current yathesis.cwl.

sunderme: General recommendation: don't use the %<...%> syntax

dbitouze: > %<...%> are explicit placeholder markers but pratically disable syntax description. However <%options%> is explicitly recognized as optional/keyval argument (see manual)

OK.

> The second argument can't use %<...%>. As seen, that does not work as txs assumes that placeholders are explicitly marked and not implicitly derived. Use the same style as the first argument. It works well in the current yathesis.cwl.

Sorry, I don't see what you mean with "first" and "second" argument.

sunderme: ?
Any argument will be treated as placeholder.

dbitouze: > Any argument will be treated as placeholder.

That's not the case with:
```
\supervisor[%<options%>]{prénom}{nom}#n
#
#keyvals:\supervisor
professor
seniorresearcher
#endkeyvals
```
The arguments `prénom` and `nom` are not treated as placeholders.

sunderme: same story: <%options%> switches into placeholder syntax.
Use options%keyvals or keys or keyvals

completion/yathesis.cwl
```diff
@@ -182,8 +250,8 @@ juniorresearcher*
 #endkeyvals
 #
 # Misc
-\ordernumber[numéro d'ordre]#n
-\ordernumber#*n
+\ordernumber[%<numéro d'ordre%>]#n
+\ordernumber#n*
```
sunderme, line 254: I think, it needs to be #*n , in that order

sunderme:
Thanks

completion/yathesis.cwl
dbitouze: > I think, it needs to be #*n , in that order

OK, done.

PR #1556, opened 2021-03-26
acronym autocomplete cwl update

GenosseFlosse:
Last update was more than seven years ago. Current package version features commands that capitalize first letter automatically by using `\Ac` instead of `\ac`.


sunderme:
thanks

PR #1626, opened 2021-05-15
Spyros fronimos patch 1 DataTeX plug-in

SpyrosFronimos:
1. Added PythonTeX to build comands
2. Added a PythonTeX build chain
3. DataTeX plug-in for Texstudio. A set of tools for creating and handling LaTeX files and personal notes databases.
With DataTeX plug-in one can create
- LaTeX files databases.
- Personal notes databases.
- LaTeX exercises ordered by Field,Chapter section etc.
- Solutions of the exercises.
- Insert files to editor.

The inspiration, a large part of the code and all the credits belong to Gwenaël Cléon from the "TeXoMaker" project.[Texomaker](https://github.com/Domlol/texomaker)


sunderme:
not sure what is necessary to build this (pdf,pdfwidget module ??)
from a first glance it looks like data is store in a sql database (additional dependency) and it uses its own pdf viewer ?
For sure there will be some discussion necessary, but to start this, it needs to build.

SpyrosFronimos:
Exactly. Sqlite is use to store the data.
My first attempt was to use the built-in pdf viewer but I couldn't make it work and that's why I try to add a seperate pdf viewer.

sunderme:
well. It does not build.
Until that is resolved, we don't need to go into details.

SpyrosFronimos:
Just found out that I need to update some resources which I forgot to.

SpyrosFronimos:
I have temporarily disabled qtpdf and all its dependencies. I suppose now it should build.

sunderme:
1) I still have to tweak the sources to get a compilation
2) I have no idea what the tool should do

SpyrosFronimos:
I will give you some information that might help.
1. This tool is for creating databases and store exercises by field chapter etc. We can create LaTeX files such as exercises, solutions of these exercises, definitions, theorems, examples and many more. 
2. Having these files stored into databases we can easily insert them in an exam or a book and generally inside an editor.
3. We can also create and store our exams in similar databases and custom them as we like.
4. I work on Linux and before I upload this project I got everything working correctly. But its the first time that I use Github and If you see a lot of mistake its because I'm trying to find out thw workflow. I could really need some help on that.
5.  So far I think that with the latest changes it should build properly

I will also try to make a simple video to show the use. Below there are some pictures that may help.

toolbar

newfile

solve

insert

documents

tables

settings

sunderme:
The GUI is not very intuitive and also not really interesting for most users.
Why not have a separate program and just send/insert text into txs ? (like bib-tools do)

SpyrosFronimos:
Haven't thought of that. As a user I thought that I would like to have the toolbar close to the editor so I could see and add/remove exercises in real time. Of course the GUI needs a lot of improvements and any suggestion is most welcome.

sunderme:
I would really prefer an approach like zotero
The program starts a local web server and data can be exhanged via localhost/...
This way, changes in txs can probably be done with script (see zotero-picker) whereas datatex can be developed completely independently at its own pace. (and it can work with other editors,too).
Main reason, I don't see a necessity for this functionality, most users probably don't have a necessity and I don't want to make this code my code (which is what basically happens, once the pull-request goes through)

SpyrosFronimos:
I see what you mean and I like this approach. 
Before I close this pull-request I would like to say again that besides the datatex tool I also added PythonTeX as a build command in the menu as well as a PythonTeX chain. Personally I find it handy when compiling.
If I remember well these additions are in the buildmanager and configmanager files. Feel free to add these changes in the project I you think they will be usefull.

sunderme:
I think about pythontex but I will not make a new entry.
Right now you can already use custom commands to run whatever command, you need.

SpyrosFronimos:
Ok that is usefull. Thanks for the help and your advices!

PR #1645, opened 2021-06-07
Update siunitx.cwl for siunitx version 3.0

### LINKED ISSUE #1643: Update siunitx.cwl for siunitx version 3
### Environment
- TeXstudio: TeXstudio 3.1.2 (git 3.1.2)
- Qt: Using Qt Version 5.15.2, compiled with Qt 5.15.2 R
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior
Autocompletion recognizes the quantity `\qty{num}{unit}` or `\unit{unit}` macros introduced in siunitx v3 and does not recognize units deprecated from v3 such as `\elementarycharge`. It should recognize but mark as unusual or S old `\SI...` macros, as they remain functional but are not recommended for new use.

### Actual behavior
As v3 of siunitx is very new (2021-05-17 for v3.0.0 -- newer than the latest TeXstudio release), the cwl has not yet been updated and so is missing new additions and removals.

kiriya-aoi:
Closes #1643
Initial support for additions/removals/changes in siunitx version 3.0, first released as `Version 3.0.0 2021-05-17`.
Includes additional modifications from versions prior to 3.0 that were not yet reflected in this completion file.

### DIFF
--- a/completion/siunitx.cwl
+++ b/completion/siunitx.cwl
@@ -1,22 +1,19 @@
 # dbitouze/2015-03-20 for siunitx v2.6e
 # thatlittleboy/2018-06-18
+# kiriya-aoi/2021-06-07 for siunitx v3.0
 
 \A#*
 \ampere
 \amu#*
-\ang[%<options%>]{%<degrees%>}
-\ang[%<options%>]{%<decimal number%>}
-\ang{%<degrees%>}
-\ang{%<decimal number%>}
-\angstrom
+\ang[%<options%>]{%<degrees;minutes;seconds%>}
+\ang[%<options%>]{%<number%>}
+\ang{%<degrees;minutes;seconds%>}
+\ang{%<number%>}
 \arcminute
 \arcsecond
 \as#*
 \astronomicalunit
-\atomicmassunit
 \atto#*
-\bar
-\barn
 \becquerel
 \bel
 \bit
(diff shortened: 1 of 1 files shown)


sunderme:
as version 3.0 is new, I,don‘t think,we can discard the old siunitx already.
So, either add the new commands and keep the old ones or make a new file (siunitx3…) especially on linux, not always the latest packages are used.

kiriya-aoi:
I don't think it's advisable to keep old and new in one completion file, since in this case completions will suggest incompatible things to users using both versions.

I understand the motivation behind having a separate `siunitx3.cwl` file for compatibility, but is there a way in texstudio to ignore the built-in `siunitx.cwl` file in favor of the built-in `siunitx3.cwl` file? Otherwise that solution is equivalent to keeping both versions in one file for version 3 as `\usepackage{siunitx}` will continue to activate `siunitx.cwl` regardless of which version the user actually needs.

sunderme:
the advantage of two files is that is gives room to optimize later and allows a switch-over

kiriya-aoi:
I can see no user experience benefits from the viewpoint of someone migrating their active documents from siunitx v2 to v3, or writing new documents using v3. Of course, one must consider that any alternate solution would negatively impact users still using v2, which just changes one group of negatively impacted people for another.

The only option I can think of that does no harm while providing a tangible benefit is to keep the autocompletion file I actually want to use in `.config/texstudio/completion/user` on all of my writing computers to override the built-in version, so I will do that instead. I will keep tweaking my file as needed if someone else wants to integrate it, or ignore it and make a different one, etc.

sunderme:
I still see benefit in having a siunitx3 file. Then I can try to figure out how to accommodate the different versions/user groups.

PR #1648, opened 2021-06-10
Add Apple M1 in BUILD.sh

BrushXue:
Add Apple M1 compatibility, tested with Qt 5.15.2 (installed via Homebrew).


sunderme:
thanks

PR #1652, opened 2021-06-11
Make `git describe --tages` work again

### LINKED ISSUE #1651: Git SHA missing in "About TeXstudio"
### Environment
- TeXstudio: [artifact](https://github.com/texstudio-org/texstudio/actions/runs/925781164) for commit https://github.com/texstudio-org/texstudio/commit/451ac852cbdc3085ac4114b60660e761bc20864c
- Qt: Using Qt Version 5.15.2, compiled with Qt 5.15.2 R
- OS: macOS 10.11
- TeX distribution: irrelevant

### Expected behavior
The first line in "About TeXstudio" reads 
> TeXstudio 4.0.0 (git \<sth containing commit sha\>)

### Actual behavior
For normal CD builds, the git commit SHA is missing.
image

### How to reproduce
Open a CD build for macOS.

muzimuzhi:
Following the workaround suggested in https://github.com/actions/checkout/issues/290, this PR tries to make the `git describe --tags` work again, hence fixes #1651.

The current workaround will "fetch all history for all branches and tags" (quoting [`actions/checkout/README.md`](https://github.com/actions/checkout)). Fetching only the needed branch is better, and fetching only the required number of commits from the latest to the one committed not-latter than the [most recent tag](https://docs.github.com/en/rest/reference/repos#list-repository-tags), is even better. Since the increased fetching time is relatively small compared to the total time, the optimization is not urgent.

To check if it works from inside the action log, I deliberately added `echo ${git describe --tags}` to the workflow.


sunderme:
thanks

muzimuzhi:
Hope the almost-must-use action `checkout` will overcome this limitation in future versions.

PR #1654, opened 2021-06-11
improve textToLatex()

bogger33:
1. Fix a bug where due to the replacement order, backlashes were not correctly converted (since the curly braces of \textbackslash{} would subsequently also be escaped)
2. add support for greek characters
3. replace quotation mark pairs with `` and '' to allow pasting in quotes without having to manually go hunting for quotation marks


src/smallUsefulFunctions.cpp
```diff
@@ -226,7 +226,7 @@ QString textToLatex(const QString &text)
 	QList<QPair<QString, QString> > replaceList;
 	// replacements for resevered characters according to
 	// http://en.wikibooks.org/wiki/LaTeX/Basics#Reserved_Characters
-	replaceList.append(QPair<QString, QString> ("\\", "\\textbackslash{}"));
```
sunderme, line 229: {} is a valid construct. Not sure why that should be changed.

src/smallUsefulFunctions.cpp
```diff
@@ -236,11 +236,51 @@ QString textToLatex(const QString &text)
 	replaceList.append(QPair<QString, QString> ("~", "\\~{}"));
 	replaceList.append(QPair<QString, QString> ("_", "\\_"));
 	replaceList.append(QPair<QString, QString> ("^", "\\^{}"));
+	replaceList.append(QPair<QString, QString> ("\u0393", "$\\Gamma$"));
```
sunderme, line 239: There are people in the world who use greek letters. This replacements does not make sense for someone who wants to put in a greek filename for example.
Also starting a math env for every letter is not useful.
So, the whole greek letter part is not useful.

src/smallUsefulFunctions.cpp
```diff
@@ -236,11 +236,51 @@ QString textToLatex(const QString &text)
...
+	replaceList.append(QPair<QString, QString> ("\u03D1", "$\\vartheta$"));
+	replaceList.append(QPair<QString, QString> ("\u03D2", "$\\Upsilon$"));
+	replaceList.append(QPair<QString, QString> ("\u03D6", "$\\varpi$"));
+	replaceList.append(QPair<QString, QString> ("\u03F5", "$\\epsilon$"));
 
 	QString result = text;
 	for (QList<QPair<QString, QString> >::const_iterator it = replaceList.begin(); it != replaceList.end(); ++it)
 		result.replace(it->first, it->second);
 
+	result.replace(QRegularExpression("\"(.*?)\""), "``\\1''");
```
sunderme, line 282: This is maybe an acceptable replacement but it is English centric ...

src/smallUsefulFunctions.cpp
bogger33: because it will be turned into \{\} by the two lines below it

src/smallUsefulFunctions.cpp
bogger33: are there countries that use this type of quotation mark but would not want it converted in this way? I know france uses «these», but I don't think they would use " to represent these to begin with, if I'm not mistaken. But maybe you are referring to something else

src/smallUsefulFunctions.cpp
bogger33: I'll remove it

src/smallUsefulFunctions.cpp
sunderme: then it is probably best to move this line further down
The argument for ```\textbackslash{}``` usually is that you can add a space which is then shown or add text, which is placed directly behind.

src/smallUsefulFunctions.cpp
sunderme: There are also German quotes and keyboards down offer anything specific to those.
But I think, that line is acceptable as is and let's see if someone complains.

benibela: It should probably use the quotes from the config https://github.com/texstudio-org/texstudio/blob/adcc398a8bdfdc00d229ec322ceb30830f456a6d/src/configmanager.cpp#L2095

sunderme: This will probably need some work as the replacements are not easily accessible, if I read the code correctly ...

src/smallUsefulFunctions.cpp
bogger33: if you move it further down, the backslashes of escaped curly braces will be turned into `\\textbackslash`, that was the first thing I tried. I considered replacing `\textbackslash\{\}` with `\textbackslash{}` again, but that seemed a bit hacky

sunderme: I see

sunderme: I fixed the issue in the repository

bogger33: I think you accidentally put the line in the wrong place, it now escapes backslashes for the other special characters.

e.g. try to convert `\{}#$%&~_^\\{}\` to latex

sunderme: true, fixed

sunderme:
will there be adaptation, so I can merge the quotes part ?

bogger33:
done! Hope I did that correctly, I've never done this before

sunderme:
thanks

PR #1655, opened 2021-06-12
add cwl file for nicematrix

mbertucci47:
There was no cwl file for the package `nicematrix`, so I created one. I'm not entirely familiar with the process of submitting a cwl file for review so please let me know if I should be submitting this somewhere else.


completion/nicematrix.cwl
```diff
@@ -0,0 +1,229 @@
...
+
+#include:amsmath
+#include:array
+#include:l3keys2e
+#include:pgfcore
+#include:xparse
+
+\NiceMatrixOptions{keyvals}
+
+\begin{NiceTabular}[keyvals]{preamble}[keyvals]#\tabular
```
sunderme, line 12: it seems not reasonable to have arguments ```[keyvals]{preamble}[keyvals]``` for the command completion
I would declare one variant for completer and maybe the complete one for syntax checking (#S)

mbertucci47: I've made the changes, however I was unsure if the `#S` lines should have `\tabular` and `\array` removed from the end, so I left them.

sunderme:
looks good. thanks

PR #1663, opened 2021-06-18
make top, here and force the default positioning

bogger33:
I'm not sure if this is just me, but I almost always want this configuration so my images don't appear in a random location, outside their context. So my proposal is to change the default positioning setting for the insert graphics widget to !th

### DIFF
--- a/src/insertgraphics.cpp
+++ b/src/insertgraphics.cpp
@@ -531,7 +531,7 @@ void InsertGraphicsConfig::readSettings(QSettings &settings)
 	center = settings.value("/center", true).toBool();
 	useFigure = settings.value("/useFigure", true).toBool();
 	captionBelow = settings.value("/captionBelow", true).toBool();
-	placement = settings.value("/placement", "").toString();
+	placement = settings.value("/placement", "!th").toString();
 	spanTwoCols = settings.value("/spanTwoCols", false).toBool();
 	settings.endGroup();
 }


sunderme:
changing the default is not the way to go.
The code looks like it intends to reuse previous settings but fails to really do so.
So, maybe repair the code ?

bogger33:
Good point, that would probably make even more sense. I will try

dbitouze:
I strongly advise against `!h`. The images don't appear in a random location, but following a [specific algorithm](https://tug.org/TUGboat/tb35-3/tb111mitt-float.pdf) which does his best to apply some typographic rules. Typically, with `!h`, you're likely to end up at the bottom of some pages with very high blank vertical spaces, which will ruin the [type color](https://en.wikipedia.org/wiki/Type_color). Hence, I advised to not make `!h` the default and to, possibly, make use of the [fewerfloatpages package](https://ctan.org/pkg/fewerfloatpages).

bogger33:
Usually when I include images they are very context specific, so I don't really have the option to choose their placement. But I think if it just remembers your most recently picked selection it would benefit everyone's use cases

dbitouze:
>Usually when I include images they are very context specific, so I don't really have the option to choose their placement. 

I can understand. You nevertheless can refer to them with `\ref`, `\pageref`, (`varioref`'s, possibly with `cleveref`) `vref`, etc.

>But I think if it just remembers your most recently picked selection it would benefit everyone's use cases

Agree.

sunderme:
save as default does the trick.
grafik

dbitouze:
> save as default does the trick.

Nice feature, but I guess not known/understood: IMO, "as Default" isn't explicit enough. What about something like "Save the current choices for future use" and as a checkbox rather than a button?

sunderme:
I agree, GUI could be better.

PR #1679, opened 2021-06-24
remove #c from \tcbset in tcolorbox.cwl

### LINKED ISSUE #1678: bug in keyval highlighting of \tcbset
The highlighting for keyvals of the `\tcbset` command of `tcolorbox` works as expected for the keyvals defined in lines 547-589 of the [cwl file](https://github.com/texstudio-org/texstudio/blob/master/completion/tcolorbox.cwl), but does not recognize the keyvals defined in lines 525-545 and 594-1513. The obvious difference is that in the latter, the "completion only" classifier `#c` is used, but I don't know what the expected behavior is for commands with multiple specified keyval lists using different classifiers.

### Environment
- TeXstudio: 3.1.2
- Qt: 5.15.2
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior
Recognize the keyvals in lines 525-545 and 594-1513 of the cwl.

### Actual behavior
The key `notitle` is in one of the `#c` keyval lists, while `doc keypath` is in one of the keyval lists without `#c`.
Screenshot 2021-06-24 091242

mbertucci47:
closes #1678


sunderme:
Thanks

PR #1692, opened 2021-07-09
Updated xy.cwl

RamonAbudAlcala:
Updated many common commands from the xy-pic package
Added support for the "two cell feature"


sunderme:
thanks

PR #1698, opened 2021-07-13
Update Info.plist

nivek88:
macOS finder doesn't show the TeXstudio icon for .tikz files unless it is defined here. Finder just shows a generic file icon (blank paper sheet) and that can be confusing. One could mistake .tikz files for .log or other files that show a generic icon and these could be trashed accidentally if you clean up TeX folder manually.


sunderme:
thanks

PR #1702, opened 2021-07-23
new and updated cwl files

mbertucci47:
Updated `mnsymbol` and made name lowercase; updated `microtype`, `todonotes`, and `unicode-math`; added `comicneue`, `dejavu-otf`, `domitian`, `fdsymbol`, `firamath-otf`, `fourier-otf`, `knowledge`, `logix`, `macrolist`, `mathcommand`, `pst-xkey`, `raleway`, `sourcecodepro`, `sourcesanspro`, `sourceserifpro`, `xkeyval`, and `xkvview`.


sunderme:
Hmm, why did you change MnSymbol to lower case ? 
I see that the style is called MnSymbol.sty and at least on Linux, you can't use ```\usepackage{mnsymbol}```

mbertucci47:
Interesting, I did not know that package names were case sensitive depending on the operating system. On Windows, `\usepackage{mnsymbol}` works identically to `\usepackage{MnSymbol}`, however with the former, TeXStudio does not recognize `MnSymbol.cwl` which is why I changed it. Do you suggest changing it back to `MnSymbol.cwl`?

sunderme:
yes please
(Windows file system is case insensitive, all others are case-sensitive)

mbertucci47:
Gotcha, should be good now

sunderme:
thanks

PR #1716, opened 2021-08-04
Add cwl for derivative and geometry

blueschu:
Added cwl for the [derivative](https://ctan.org/pkg/derivative?lang=en) package. Updated [geometry](https://ctan.org/pkg/geometry?lang=en) cwl to include keyvals for `\geometry`/`\newgeometry` and a definition class for `\savegeometry`/`\loadgeometry`.


sunderme:
thanks

PR #1717, opened 2021-08-04
Add cwl definition class for LaTeX counters

blueschu:
Added a cwl definition class (`%counter`) for LaTeX counter completion. Counter names will be added to this class when the following commands are used:

- `\newcounter` (from `latex-dev.cwl`)
- `\pgfmathnewcounter` (from `pgfmath.cwl`)

Code completion for counters is added for the following commands:

- `ctex.cwl`: `\chinese`
- `etoolbox.cwl`: `\defcounter`
- `latex-dev.cwl`: `\setcounter`, `\addtocounter`, `\refstepcounter`, `\stepcounter`, `\usecounter`, `\value`
- `latex-document.cwl`: `\Alph`, `\alph`, `\arabic`, `\fnsymbol`, `\Roman`, `\roman`
- `pgfmath.cwl`: `\pgfmathsetcounter`, `\pgfmathaddtocounter`, `\pgfmathmakecounterglobal`

I have also hard-coded completions for the counters defined in the LaTeX base document classes (taken from `classes.dtx` in [latex-base](https://ctan.org/tex-archive/macros/latex/base)). These counters are `part`, `chapter`, `section`, `subsection`, `subsubsection`, `paragraph`, `subparagraph`, `page`, `figure`, `table`, `footnote`, `mpfootnote`, `enumi`, `enumii`, `enumiii`, `enumiv`, and `equation`.

image


sunderme:
I don't see a benefit in offering (and thus cluttering) built-in counters for commands like \stepcounter.
I have not come across a situation where that is wanted. And if it is unlikely, no need to fill the completer with stuff which is not useful for the user.

blueschu:
That's a fair point. I agree that using `\stepcounter` et al. with most built-ins is unlikely. The main plus in having built-in counter completion for commands like `\stepcounter` is that it adds some consistency. This way all locations that expect a counter will show the exact same completions. If we have built-in counters show in some contexts but not others, the interface could feel buggy for the end users. 

There's also some cases where using those commands with built-in counters does come up, in particular with the `\enumX` counters. I've personally used commands like `\stepcounter{enumX}` and `\setcounter{enumX}{x}` when writing reports that need to reference only certain parts of some other document. But I do so I see the benefit in not cluttering the completer for their sake.

Built-in counters are probably most likely to be used with `\chinese`, `\value`, and the `latex-document` commands. If you like I can remove the built-in completions for all but those commands.

sunderme:
how often do you expect \arabic,etc to be used with builtin counters ?
All tutorial basically use it with new counters.
And your image above nicely illustrates the issue. If "MyCounter" was not at first position but last, it would necessitate scrolling/adding letters to select whereas in the current set-up it would be the only (sensible) choice.

blueschu:
I could be wrong, but my guess would be that it's more common for users to want to reconfigure existing structures than to create their own. And with that, it's probably more common that users will want do things like changing the section, footnote, or list numbering with commands like `\renewcommand*\thesection{\Roman{section}}` than creating their own systems of counters. A quick search of [Github repos that use \arabic](https://github.com/search?p=2&q=language%3ALaTeX+%5Carabic&type=Code) seems to support this. In the first few pages of results, I see a lot more uses with built-ins than with non-built-ins.


Regarding the position of custom counters vs built-in counters, I'm not sure I see the issue. It already takes at least one keypress to bring-up the completion dialogue, and at that point the number of completions in the dialogue can be severely limited. In the worst-case scenarios, when the user defines a new counter that begins with `su`, `pa`, or `en`, it will take at most 3 or 4 key presses to navigate to it. Any other prefix would take between 1 and 2 key presses to narrow the search. That doesn't seem all that severe to me, especially considering that getting completion for a command like `\section` already takes at least 4 key presses, regardless of what combination of typing prefixes vs navigating the dialogue you use.

I think built-in completions would be a net plus, but I'm happy to drop them if you're hesitant.

muzimuzhi:
Supporting only built-in counters is limited and storing the counter list several times in cwl files is not ideal. Perhaps texstudio need to support new special lists to be defined in cwl, which seems to be restricted to `%color` currently.

blueschu:
Agreed in full @muzimuzhi. Repeating the built-in counters in every cwl file is definitely a messy hack. Looking now at the [cwl parsing code](https://github.com/texstudio-org/texstudio/blob/74caae433dd00544c43d2c328fb60f3323c73e35/src/latexpackage.cpp#L87), it should actually be pretty trival to modify to cwl format to allow for arbitrary special list definitions going forward. I'm in favor of closing this for now and revisiting later with proper cwl support.

I'll take a stab updating the cwl format in the next few days.

sunderme:
so we close this for now ?

stefgers:
> I don't see a benefit in offering (and thus cluttering) built-in counters for commands like \stepcounter. I have not come across a situation where that is wanted. And if it is unlikely, no need to fill the completer with stuff which is not useful for the user.

maybe I can contribute on that thought from the user perspective, as I'm making heavy use of custom counters based on the `etoolbox` package _(~70 user defined counters to count the use of different environments and pages dependent on some individual section status)_ ... but that is: **_on side of the template_**. I've implemented those in a couple of hours and never touched it again in years of creating content within `\begin{document}` and `\end{document}`. Most of the counters are only used `1-2 times` throught the hole project (only in *.cls, root.tex and some preamble inputs).

It think we can assume that:
- custom/built-in counter manipulation _(as the example above with `\renewcommand` also shows)_ is usually part of  editing the template and preamble or the root document. 
  - `< 1% of usage` (my personal value)
- main use-case for the completer should be "user assisting in writing content"
  -  `> 99%  of usage` (my personal value)

PR #1792, opened 2021-10-03
Disable CRASH_HANDLER to make it compile on RISC-V

alexfanqi:
Backtrace using frame pointer does not work / is not supported for RISC-V.
Recommended approach is using dwarf cfi info or external lib like libunwind.

See also: riscv-software-src/riscv-tools#136


sunderme:
thanks

PR #1811, opened 2021-10-06
Clock font in PDF viewer

tmelorc:
I reduced the font in the clock. 
It was too big so that the descending of the glyph (like in seg (from sec)) was cut off.

Now looks good.
Seleção_169


tmelorc:
I created a new branch (clock) and uploaded the new cpp file. Then I did the PR.
But I still don't know why it says  that I added 7 commits 7 days ago.
I really don't know how to work with Github. Sorry.

PR #1839, opened 2021-10-15
Refactors / Rewrites

IHateYourCode:
Some `refactors / rewrites`, please see the
commits for more detailed information.


sunderme:
As a general note, the PRs should be smaller. One per file makes reviewing much simpler.
About dialog dialo, unix utils and utilsVersion I would accept, but with one big chunk, that is not possible.

PR #1849, opened 2021-10-17
XKeyboard

IHateYourCode:

Moved `CHECK_MSG` from `X11Exception.h` to `XKeyboard.cpp` (as it is the only file that is using it).
Refactored and renamed it to `assert`  and made it local (`#undef`)

Added missing error msg for more detailed debug info.
Msg might have to be adjusted in a future rework of `XKeyboard.cpp`.

### DIFF
--- a/src/xkb/XKeyboard.cpp
+++ b/src/xkb/XKeyboard.cpp
@@ -308,5 +308,5 @@ string_vector parse3(const std::string& symbols, const string_vector& nonsyms)
 
 }
 
-#undef assert
+#undef assertMsg
 


sunderme:
I am not too happy about the macro name "assert" as an official macro of the same name exists (defined a bit differently), see assert.h ...

IHateYourCode:
> I am not too happy about the macro name "assert" as an official macro of the same name exists (defined a bit differently), see assert.h ...

Right, let me change it then.
Any preferences? `check`, `assertTrue`, `assertThat` or something else?

PR #1855, opened 2021-10-17
Macros - HTypes

IHateYourCode:

Expanded `HUNSPELL_THREAD_LOCAL` as it will always be `thread_local`.

Replaced macros `HENTRY_WORD` and `ROTATE` with inline functions.

Replaced macro constants with constants.


sunderme:
thanks

PR #1857, opened 2021-10-17
Macro - LatexParsing

IHateYourCode:

Replaced macros `COMPARE_TOKENTYPE` and `ARG1`
with inline functitons.

`COMPARE_TOKENTYPE` replacement return `bool` and
every use is now an if statement which is actually more 
explicit, considering before the macro basically had a hidden
return statement. 


sunderme:
obsolete

PR #1874, opened 2021-10-20
Template for an article class with French language settings and luatex engine

dbitouze:


sunderme:
any particular reason why the lualatex needs to be set as comment ? (especially as there is already a French article in the templates)

sunderme:
I would assume that you change the default compiler to lualatex anyway ?

sunderme:
Personally I find that template too special.
Since there are not many templates set, I merge.
In case that number of templates is a concern (in a far, little likely future), that one is probably one to ditch.

PR #1905, opened 2021-10-30
Add standalone package cwl

Zheoni:
Hi! This is my attempt at creating a cwl file for the `standalone` package. As it's the first time I do this, I would appreciate if you can check this with extra care before merging something that can be wrong.


sunderme:
looks good
thanks

Issue #1334, opened 2020-11-02
Options/keyvals don't work for some macro in custom CWL


tweh:
### Environment
- TeXstudio: 3.0.1 (git 3.0.1)
- Qt: 5.15.0
- OS: macOS 10.15.7
- TeX distribution: texlive

### Expected behavior
I wrote the following custom CWL code for some macros. The keyvals work as expected for \elementsettings, but they don't for `\element`
```txt
\element{element}{xpos}{node ID}{ID}
\element{element}{xpos}{node ID}{ID}[options%keyvals]
\element*{element}{xpos}{node ID}{ID}
\element*{element}{xpos}{node ID}{ID}[options%keyvals]
\elementsettings{options%keyvals}
#keyvals:\element#c,\element*#c,\elementsettings#c
align=#left,right
type=#pattress,outlet
height=#outlet,switch,controller,sensor,movementsensor,lamp,%<height%>
ductwork=#none,up,down
components=%<components%>
function=%<function%>
name=%<name%>
component-alignment=#above,below,left,right
id-alignment=#above,below,left,right
info-alignment=#above,below,left,right
check-id=#true,false
#endkeyvals
```

### Actual behavior
The keyvals should also work for `\elements`. However I fear this is more a problem of my definition rather than an actual bug …

### How to reproduce
add code to cwl, restart TeXstudio, test

Thanks!

sunderme:
my guess would be that [options%keyval] can't be behind the mandatory arguments for it to work in txs

tweh:
Possible. But in that case I’d consider it to be a bug.

sunderme:
as stated earlier, the issue is the optional command after obligatory commands.
txs does not handle that situation and it is unlikely to change

tweh:
Thanks again. I know my macro syntax is a bit off the standard, but perfectly valid TeX code, so maybe one could take into consideration to make option list work her as well. As far as I can see it should’t be a great difference for TeXstudio to autocomplete such a list in braces instead of brackets at any point in the syntax definition (of course it must be marked with `%keyval` in the CWL file.

sunderme:
won't change

PR #1923, opened 2021-11-06
new and updated completion files

mbertucci47:
New cwls for `algpseudocodex`, `algxpar`, `backref`, `btxdockit`, `changebar`, `changepage`, `class-ltxdockit`, `combine`, `dlfltxbcodetips`, `dlfltxbmarkup`, `dlfltxbmisc`, `expkv-cs`, `expkv-def`, `expkv-opt`, `expkv`, `feyn`, `fp-addons`, `fp-basic`, `fp-eqn`, `fp-eval`, `fp-exp`, `fp-pas`, `fp-random`, `fp-snap`, `fp-trigo`, `fp-upn`, `hypcap`, `imakeidx`, `index`, `ltxdockit`, `notespages`, `ntheorem`, `numerica`, `pseudo`, `rerunfilecheck`, `sidenotes`, `statrep`, `tabto`, and `texdimens`.

Updates for `biblatex`, `csquotes`, and `fp`.


sunderme:
thanks

muzimuzhi:
@mbertucci47 Seems you're adding/reviewing cwls lexicographically. What an enthusiasm!

mbertucci47:
Haha yes, something like that. I suppose there are worse ways to procrastinate than writing cwls

Basmah1961:
Hello,

Thank you for your efforts. In `nicematrix`, could you see why TeXstudio does not support autocompletion when putting the options of the environment after the preamble of the document?

One related topic from TeX StackExchange: [keyvals not recognized from *.cwl file](https://tex.stackexchange.com/questions/600525/keyvals-not-recognized-from-cwl-file)

Regards,
Basmah

muzimuzhi:
@Basmah1961 According to #1334,

> @sunderme: my guess would be that [options%keyval] can't be behind the mandatory arguments for it to work in txs
> [...] and it is unlikely to change

mbertucci47:
There's no problem with `options%keyvals` being the mandatory argument. The problem is that there's not currently support for multiple keyval fields, whether or not they have the same arg specification (also mentioned [here](https://github.com/texstudio-org/texstudio/pull/1930#discussion_r747810837)).

PR #1937, opened 2021-11-12
cwl: Add missing args in `pgfcore.cwl`

muzimuzhi:
This PR adds missing args for pgf point commands in `pgfcore.cwl`.


sunderme:
thanks

PR #1938, opened 2021-11-13
cwl: fix arg-spec for `\AddToHook`

muzimuzhi:
`\AddToHook` has syntax
```tex
\AddToHook{<hook>}[<label>]{<code>}
```
in which the `<label>` is optional and delimited by `[...]`.

Currently `<label>` is recorded as a mandatory arg and this (tiny) PR fixes this.


sunderme:
thanks

PR #1948, opened 2021-11-19
cwl: add missing arg for lrbox env

muzimuzhi:
`lrbox` env requires a mandatory arg representing the box register that will store the env content. 

Check its doc: https://latexref.xyz/lrbox.html.


sunderme:
thanks

PR #1981, opened 2021-11-30
AMS package amsfont loaded by amssymb so the latter is enough (continue)

dbitouze:
Sorry for not having done this earlier.


sunderme:
thanks

PR #1982, opened 2021-11-30
\title{} before \author{} since \maketitle needs the former but not the latter (continue)

dbitouze:
(Sorry for not having done this earlier.) and `babel` should be the last one (though with some exceptions).


sunderme:
thanks

PR #1996, opened 2021-12-06
updated cwls

mbertucci47:
New cwls for `callouts`, `dynamicnumber`, and `titleps`. Updated cwls for `titlesec` and `tikz`, and small fixes elsewhere.


sunderme:
thanks

Issue #1969, opened 2021-11-28
Configure texstudio -> General --> "Check Now" is not working


pagenum:
### Environment
- TeXstudio: 4.1.1
- Qt: 6.2.1
- OS: Windows 10
- TeX distribution: texlive

### Expected behavior
A message box with text 'texstudio is up-to-date' appears
Time of 'Last Checked' changes

### Actual behavior
Nothing happened

### How to reproduce
Press "Check Now" button

sunderme:
I have noticed and I have no idea why.
It is not the code, but the distributed dlls. Probably one is missing, but no idea which.

Lleafll:
windeployqt deploys a tls folder with these dlls:
````
qcertonlybackend.dll
qopensslbackend.dll
qschannelbackend.dll
````
Adding these to the packaged TeXstudio binaries for Windows fixes the tls initialization issues and thus the update check.

benibela:
It probably does not need all of these dlls (schannel is the Microsoft TLS implementation)

Lleafll:
You are right, qopensslbackend.dll is sufficient (at least in this context).

sunderme:
fixed as suggested

amelchiori5:
Hi everybody, I got what the error is on this topic. The only thing is that I'm not able to understand how to fix it on my own!

I donwnloaded the missing _qopensslbackend.dll_, but I do not understand what to do with it.

Help will be much appreciated!

sunderme:
4.1.2 should work.

amelchiori5:
@sunderme I have 4.1.2 installed on my PC and it is still not working atm

sunderme:
apparently it works starting with the dev. build after the 4.1.2 release.
So, you can use one of those (https://github.com/texstudio-org/texstudio/actions/workflows/cd.yml)

pagenum:
> 



> 4.1.2 should work.

I also installed 4.1.2
Still no update check

PR #1997, opened 2021-12-07
Fix missing error popup on update check fail

Lleafll:
Relates to issue #1969. Only fixes missing error popup but does not fix missing DLLs.


sunderme:
fails to build on qt5 ?

Lleafll:
Did not keep in mind TeXstudio targetting older Qt versions. Compiles at least with 5.14.2 now, HEAD does not compile with 5.13 and earlier for me.

sunderme:
implemented by hand to check functionality
Thanks

PR #2000, opened 2021-12-07
Fixes the DisplayName being too long.

ItzLevvie:
Solves complaints in the https://github.com/microsoft/winget-pkgs repository, and complaints from other people about the DisplayName being too long in Control Panel.

Almost all applications have a DisplayName that isn't too long, i.e. `Microsoft OneDrive`; `Microsoft Teams`, etc.

This change was also done to make it a bit more consistent with other programs.


sunderme:
Thanks

PR #2026, opened 2021-12-18
Gather (un)numbered "equations"

dbitouze:
For the Math → Math Equations menu to be consistent with the LaTeX → Sectioning one, the numbered and unnumbered "equations" it contains are not intermixed.


sunderme:
thanks

PR #2028, opened 2021-12-18
cwl updates

mbertucci47:
Updated the `amsxxx` cwls to accurately reflect their dependencies.
New cwls for `amstext`, `braille`, `class-amsart`, `class-amsbook`, `class-amsproc`, `class-pylatex`, `duerer`, `euler`, `pylatex`, `pymacros`, `pythonhighlight`, `texments`, `todo`, and `varsfromjobname`.
Updated cwls for `changes`, `fancyvrb`, `minted`, `spbmark`, `tcolorbox`, and small fixes elsewhere.

PR #2029, opened 2021-12-19
cwl: remove args of two `fancyvrb` macros

muzimuzhi:
In `fancyvrb` package,
- `\FancyVerbGetLine` has its first and the only argument delimited by `^^M` (the carriage return character)
- `\FancyVerbGetVerb` is redefined (in a local group) every time when `\Save(G)Verb{name}<delim char>code<delim char>` is used and has its second argument delimited by the `<delim char>`.

They both cannot be simpliy recorded in the form `\cmd{arg1}{arg2}`. Hence this PR removes their arguments, which are introduced in PR #2028.


sunderme:
thanks

PR #2041, opened 2021-12-30
cwl: fix arg-spec for `\LogHook` and `\RemoveFromHook`, add variants

muzimuzhi:
- Like `\ShowHook{<hook>}`, `\LogHook{<hook>}` accepts a mandatory argument, which is missing in the current cwl file.
 - The `<label>` of `\RemoveFromHook{<hook>}[<label>]` is optional, which is currently denoted as mandatory arg.

This PR fixes these two minor problems and adds two command variants, just for the completeness.


sunderme:
thanks

PR #2095, opened 2022-01-31
cwls for extreport class, etc.

mbertucci47:
New cwls for `class-extbook`, `class-extletter`, `class-extproc`, `class-extreport`, `lwarp`, and `sagetex`. Fixed arg in `chemfig.cwl` to use `%definition` instead of `%formula` to avoid incorrectly highlighting characters like ^ and _ as wrong. Class options added to standard classes.
Beamer subpackages added and class-beamer cwl changed to reflect true file dependencies, which allows them to be `#include:`ed by other packages (like `beamerarticle`).


sunderme:
thanks

PR #2222, opened 2022-04-09
standard ntheorem envs and mathtools' prescript

octaeder:
Added menu Math Definitions containing 14 environments given by \usepackage[standard]{ntheorem}. Added (was deactivated) sideset (from amsmath) and added prescript (from mathtools) in Math Stacking Symbols. Added action selector E for Math Equations, S for Math Stacking Symbols (now Symbols instead of symbols).

@sunderme Please check, thanks

### DIFF
--- a/uiconfig.xml
+++ b/uiconfig.xml
@@ -240,10 +240,10 @@
   <insert id="array" text="\begin{array}" insert="\begin{array}{%&lt;columns%>}%\%&lt;content%|%>%\\end{array}" info="\begin{array}{col1col2...coln}%ncolumn 1 entry &amp; column 2 entry ... &amp; column n entry \\%n...%n\end{array}"/>  
 
   
-  <menu id="equations" text="Math Equations">
+  <menu id="equations" text="Math &amp;Equations">
       <insert id="equation" text="env equation" insert="\begin{equation}\label{%&lt;key%>}%\%|%\\end{equation}" info="The equation environment centres your equation on the page and places the equation number in the right margin." shortcut="Ctrl+Shift+N"/>
       <insert id="equation*" text="env equation* (amsmath)" insert="\begin{equation*}%\%|%\\end{equation*}"/>
-	  <insert id="align" text="env align (amsmath)" insert="\begin{align}%\%|%\\end{align}"/>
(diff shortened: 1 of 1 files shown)


sunderme:
Looks like clutter, especially as many entries seem to be redundant (Example/Beispiel).
I would suggest to use only English names.

octaeder:
For german texts we need the german ones, that's why I included them. I was thinking about a way to present only german ones if german is selected as language in the options and so on. But I don't know how to do so. I have seen that there is alreay "executable" code. Can you tell how to do menu depending on language?

sunderme:
no, the entries language independent (the menu texts can be translated)
Usually the language change is done via tex, see caption.

octaeder:
I see that some words in the menu change on language change. Where can I see which words are those? What is caption? If f. ex. I use env Example the result in the document is Example, it is not translated to Beispiel, what I need. Can I change the code to resolve this? Sorry for all the questions.

sunderme:
I was talking about ```\caption{}```, I have not looked into the tex code. I assume babel is involved.

sunderme:
I would expect, that you want to use env completion for this problem.
E.g ```Beispiel ctrl+alt+space```
I am not convinced that it needs to be in the menu.

octaeder:
so we forgett about german, I will remove those envs. Should I remove the menu "math definitions" at all (as it was bevore) or keep the english ones? And I would keep all other changes, right?

sunderme:
you can keep the english theorems and the rest of the changes

octaeder:
@sunderme How are translations affected by my changes? For ex. here (file texstudio_de.ts and other translation files):

    <message>
        <location filename="../src/additionaltranslations.cpp" line="25"/>
        <source>Math Equations</source>
        <translation>Gleichungen</translation>
    </message>

I changed Equations to \&amp;Equations in uiconfig.xml, and I assume that \<source\> will be updated automatically in the part above. But I have to change Gleichungen in say \&amp;Gleichungen. Can I do this immediately or do I have to wait until update is done?

sunderme:
translations can be done on transifex (see texstudio.org) after I upload the source.

sunderme:
thanks

PR #2250, opened 2022-04-18
disable backtrace on loongarch64 as code does not compile

wming126:
./BUILD.sh need build release version.

Signed-off-by: wangming <wangming@loongson-pc>


sunderme:
thanks

Issue #2258, opened 2022-04-19
Description of TOC


octaeder:
Since usermanual_en.html is created automatically, I can't add a description by myself. I would like do so because of issues #2152 and #2237. I suggest the following and hope that you can do it for me:

Change the content of _2.4 Browsing your document_ as follows:
- use new subsections _2.4.1 The Structure View_ and _2.4.3 Using Bookmarks_ for already present content.
- Add new subsection _2.4.2 The TOC View_ and insert following content (attached two images needed):

2.4.2 The TOC View
The side panel on the left offers a TOC view (if not, use right mouse button and select it). The TOC shows the structure of your whole document by means of section commands (part, section, subsection, …). A mouse over shows you, in which file the section actually is.
 
![grafik](https://user-images.githubusercontent.com/102688820/164060131-65068a86-3174-45fc-9c80-c79aeb20a6c3.png)

The image shows this: A root file named document.tex that includes the file named file1.tex. The root file contains the section „Main text“, while file1.tex contains the subsection „External“. By placing the mouse pointer over this subsection in the TOC view, you find the file name file1.tex in the mouse over. This works even if the file is not open. In this case a mouse click on the subsection in the TOC will open the file for you. Notice the different shades of color that indicate the sections are in different files. It can happen, that an included file doesn’t contain any sections. Then the file name itself appears in the TOC (without file extension):

![grafik](https://user-images.githubusercontent.com/102688820/164060153-075be154-f8db-456f-bfdb-e7329504b421.png)

The mouse over now would show the name of the file that includes file1.tex (namely document.tex). The indentation shows that the text of file1.tex is part of the section “Main text”. You may like to choose a different color in the configuration dialog (s. Use color in global TOC in advanced editor settings).

Attachements:
![TOC_file](https://user-images.githubusercontent.com/102688820/164062218-0daeeb31-5949-4a59-b1ef-984b02967181.png)
![TOC_standard](https://user-images.githubusercontent.com/102688820/164062220-183d1afd-4b01-4b23-9587-7969efff5146.png)

sunderme:
you can edit the usermanual_en.html , it is not derived from anything else.

octaeder:
sorry, I see how I can edit it in github, but how can I download the file for edit it separatly? I can't see it :-(

sunderme:
it is under utilities/manual in the source code.

octaeder:
[...]

Issue #2220, opened 2022-04-07
Enhance creating bookmarks


octaeder:
### Describe the feature and the current behavior/state
Left to the linenumbers of the document is an area where a click creates a bookmark. But with documents with thousands of lines it is not quite clear where this area is. Images show the same document. I tried to depict the area with read lines:

![grafik](https://user-images.githubusercontent.com/102688820/162301052-87520a0c-1574-4706-b99d-ed03f2230636.png)
In the case above about 50% of the area between center bar and line numbers can't be clicked for bookmarks. Moving the pointer as close as possible to the center tool may end up in clicking a tool instead of creating a bookmark. So creating the bookmark is more strenuous.

![grafik](https://user-images.githubusercontent.com/102688820/162300826-c29b00ba-b544-4ce4-9d5f-7460397ba593.png)

Maybe you could make this area more visible in some way, for example with the two lines. Or the columns for bookmarks and line numbers could be interchanged (s. notepad++). Thus both columns would always stay a constant distance from the text area away.

### Who will benefit with this feature?
People using line numbers and bookmarks for large documents.

tmelorc:
Maybe a darker background color on it...

sunderme:
i would rather make the line number part change booksmarks as well.
No real need that the cursor jumps to the line instead.

octaeder:
@sunderme I like your idea not to move the cursor when creating a bookmark, this is sometimes not what I want. Using the numbers at first glance might be good, too. But if I want the cursor at the beginning of a specific line for some purpose it would be easier to click on the number (as it is currently). But I would recommend interchanging both columns as said before. Just let me clarify the reason with two images from documents of different size:

Doc w, 20 Lines:
![grafik](https://user-images.githubusercontent.com/102688820/162494159-457ea354-aa05-4709-b3db-ae98fbe4a34b.png)

Doc with 1000 lines:
![grafik](https://user-images.githubusercontent.com/102688820/162494431-0939e1bb-e49f-4eca-8d01-8d9f800def01.png)

The distance between mark and text is depending of the documentsize (increasing with number of digits of largest line number). I think in this way the eye can better correlate text line and both line number and bookmark.

octaeder:
One note: There might be no line numbers at all or only important ones. And following image the area is quite not clear also:
![grafik](https://user-images.githubusercontent.com/102688820/162996608-1d4618e6-3460-4ea8-b806-7377192f28db.png)

octaeder:
[...]

PR #2261, opened 2022-04-20
usermanual_en.html

octaeder:
- Added section 0. A first look at TeXstudio
- Added section 2.4.2 The TOC View
- Minor text changes


octaeder:
Fixes #2258

utilities/manual/usermanual_en.html
```diff
@@ -392,8 +418,25 @@ <h2 id="SECTION14">2.4 Browsing your document</h2>
 <p>
 The structure view also offers a context menu which allows one to copy/cut all text which belongs to a section (including subsection) and paste it before or after a section.
 Section can be indented/unindented which means that the hierarchy level is changed by one, i.e. \section is changed to \subsection, and all subsections are treated accordingly
+</p>
 
+<h2 id="SECTION142">2.4.2 The TOC View</h2>
+
+<p>The side panel on the left offers a TOC view (if not, s. 0. A first look at TeXstudio). The TOC shows the structure of your whole document by means of section commands (part, section, subsection, …). A mouse over shows you, in which file the section
```
sunderme, line 425: the reference should be a link, so the user can jump to the section easily

sunderme:
looks good. Maybe more links, especially for the one referencing an earlier section.

octaeder:
I had the backref \<a href="#SECTION00"\>0. A first look at TeXstudio\</a\>, was quite easy, cause a simple copy of the contents line. But since it did not work I concluded, that html jumps only forward and removed the (dead) link. Now I see that the target id was #SECTION00 instead of SECTION00.

Btw two notes: 1) My changes affect the description of bookmarks. This description is totally outdated. Funny enough it says "just click on a line number to add or remove a bookmark" (see your suggestion in #2220). 2) When I decided to add the introduction to TeXstudio I realized that all the section numbers are hard coded. Moving parts of the manual to other places or inserting new ones seems to be much work (that's why I used 0 not 1), tasks that could be done quite easily with a LaTeX document. Wouldn't a LaTeX source for the manual be a great example for the work with the superior TeXstudio?

PR #2263, opened 2022-04-20
usermanual_en.html

octaeder:
command from graphics package is named \includegraphics, not \insertgraphics


sunderme:
please use a new branch for PR. Otherwise old commits are also listed which makes reviewing awkward.

sunderme:
thanks

PR #2271, opened 2022-04-24
latex2e.html

octaeder:
Fix typo: \multicolumn instead of multicolumn in worked example.

### DIFF
--- a/utilities/latex2e.html
+++ b/utilities/latex2e.html
@@ -8851,7 +8851,7 @@ range en-dash.
 <div class="example">
 <pre class="example">\begin{tabular}{l|r@{--}l} 
   \multicolumn{1}{c}{\textsc{Period}}  
-    &amp;multicolumn{2}{c}{\textsc{Span}} \\ \hline
+    &amp;\multicolumn{2}{c}{\textsc{Span}} \\ \hline
   Baroque          &amp;1600           &amp;1760         \\
   Classical        &amp;1730           &amp;1820         \\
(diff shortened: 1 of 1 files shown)


dbitouze:
> Fix typo: \multicolumn instead of multicolumn in worked example.

This should be reported [upstream](https://latexref.xyz/dev/) but is already fixed in [the latest version (January 2022)](https://latexref.xyz/dev/latex2e.html#g_t_005cmulticolumn). But it is worth for TeXstudio to provide this latest version. And it is probably worth to provide in the Help menu not only the _LaTeX2e unofficial reference manual_ but also a link to the LaTeX tutorial https://www.learnlatex.org/.

octaeder:
oh, didn't know that this help comes from somewhere else. Thanks for the hint. The link to learnlatex is on the firt line of the help, so let' see what sanderme says.

dbitouze:
> oh, didn't know that this help comes from somewhere else. Thanks for the hint.

You're welcome.

> The link to learnlatex is on the first line of the help, so let' see what sanderme says.

Indeed but most people don't carefully the first lines of such documentation and will likely miss this link. IMHO, it is worth for the Help menu to contain two distinct and clear entries:

- LaTeX Tutorial
- LaTeX Reference Manual

(hence it is clear the second one is not a tutorial).

sunderme:
as said above, we take upstream version to include here. So, we don't patch latex2e.html

PR #2272, opened 2022-04-24
new cwls

mbertucci47:
New cwls for the `biblatex-ext` packages and biblatex styles, `biblatex-multiple-dm`, `biblatex-shortfields`, `parrun`, `postnotes`, `projlib-math`, `table-fct`, `unisc`, the `ctex` classes, and the `easybook` class.
Updated cwls for the `tikzling` packages and other small updates.


sunderme:
thanks

PR #2274, opened 2022-04-25
usermanual_en.html

octaeder:
Since 4.16 "Miscellaneous Features" only describes % begin_fold and % end_fold (in general can be seen as a magic comment) I resolved this section under 4.11 "Advanced header usage", where I changed title to "Advanced header usage with magic comments". The last because following a link from "magic comment" it's easier to recognize the target text.

Btw: 4.11 states that magic comments were introduced in TeXshop, even so the given url contains the word texworks. The web site taks about texworks. Maybe there was a renaming?

### DIFF
--- a/utilities/manual/usermanual_en.html
+++ b/utilities/manual/usermanual_en.html
@@ -73,12 +73,11 @@
 	<li><a href="#SECTION35">4.8 MetaPost support</a></li>
 	<li><a href="#SECTION36">4.9 The "Convert to Html" command</a></li>
 	<li><a href="#SECTION37">4.10 "Forward/Inverse search" with TeXstudio</a></li>
-	<li><a href="#TEXCOM">4.11 Advanced header usage</a></li>
+	<li><a href="#TEXCOM">4.11 Advanced header usage with magic comments</a></li>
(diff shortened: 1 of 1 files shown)


sunderme:
In my understanding, header is the start of the document and magic comments are basically ```% !Tex...```
```%BeginFold...``` can be placed anywhere for a very special purpose. It really has no connection to magic comments (which actually originate from other programs...)

octaeder:
The notation is very similary. My understanding is a slight generalization: a comment that has side effects (besides being a comment) is a magic comment. That's true for all of them. I would suggest changing the header to smoething that matchtes better. Maybe "Using Magic Comments"?

octaeder:
Btw. I have a first pdf version of the manual gererated from LaTeX. Interested?

sunderme:
no. magic comment here is a fixed concept which is also understood by other editors (I think texworks).
The folding part is something else.

octaeder:
So I revert everything. Should I change TeXshop to texworks? What about line 338: <p>One Custom Toolbar is present in TMX. What is ment by TMX?

sunderme:
TeXshop is correct.
TMX is the old abbreviation for texmakerx, the first name of this fork before it was renamed to texstudio.

Issue #2275, opened 2022-04-25
Bookmark 0 in the menus and shortcuts


octaeder:
I have follow findings concerning bookmark shortcuts using current txs 4.2.3 Qt6 on Win 10:

For all bookmarks n from1 to 9 we have go to n with shortcut Ctrl+n, but not for n=0. Please add shortcut. Be aware that
section 4.13 of the manual uses the item name "Set Bookmark" instead of "Toggle Bookmark". That should be adjusted. It states shortcuts for both functions for bookmark 0.

If you like, I can update the manual accordingly, but before let me know what you do.

Thx

sunderme:
ctrl+0 was intentionally left out, why I can't say, but I assume that there is a reason.
So I leave it as is.
You can change the manual if you like.

octaeder:
Cant't figure out how to remove the shortcut shown in the menu Toggle Bookmark.

octaeder:
Btw: U answered my question about ctrl+(+shift+)g. It is used on Mac for find next (previous) and you mentioned that ctrl+l is goto line for Mac. Since the fist 2 are noted in 4.13 Keyboard shortcuts I would conclude that ctrl+l has to be noted there, either. But this is not the case. Whow can I decide what should be there and what not?

octaeder:
Usermanual is updated to reflect that Ctrl+0 is not used for jump to bookmark 0.

octaeder:
@sunderme Note, Ctrl+0 is bound to Reset zoom. Since Zoom in and Zoom out are bound to Ctrl++ and Ctrl+- respectively, I bound Reset zoom to Ctrl+* (for german keyboards either use * on the ten key block or Ctrl with shift and plus). So I can use Ctr+0 for jump to bookmark 0 as usual. Seems quite nice to me.

PR #2283, opened 2022-04-26
Update usermanual_en.html

octaeder:
- removed Ctrl+0 and Ctrl+Shift+0 for Go to / Toggle (not Set) Bookmarks, not working and will not be changed
- fixed typo CTrl
- inconsistent usage of colon in 4.13

### DIFF
--- a/utilities/manual/usermanual_en.html
+++ b/utilities/manual/usermanual_en.html
@@ -1492,7 +1492,7 @@ The following list is a rough overview of the defaults keyboard shortcuts. Depen
 	<li>Go to previous change : Ctrl+H</li>
 	<li>Go to to next change : Ctrl+Shift+H</li>
 	<li>Go to Bookmark 0..9 : Ctrl+1..9</li>		<!-- 04/26/2022 it is assumed that Ctrl+0 would make problems, s. #2275 -->
-	<li>Toggle Bookmark 0..9 : Ctrl+Shift+1..9</li>		<!-- 04/26/2022 it is assumed that Ctrl+Shift+0 would make problems, s. #2275 -->
+	<li>Toggle Bookmark 0..9 : Ctrl+Shift+0..9</li>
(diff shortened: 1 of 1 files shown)


octaeder:
s. further comments/questions here #2275

utilities/manual/usermanual_en.html
```diff
@@ -1487,37 +1487,37 @@ <h2 id="SHORTCUTS">4.13 Keyboard shortcuts</h2>
...
-	<li>Go to to next change: Ctrl+Shift+H</li>
-	<li>Go to Bookmark 0..9: Ctrl+0..9</li>
-	<li>Set Bookmark 0..9: Ctrl+Shift+0..9</li>
-	<li>Set Unnamed Bookmark: Ctrl+Shift+B</li>
-	<li>Next Marker: Ctrl+Down</li>
-	<li>Previous Marker: Ctrl+Up</li>
+	<li>Go to previous change : Ctrl+H</li>
+	<li>Go to to next change : Ctrl+Shift+H</li>
+	<li>Go to Bookmark 0..9 : Ctrl+1..9</li>		<!-- 04/26/2022 it is assumed that Ctrl+0 would make problems, s. #2275 -->
+	<li>Toggle Bookmark 0..9 : Ctrl+Shift+1..9</li>		<!-- 04/26/2022 it is assumed that Ctrl+Shift+0 would make problems, s. #2275 -->
```
sunderme, line 1495: ctrl+shift+0 exists, only ctrl+0 is not used

octaeder: You are right, another fault of the microsoft sandbox: Both shortcuts do not work there (I asssume that the reason is not Version TeXstudio 4.2.4 (git 348e51fe6) I currently use in the sandbox).
I understand that txs doesn't use ctrl+0 for anything. So we could add it, because without jumping to the bm, bm 0 is not so usefull. Or do you want to keep the state as is?

sunderme:
thanks

PR #2286, opened 2022-04-27
Update usermanual_en.html

octaeder:
of course the number in the text has to be changed too, not only the one in the shortcut.


sunderme:
thanks

Issue #2291, opened 2022-04-28
FR: Add LuaLaTex to preview commands


octaeder:
There are already some commands to choose from a list in the Command option in segment Preview. Please add LuaLaTeX as it is already configured in the Commands like PdfLaTeX. txs version 4.2.3 Win10.

Thx

octaeder:
@sunderme Now command for ini mode is extracted with path if necessary (s. buildmanager.cpp), it's no longer hard coded. No test with "no poppler" (how to do this?) and first process has errors. Sources see octaeder/luapreview. How to proceed?

Commands defined:
![grafik](https://user-images.githubusercontent.com/102688820/167314146-deb3a74c-30a2-498a-8201-a5bc7d12a4cb.png)
Preview select lualatex:
![grafik](https://user-images.githubusercontent.com/102688820/167314181-63a61a71-9ac5-4d11-8643-ac5e2bd2ed45.png)
![grafik](https://user-images.githubusercontent.com/102688820/167314173-3476395d-c610-4f72-a822-e100a04fb2de.png)
First process starts luatex (replacement for lualatex), second process starts lualatex command as defined:
![grafik](https://user-images.githubusercontent.com/102688820/167314249-cd25ea97-8098-40e0-842a-ef718b3146fd.png)
But ends with error(s). No clue why.
If command is with path:
![grafik](https://user-images.githubusercontent.com/102688820/167314325-7944355e-076f-400f-8807-711152662615.png)
First process starts luatex correctly with path, second process starts lualatex command as defined:
![grafik](https://user-images.githubusercontent.com/102688820/167314455-9f9b27bf-127f-4a10-b4e3-7483639f63d9.png)
This will work even if file extension .exe is missing.

octaeder:
@sunderme Now command for ini mode is extracted with path if necessary (s. buildmanager.cpp), it's no longer hard coded. No test with "no poppler" (how to do this?) and first process has errors. Sources see octaeder/luapreview. How to proceed?

Commands defined:
![grafik](https://user-images.githubusercontent.com/102688820/167314146-deb3a74c-30a2-498a-8201-a5bc7d12a4cb.png)
Preview select lualatex:
![grafik](https://user-images.githubusercontent.com/102688820/167314181-63a61a71-9ac5-4d11-8643-ac5e2bd2ed45.png)
![grafik](https://user-images.githubusercontent.com/102688820/167314173-3476395d-c610-4f72-a822-e100a04fb2de.png)
[...]

PR #2299, opened 2022-05-01
Luapreview

octaeder:
Add support for pdfpreview with lualatex (s. #2291). Please check carefully as always, cause I'm not sure if I have done all that is necessary.
configdialog.ui should be updated with a new item in the combobox "comboBoxDvi2PngMode". But I assume that I shouldn't edit the file so someone else has to do it with qt-tool. Otherwise I would add:
```
                <item>
                 <property name="text">
                  <string>Preview with lualatex</string>
                 </property>
                </item>
```
Thx

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1866,12 +1866,18 @@ void BuildManager::preview(const QString &preamble, const PreviewSource &source,
     ProcessX *p1 = nullptr;
 	if (dvi2pngMode == DPM_EMBEDDED_PDF) {
 		// start conversion
-		// tex -> dvi
+		// tex -> pdf
 		p1 = firstProcessOfDirectExpansion(CMD_PDFLATEX, QFileInfo(ffn)); //no delete! goes automatically
 	} else {
-		// start conversion
-		// tex -> dvi
-		p1 = firstProcessOfDirectExpansion(CMD_LATEX, QFileInfo(ffn)); //no delete! goes automatically
+		if (dvi2pngMode == DPM_LUA_EMBEDDED_PDF) {
+			// start conversion
+			// tex -> pdf
+			p1 = firstProcessOfDirectExpansion(CMD_LUALATEX, QFileInfo(ffn)); //no delete! goes automatically
+		} else {
+			// start conversion
+			// tex -> dvi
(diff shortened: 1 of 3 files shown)


sunderme:
Have you tested the code ?
From first glance, it looks like you missed some changes: 
buildmanager.cpp line 2074: ```if (dvi2pngMode == DPM_EMBEDDED_PDF) {``` ...

Any particular reason why you can't use qt designer ?

octaeder:
I can search for the missed lines, there are more. How could I do builds locally (in my fork, branch) when working with github? I don't know qt designer, is it easy to work with?

sunderme:
You should build txs with the changes locally, see wiki.
Usually you clone it locally, make the changes in a new branch, push to your fork and from there you make a pull request.
Depending on your system, qtcreator is the IDE to use which also includes the qt-designer which is easy to use.

Untested code is usually considered as faulty code.

octaeder:
@sunderme I'm working hard on this issue. But still with problems. The new drop down may show part of my progress:
grafik
So to finish this I hope you can help me. Be aware that I don't know anything about C/C++ or oop, I'm sorry.
I have problems with the creation of the fmt file (precomile of lualatex.ini and preamble) that should be created by a first process (s. buildmanager.cpp around l. 1810, compare with my branch luapreview). It ends with "Process exited with error(s)". I can't figure out what errors. The second process is started anyway with lualatex and shows the preview. But it doesn't work with larger examples.

As a result of my work some notes: The creation of the first process can't be handled in the same way as for pfdlatex and latex. These two extract the program (possibly incl. path) from the command definitions in the config (technically from CMD_PDFLATEX and CMD_LATEX), to start it with the -ini option (f. ex. pdflatex ... -ini ...). This will create the fmt file. But for lua it has to be "luatex.exe ... -ini ..." while CMD_LUALATEX uses lualatex.exe (if the user didn't enter somthing different). So I hard coded luatex.exe and assume that it will be found in the path of executables. I could do better if someone could provide code that replaces lualatex.exe by luatex.exe in path\lualatex.exe.

sunderme:
the -ini part is for active precompilation of headers, so that could be skipped for later.
luatex should not be hard-coded, better use the lualatex command and change the actual command (but keeping the path)

Issue #2324, opened 2022-05-11
Get pdf x and y position


egidioln:
Hey all, following up on my question in #2307, I implemented the following feature:

Ctrl+shift+LeftMouseClick on the pdfViewer copies to clipboard the mouse coordinates x and y in cm with respect to the bottom left corner of the page. This is useful to add floating elements (especially in beamer presentations) as in the following example:

```
\documentclass{beamer}
\usepackage{tikz}

\tikzstyle{float} = [overlay,remember picture, shift=(current page.south west)]

\begin{document}
		\tikz[float] \draw (0,0)  grid (16,10);
		\tikz[float] \node at (8.6955,4.32964) {\Huge a}; % this coordinate was obtained by ctrl+shift+click on the pdf and ctrl+v here
\end{document}
```
 
It's already working in the commit b7e0e9b80de943bb9389420b6809038efbd96bf8.

Let me know if there is interest in merging this.

sunderme:
fine with me

octaeder:
I observed that the mouse pointer changes to a cross when pressing Ctrl-Alt. But this is true for the magnifier pointer but not for the scroll (hand) pointer. I think it should be the same for the latter one.
Thx

octaeder:
@sunderme @egidioln Did anyone get my last remark?

egidioln:
@octaeder Not really... Here, with the magnifier tool selected, shift turns the pointer into a plus and alt, into a minus, which allows zooming in and out by clicking. Ctrl-alt has the same effect as alt and ctrl+shift, the same as shift... the new feature (ctrl+shift+click to get x and y) still works with the magnifier tool.

What maybe should be done is to deactivate the changing cursor when ctrl-shift is pressed (or even change it into something else, like a crosshair). Is that what you meant?

octaeder:
@egidioln Thanks for clarification, I thought that the plus is the icon for your new feature, so one can better locate the point where you capture the coordinates. Newer seen that plus/minus before, now I see them in the View menu (only windowed pdfviewer). I use double click or scroll wheel with the scroll hand for zooming. And here Alt aswell as Shift doesn't change the hand to something else. And my point was that with Alt+Shift the hand also doesn't change the cursor into the plus (what I assumed to be your crosshair).
[...]

PR #2326, opened 2022-05-12
New Feature: getPosFromClick

egidioln:
ctrl+shift+leftClick on pdf view copies to clipboard the cursor coordinates x and y (in cm, w.r.t. bottom left corner of the page)

A detailed explanation of the motivation is given in #2324 and #2307

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -2754,13 +2754,13 @@ void PDFDocument::getPosFromCLick(const QPointF &pos)
 {
 	int page = pdfWidget->pageFromPos(pos.toPoint());
 	if (page < 0) return;
-	float height = (pdfWidget->pageRect(page).height()/pdfWidget->totalScaleFactor()) / 28.45;
+	const float ptToCm = 2.54 / 72; // 1pt = 1/72 inch = 2.54/72 cm.
+	float height = (pdfWidget->pageRect(page).height()/pdfWidget->totalScaleFactor()) * ptToCm;
 	QClipboard *clipboard = QGuiApplication::clipboard();
 	QString tmp;
-	QTextStream(&tmp) << "" << pos.x() / 28.45  << "," << height- pos.y() / 28.45;
+	QTextStream(&tmp) << "" << pos.x() * ptToCm  << ", " << height- pos.y() * ptToCm;
 	clipboard->setText(tmp);
 	//qDebug() << "Position: " << qPrintable(tmp) << "\n";
-	
 }
 /*!
  * \brief the shortcuts will only be triggered if this widget has focus (used in embedded mode)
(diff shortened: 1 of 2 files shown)


src/pdfviewer/PDFDocument.cpp
```diff
@@ -2733,7 +2747,21 @@ void PDFDocument::setupMenus(bool embedded)
...
-
+/*
+ * Copies to clipboard position in cm captured from the pdf viewer (w.r.t. south west corner)
+ */
+void PDFDocument::getPosFromCLick(const QPointF &pos)
+{
+	int page = pdfWidget->pageFromPos(pos.toPoint());
+	if (page < 0) return;
+	float height = (pdfWidget->pageRect(page).height()/pdfWidget->totalScaleFactor()) / 28.45;
+	QClipboard *clipboard = QGuiApplication::clipboard();
```
sunderme, line 2758: are you sure about the 28.45 ?
I assume 72/inch = ?

egidioln: Hey! No, I'm not sure... I found this approximate conversion from pt to cm online and it worked just fine... Sorry I didn't get your suggestion, how is it?


By do way, any suggestions on how to document this feature?

Best!
Lucas

octaeder: 1pt = 1inch/72.27 = 2.54cm/72.27. So x pt = x / (1/(1pt)) = ( x / (72.27/2.54) )cm where 72.27/2.54 = 28,452756

octaeder: If pt is postscript point not TeX point then use 72.00 instead of 72.27. Hence result is 28.346457.

sunderme: poppler says that pageSize is given in points (i.e. 1/72th of an inch)

egidioln: Thanks! I pushed a new commit with this fix on the conversion. I tested it by picking points from a grid that I drew and the results are way more precise :)

sunderme:
thanks

egidioln:
> In case you want other users to use this, there should be a note in the manual.

Sure, that's also in my todo list... Should I edit https://github.com/texstudio-org/texstudio/blob/master/utilities/manual/usermanual_en.html and open a PR?

sunderme:
yes, as discussed with @octaeder , there is no clear chapter for the pdf viewer yet, so add a section to `3. compile` -> 3.3 pdf viewer -> special features

octaeder:
@egidioln  s. PR #2396, please add your description

egidioln:
> @egidioln s. PR #2396, please add your description

@octaeder  Thanks! I added a paragraph, does it look good?

octaeder:
@egidioln Thanks, I commited your change.

octaeder:
I added your ctrl+shift+click to the magnifier and scroll actions lists

PR #2330, opened 2022-05-15
Luapreview

octaeder:
Allow preview with lualatex.
Note: it seems that package tikz is source of an unresolved problem with format files (s. texstudio-org#2291). In this case disable "Precompile Preamble" in Preview options).
Main changes:
* configdialog.ui: "Preview with lualatex" in drop down list
* buildmanager.cpp: Amend lualatex (CMD_LUALATEX) with -fmt option to use format file. Added pair of braces in old code since compiler complained about ambiguity.


sunderme:
thanks

PR #2333, opened 2022-05-15
add `\/` to latexparser.cpp

mbertucci47:
Add italic correction primitive `\/` to recognized list in latexparser.cpp. It is currently marked as invalid in settings where it shouldn't be, as shown below:
Screenshot 2022-05-15 125334


sunderme:
thanks

PR #2335, opened 2022-05-17
Expand/Collapse for Structure Top

octaeder:
Now the files in the structure panel offer Collapse/Expand Subtitems. Not allways needed but sometimes usefull, for ex. when you have collapsed subitems.

grafik
Don't know why it has been in comments. Subitems like chapters and sections already have it.


sunderme:
thanks

PR #2337, opened 2022-05-19
new cwls

mbertucci47:
New cwls for `procIAGssymp`, `srbtiks`, and the `xymtex` bundle.
Updated cwls for `doc`, the `koma-script` bundle, `expex-acro`, and other small fixes.


sunderme:
thanks

PR #2339, opened 2022-05-19
Update texstudio.cpp

octaeder:
I don't know when this else path it taken (I even commented it out for test). But I strongly assume that widget and name are wrong. You may also compare with next if sections.
@sunderme Please check.


sunderme:
"I don't know when this else path it taken (I even commented it out for test). But I strongly assume that widget and name are wrong. You may also compare with next if sections."

I don't change code from a hunch. Please demonstrate that the code is really wrong.

octaeder:
Because uncertainty remains high I close this request.

Issue #2340, opened 2022-05-19
Side panel name wrong


octaeder:
Open the side panel, choose the symbols panel (or any of them that is not at the top of the secondary toolbar -- which is normally the structure panel). Close txs. The index of the active panel is saved in the ini file. Thus starting txs will restore that panel. But the side panel name shows always "Structure" at that point.

![grafik](https://user-images.githubusercontent.com/102688820/169394785-b7757904-7965-4f1a-a7ee-c27eee55dc60.png) Choose **symbols** panel, then exit txs

![grafik](https://user-images.githubusercontent.com/102688820/169394901-3decef3d-3521-46fe-b54b-7f8772cea336.png) txs remembers the **symbols** panel, but the side panel name shows **Structure**!

If any of the panels is hidden that is on top of the current selected one, then txs fails to restore the same panel and starts with a different one (s. also next comment):

![grafik](https://user-images.githubusercontent.com/102688820/169506027-cd72428a-4aac-4dbd-a2b7-5da42fa67ba2.png) Symbols is active panel, Strucure panel is hidden, then exit txs

txs saves GUI\sidePanel\currentPage=2 in the ini file. Counting starts with 0=TOC, 1=Bookmarks, 2=Symbols in side panel's vertical toolbar (not in the _widgets_ list in the context menu).

![grafik](https://user-images.githubusercontent.com/102688820/169506305-33d141b5-c7bb-4a0f-b7b5-e4625cea34a0.png) After restart of txs the Bookmarks panel is active instead of Symbols (and the side panel name is wrong)
Be aware that Bookmarks is second counting in the _widgets_ list.

### Environment
- TeXstudio: 4.2.3 and devs
- Qt: 6
- OS: Win10
- TeX distribution: miktex

octaeder:
I observed that my solution did not work when the structure panel (it is the first one) is hidden. In this case the current txs (without my changes) already shows a slight error in the gui. If you uncheck (hide) the structure panel then the next (now first) panel (TOC) is activated:
![grafik](https://user-images.githubusercontent.com/102688820/169416406-bc86eaef-d799-4883-95f5-a28b404456f6.png)
Compare this when I (re)check the structure panel:
![grafik](https://user-images.githubusercontent.com/102688820/169416714-1359a396-4562-46a4-8d7b-666949e555f0.png)
In the first case the blue background of the icon (showing that this panel is the current showing) is missing.

Problems:
[...]

PR #2347, opened 2022-05-21
fix side panel issues

octaeder:
s. #2340 
Thanks to @EddieMUC who helped me with an if statement after looking at some files.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -710,12 +710,12 @@ void Texstudio::setupDockWidgets()
 
     leftPanel->showWidgets();
     // restore selected view in sidepanel
-	QList hiddenWidgetsIds = leftPanel->hiddenWidgets().split("|");
+	QList hiddenWidgetsIdsList = leftPanel->hiddenWidgets().split("|");
     int viewNr = configManager.getOption("GUI/sidePanel/currentPage", 0).toInt();
 	int k = -1; // index of visible tool found
 	for (int i = 0; i < leftPanel->widgetCount(); i++)  {
 		QString currentWidgetId = leftPanel->widget(i)->property("id").toString();
-		if (!hiddenWidgetsIds.contains(currentWidgetId)) {
+		if (!hiddenWidgetsIdsList.contains(currentWidgetId)) {
 			k++;
 			if (k == viewNr) {
 				leftPanel->setCurrentWidget(leftPanel->widget(i));
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -710,20 +710,21 @@ void Texstudio::setupDockWidgets()
 
     leftPanel->showWidgets();
     // restore selected view in sidepanel
+	QList hiddenWidgetsIds = leftPanel->hiddenWidgets().split("|");
     int viewNr = configManager.getOption("GUI/sidePanel/currentPage", 0).toInt();
-	int k = viewNr;
-	int i;
-	for (i = 0; i < leftPanel->widgetCount(); i++)  {
-		if (!leftPanel->hiddenWidgets().split("|").contains(leftPanel->widget(i)->property("id").toString())) {
-			if (k == 0) {
+	int k = -1; // index of visible tool found
+	for (int i = 0; i < leftPanel->widgetCount(); i++)  {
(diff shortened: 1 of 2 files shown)


src/texstudio.cpp
```diff
@@ -710,8 +710,21 @@ void Texstudio::setupDockWidgets()
...
     leftPanel->showWidgets();
     // restore selected view in sidepanel
-    int viewNr=configManager.getOption("GUI/sidePanel/currentPage", 0).toInt();
-    leftPanel->setCurrentWidget(leftPanel->widget(viewNr));
+    int viewNr = configManager.getOption("GUI/sidePanel/currentPage", 0).toInt();
+	int k = viewNr;
+	int i;
+	for (i = 0; i < leftPanel->widgetCount(); i++)  {
+		if (!leftPanel->hiddenWidgets().split("|").contains(leftPanel->widget(i)->property("id").toString())) {
+			if (k == 0) {
```
sunderme, line 718: 1) please don't nest to many methods
leftPanel->hiddenWidgets().split("|") -> stringlist abc
def = leftPanel->widget(i)->property("id").toString()
abc contains def

the counting up and counting down of k looks strange

sunderme: so I guess you want to show the nth visible widget, so count k up from 0 and compare it to viewNr ...

src/toolwidgets.cpp
```diff
@@ -539,19 +539,24 @@ void CustomWidgetList::showWidgets()
 	foreach(QAction *act, toolbar->actions()) {
 		toolbar->removeAction(act);
 	}
-	for (int i = 0; i < widgets.size(); i++)
+	bool foundFirstNonHiddenWidget = 0;
```
sunderme, line 542: a boolean is true or false (don't use 0)

src/texstudio.cpp
octaeder: ? I want to show the (viewNr+1)th visible widget (which has index viewNr). If k equals zero then the loop entered (viewNr+1) times the if and we are done. I'll reverse counting.

Tell me more about how do define abc. I already had that idea. But reading the [manual](https://doc.qt.io/qt-6/qlist.html) I found that I should give a max size: QList<QString> list(200);. But I didn't want to set up a limit (what would be a good value?).

octaeder: @sunderme both resolved by latest update

src/toolwidgets.cpp
octaeder: @sunderme resolved by latest update

src/texstudio.cpp
```diff
@@ -710,8 +710,22 @@ void Texstudio::setupDockWidgets()
...
+		if (!hiddenWidgetsIdsList.contains(currentWidgetId)) {
+			k++;
+			if (k == viewNr) {
+				leftPanel->setCurrentWidget(leftPanel->widget(i));
+				emit leftPanel->titleChanged(leftPanel->widget(i)->property("Name").toString());
+				break;
+			}
+		}
+	}
+	if (k == -1) // there are no visible tools
```
sunderme, line 727: I guess, that now it is possible that viewNr is larger than the number of visible widgets (i.e. not widget selected) and k is already increased. Hence line 727 is not true (but should)

sunderme: that should be the last fix.

octaeder: How do you activate a tool with an index viewNr higher than the highest index of the visible tools? Anyway, to do so I testet what happens when I change the number to 20 (which is higher than the numer of availabel tools) in the ini file and start txs. Works fine. Why? Because in this case the initial configuration stays in effect. As you can see from Method showWidgets() this defaults (correctly) to the first visible tool (s.f. variable foundFirstNonHiddenWidget) or none (l. 559). So no fix from my side. One note: There where fixes and build in error (CI, CD) due to QList. I didn't see those errors in my compiles that I did before sending updates. I noticed that linux build runs with  make -j 2, so I changet to that from make -j 8. The latter is given in the wiki, so maybe this should be changed there (or make a note there)?

sunderme: `make -j n` just specifies the number of parallel build processes. It does not change the result.

sunderme:
thanks

PR #2358, opened 2022-05-26
init custom Grid dialog with current values

octaeder:
This PR enhances usage of the custom grid dialog. To reproduce follow instructions:

1. open pdf doc with more than 4 pages in windowed (internal) pdf-viewer
2. menu settings: uncheck Continuous and Single Page Step.
3. setup a grid with 1 row and 2 columns. At first it is not clear what to choose in the Grid menu. Choose 2x1, not 1x2.
4. choose Custom... from Grid menu. Dialog shows with value1 for both x and y. To add a second line to the grid we have to change both values.

**Enhancement:** Initialize the custom Grid dialog to the current grid values.

**Example:** Assume you have a current custom grid of 4x2 and you want to change this to 4x1. This can easily be done now by changing size of y:
grafik

**Btw:** To clarify the meaning of the sublist items, menu Grid now shows as Grid (x,y). So one can clearly distinguish that 2x1 is not the terminology from matrices in math (rows x columns):
grafik

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -1919,6 +1919,11 @@ void PDFWidget::goLast()
 	getScrollArea()->goToPage(realNumPages() - 1);
 }
 
+int PDFWidget::gridRows() const
+{
+	return gridy;
+}
+
 void PDFWidget::goForward()
 {
 	if (pageHistoryIndex + 1 < pageHistory.size()) {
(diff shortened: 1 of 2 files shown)


sunderme:
`Grid (x,y)` does not really make it clearer. (and it looks ugly imho)
This is one of the functions which you don't change constantly and if you do, you learn easily what it means.

sunderme:
The rest is fine.

sunderme:
I put in code here directly

PR #2359, opened 2022-05-26
fix search page display

octaeder:
this PR fixes page display with search result, which shows current page index

**Example:** Searched word _Rätsel_ is on first page.
Before fix:
grafik

After fix:
grafik


sunderme:
Thanks

PR #2362, opened 2022-05-29
simplify copy pos to clipboard

octaeder:
This PR fixes a problem with ctrl+shift+click, which copies page coodinates in cm to the clippboard

- prev. version did not copy from pages on first grid view (i.e. realPageIndex = 0) in case of grid with 2 columns (2<sup>nd</sup> function received page = -1 from pageFromPos)

**Fix:** Merged separate functions, no more Signals. Working example after fix:

grafik

**Btw:** renamed callGetPosFromClick to getPosClick

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -1057,11 +1057,20 @@ void PDFWidget::openAnnotationDialog(const PDFAnnotation *annon)
 	dlg->show();
 }
 
-void PDFWidget::callGetPosFromClick(const QPoint &p){
+/*
+ * Copies to clipboard position in cm captured from the pdf viewer (w.r.t. south west corner)
+ */
+void PDFWidget::getPosFromClick(const QPoint &p){
 	int page = pageFromPos(p);
 	if (page < 0) return;
+	const float ptToCm = 2.54 / 72; // 1pt = 1/72 inch = 2.54/72 cm.
 	QRect r = pageRect(page);
-	emit getPosFromCLick(QPointF(p - r.topLeft()) / totalScaleFactor());
+	QPointF pos = (p - r.topLeft()) / totalScaleFactor();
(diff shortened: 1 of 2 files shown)


src/pdfviewer/PDFDocument.cpp
```diff
@@ -1057,11 +1057,17 @@ void PDFWidget::openAnnotationDialog(const PDFAnnotation *annon)
 	dlg->show();
 }
 
-void PDFWidget::callGetPosFromClick(const QPoint &p){
+void PDFWidget::getPosClick(const QPoint &p){
```
sunderme, line 1060: as this function was added by a contributor, it is not very nice thing to rename it. The contributors should be able to find their code again unless there is really a compelling reason.
Furthermore the describing comment from the now deleted function should be kept.

sunderme: so it should be either "getPosFromCLick" or "callGetPosFromClick"

octaeder: sorry, I'll change back. For explanation: It's name said that the function calls getPosFromCLick (with upper L). But this is not longer true.

sunderme: so getPosFromClick should be fine.

octaeder: sorry, changed already to call...

src/pdfviewer/PDFDocument.cpp
```diff
@@ -2758,18 +2764,7 @@ void PDFDocument::setupMenus(bool embedded)
 /*
  * Copies to clipboard position in cm captured from the pdf viewer (w.r.t. south west corner)
  */
```
sunderme, line 2766: these lines should be above callGetPosFromClick

Here they make no sense at all...

sunderme:
thanks

PR #2371, opened 2022-05-31
show editor's zoom level

### LINKED ISSUE #2233: missing information about editor zoom level
We know that Strg+0 gives a standard for zoom level. But after increasing or decreasing zoom level it is not so obvious which zoom level we currently have. But sometimes I would like to know for easy setup next time. Maybe you could offer this info in the status line, for example. Probably it is known what the zoom factor is (not only the number of steps up or down).

octaeder:
This PR enhances the status line of the editor and resolves #2233 (enhancement request).

**Example:** Step by step decreasing zoom level by pressing ctrl+-:

grafik

grafik

grafik


foreachthing:
If anything, I'd like a percentage instead.
But to be honest, I never thought it was missing at all... imho.

PR #2379, opened 2022-06-04
new cwls

mbertucci47:
New cwls for `apacite`, the `biditufte` bundle, `class-sn-jnl`, `class-svmono`, and `multifootnote`. Updated cwl for `tabularray` and other small fixes.


sunderme:
thanks

PR #2389, opened 2022-06-08
detect screen dpi for config

### LINKED ISSUE #2385: txs sets Screen Resolution to 72dpi
### Reproduce:
1. Delete ini file, then start txs on a screen with 142dpi
2. Open document with several pages, open pdf in embedded pdf-viewer
3. Press button original size (100%). But page shows in size of approx. 50% to 51%.
4. Check config options for pdf-viewer. Screen Resolution is set to 72dpi.

### Environment
- TeXstudio 4.2.3, Qt 6
- OS: Windows 10
- TeX distribution: miktex
- Display 16:9 FullHD

### Expected behavior
txs should detect screen resolution at first start. Then pages will show in original size, when zoom is 100%.

### Solution
s. [Qt](https://doc.qt.io/qt-6/qscreen.html#physicalDotsPerInch-prop)

octaeder:
This PR enhances first start of txs. Instead of setting screen resolution to 72dpi it now uses [Qt](https://doc.qt.io/qt-6/qscreen.html#physicalDotsPerInch-prop) do determine screen resolution. This will avoid irritation, when user views pdf at 100% but finds displayed pages not in original size.

Resolves #2385 

grafik

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -749,7 +749,9 @@ ConfigManager::ConfigManager(QObject *parent): QObject (parent),
 #if (QT_VERSION<QT_VERSION_CHECK(6,0,0))
     int dpi=QApplication::desktop()->logicalDpiX();
 #else
-    int dpi=72; // how to access main screen dpi?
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

Issue #2388, opened 2022-06-08
no propper going to next page after fit to window in some cases


octaeder:
### Reproduce:
1. Start txs, open document with several A4 pages, open pdf in windowed pdf-viewer, select grid 2x2, disable continuous.
2. Resize window such that the viewport has a size of 500 to 900 pixels of width and exactly 628 pixels of height. The viewport is the rectangle in light gray that holds the pages. It is surrounded by a line that is 1 pixel wide and has colour dark gray.
    Note: If you want you can set values in the ini file as given here:

        Geometries\PdfViewerWidth=735
        Geometries\PdfViewerHeight=717

    This will give you a suitable viewport of size 733 x 628 (this may depend on OS and other settings).
4. Press button fit to window. You will find that pressing arrow keys cursor down and up will move the page one pixel up and down, so going to next page is prohibited.

![grafik](https://user-images.githubusercontent.com/102688820/172501582-6a1a5215-9727-4b2d-8ea0-ab7c2ebab123.png)
This image shows a viewport of size 908 x 628 pixels. The info boxes to the right show values used in the methods, where there names are given in the box titles (fitWindow and adjustSize). The two pages of size A4 including the grid  border of 5 pixesl have a width of 445 pixels (rwidth). portHeight and portWidth are the viewport dimensions reduced by 5 pixels for the grid border. Page dimensions are in 1/72 inch (i.e. pt). Vertical scaleFactor is 37,0001%. The calculated total height (rheight) is 629 pixels, so 1 pixel heigher than the viewport.

### Reason
Rounding error in adjustSize:
```
QSize pageSize = (maxPageSizeFDpiAdjusted() * scaleFactor).toSize();
pageSize.rwidth() = pageSize.rwidth() * gridx + (gridx - 1) * GridBorder;
pageSize.rheight() = pageSize.rheight() * gridy + (gridy - 1) * GridBorder;
```
Since dpi equals 72 pageSize.rheight() should be 0,370001 * 841,89 = 311,5001... But by using QSize this is rounded to 312. Thus the final line calulates 312 * 2 + 5 = 624+5 = 629 instead of 311,5 * 2 + 5 = 628. So pageSize should be of type QSizeF.

In this case we do not need the calulation of pageSize since the product in the third line is exactly portHeight -- and that of the second line is portWidth. Maybe these values could be passed to adjustSize.

### Environment
TeXstudio 4.3.0 (git 4.3.0beta1) Qt-Version 6.3.0
[...]

PR #2390, opened 2022-06-08
better accuracy for pageSize on fitWindow

octaeder:
This PR fixes a problem (c.f. #2388) in the pdf-viewer when using fit to window. Reason: Calculating the scaled page size in screen dots is affected by loss of precision (integer arithmetic). Thus in some cases the size of the viewport is 1 pixel less than the result of the calculations. Srcolling 1 pixel appears instead of going to next/previous page.


sunderme:
thanks

PR #2393, opened 2022-06-08
show screen dpi in Pdf Viewer Config

octaeder:
This PR enhances display of the Pdf Viewer Config dialog, which now presents the system screen resolution (s, same line where you set the value). So the user need not calculate the correct value in case she uses a different display or has changed the value for some reason.

### Example
grafik

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -394,6 +394,10 @@ ConfigDialog::ConfigDialog(QWidget *parent): QDialog(parent,Qt::Dialog|Qt::Windo
     // adapt icon size to dpi
     double dpi=QGuiApplication::primaryScreen()->logicalDotsPerInch();
     double scale=dpi/96;
+	int systemdpi = qRound(QGuiApplication::primaryScreen()->physicalDotsPerInch()); // main screen dpi
+	if (systemdpi < 10 || dpi > 1000)
+		systemdpi = 96;
+	QString labelSystemdpi = tr("Screen Resolution (System: %1 dpi):").arg(systemdpi);
 
 	setModal(true);
 	ui.setupUi(this);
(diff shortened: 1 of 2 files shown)


src/configdialog.cpp
```diff
@@ -394,6 +394,10 @@ ConfigDialog::ConfigDialog(QWidget *parent): QDialog(parent,Qt::Dialog|Qt::Windo
     // adapt icon size to dpi
     double dpi=QGuiApplication::primaryScreen()->logicalDotsPerInch();
     double scale=dpi/96;
+	int systemdpi = qRound(QGuiApplication::primaryScreen()->physicalDotsPerInch()); // main screen dpi
+	if (systemdpi < 10 || dpi > 1000)
+		systemdpi = 96;
+	QString labelSystemdpi = tr("Screen Resolution") + " (" + tr("System") + QString(": %1 dpi").arg(systemdpi) + "):";
```
sunderme, line 400: tr("%1").arg(12) works, so please change the text to a single tr expression.

octaeder: thx

sunderme:
thanks

octaeder:
#### Some notes from testing
Following shows results of several functions when running in windows sandbox on a screen with approx. 96dpi:
```
qDebug() << QGuiApplication::primaryScreen()->physicalDotsPerInch() << devicePixelRatio();	// 25.4dpi 1
qDebug() << QGuiApplication::primaryScreen()->logicalDotsPerInch()
    << QGuiApplication::primaryScreen()->QScreen::devicePixelRatio();	// 96dpi 1
qDebug() << QGuiApplication::primaryScreen()->geometry();		// QRect(0,0 1920x1200)
qDebug() << QGuiApplication::primaryScreen()->physicalSize();		// QSizeF(1920, 1200)
```
Since real pixel density of the screen is 96dpi, this would suggest to use logicalDotsPerInch instead of the 4 times lower physicalDotsPerInch. But running in a real windows environment and a monitor of 142dpi, then physicalDotsPerInch shows 142dpi while logicalDotsPerInch still shows 96dpi. So it seems better to stay with physicalDotsPerInch.

octaeder:
The value of 25 dpi for physicalDotsPerInch happens if one switches to a windowed sandbox. Even after going back to maximized (fullscreen) the value keeps 25 dpi. If you now turn off and on your monitor the value will be correct and you can check the custom value in the config if it is the same.

PR #2399, opened 2022-06-09
mark grid menu entry after canceling Grid Dialog

octaeder:
As discussed [here](https://github.com/texstudio-org/texstudio/commit/05e561aa390366d8df49287476395f2b4f10eaeb#commitcomment-74580745) this PR fixes the problem after canceling the custom grid dialog that entry _Custom Grid..._ is checked in the menu instead of a normal grid when possible. Code used is same by @sunderme in 2e7a5b3, thx.


sunderme:
thanks

PR #2402, opened 2022-06-11
Added cut, copy and paste functionality to the Array and Tabular Wizard.

fjpereny:
# Cut / Copy / Paste Tabular Wizard
## Previous Issue
Users needed to manually type in each field, it did not allow cut/copy/paste/delete like a normal spreadsheet application.  This prevented users from quickly putting tables in from other applications.
## Solution
Added additional functions
Users can now cut / copy / paste / delete their selection in the Array and Tabular Wizard windows.  This data is normally tab separated.

The following shortcuts now work on the tables:

CTRL + C = COPY
CTRL + X = CUT
DELETE = DELETE
CTRL + V = PASTE

### DIFF
--- a/src/arraydialog.cpp
+++ b/src/arraydialog.cpp
@@ -120,7 +120,7 @@ void ArrayDialog::keyPressEvent(QKeyEvent *event)
         int end_row = list[0]->row();
         int start_col = list[0]->column();
         int end_col = list[0]->column();
-        for (QTableWidgetItem *item : list)
+        for(QTableWidgetItem *item : list)
         {
             if (item->row() < start_row)
             {
@@ -140,9 +140,9 @@ void ArrayDialog::keyPressEvent(QKeyEvent *event)
             }
         }
 
-        for (int row=start_row; row <= end_row; ++row)
+        for(int row=start_row; row <= end_row; ++row)
         {
-            for (int col=start_col; col <= end_col; ++col)
+            for(int col=start_col; col <= end_col; ++col)
             {
                 QTableWidgetItem *item = ui.tableWidget->item(row, col);
                 copied_text += item->text();
@@ -182,7 +182,7 @@ void ArrayDialog::keyPressEvent(QKeyEvent *event)
         int end_row = list[0]->row();
         int start_col = list[0]->column();
         int end_col = list[0]->column();
-        for (QTableWidgetItem *item : list)
+        for(QTableWidgetItem *item : list)
         {
             if (item->row() < start_row)
             {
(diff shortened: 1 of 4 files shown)


src/arraydialog.cpp
```diff
@@ -90,3 +90,206 @@ void ArrayDialog::newRows(int num) {
...
+        clipboard->setText(copied_text);
+    }
+
+    // Table Paste
+    // Allows pasting of cell data from spreadsheet applications such as LibreOffice Calc.
+    if ((event->modifiers() == Qt::CTRL) && (event->key() == Qt::Key_V))
+    {
+        QClipboard *clipboard = QGuiApplication::clipboard();
+        QString pasted_text = clipboard->text();
+        QList<QString> rows = pasted_text.split('\n');
```
sunderme, line 229: should be changed to
`QList<QString> rows = pasted_text.split('\n',Qt::SkipEmptyParts);`

Otherwise always a new empty line is added after the pasted content.

src/tabdialog.cpp
```diff
@@ -413,3 +413,206 @@ spancolor.setAlphaF( 0.2 );*/
...
+        clipboard->setText(copied_text);
+    }
+
+    // Table Paste
+    // Allows pasting of cell data from spreadsheet applications such as LibreOffice Calc.
+    if ((event->modifiers() == Qt::CTRL) && (event->key() == Qt::Key_V))
+    {
+        QClipboard *clipboard = QGuiApplication::clipboard();
+        QString pasted_text = clipboard->text();
+        QList<QString> rows = pasted_text.split('\n');
```
sunderme, line 552: same as above

`QList<QString> rows = pasted_text.split('\n',Qt::SkipEmptyParts);`

src/arraydialog.cpp
fjpereny: Thanks and good catch.

src/arraydialog.cpp
```diff
@@ -90,3 +90,206 @@ void ArrayDialog::newRows(int num) {
...
+            if (indexList.count() > 0)
+            {
+                starting_row = selection_model->selectedIndexes().begin()->row();
+                starting_col = selection_model->selectedIndexes().begin()->column();
+            }
+        }
+
+        int cur_row = starting_row;
+        int cur_col = starting_col;
+        for (QString row : rows)
```
sunderme, line 247: and maybe more a style point (as the compiler warns)
`for (const QString &row : rows)` 

same below for cols

fjpereny: Fixed, thanks.

sunderme:
Sorry about that but Qt5 uses a different constant for skip empty parts: `QString::SkipEmptyParts`
You need to distinguish Qt5/Qt6 now, e.g cleandialog.cpp lines 102..

sunderme:
thanks

PR #2403, opened 2022-06-11
fix info in pdf-viewer docks (issue #2119)

### LINKED ISSUE #2199: windowed pdf viewer opens empty content view
### Environment
- TeXstudio: 4.2.2
- Qt: 5.15.2
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior case 1
If the viewer shows a pdf and you open the contents view (menu Window/Show), then the view should show the pdf contents (if there is one) or No TOC (if there is none).

### Actual behavior case 1
If the windowed viewer starts such that the contents view is not open, then the following happens: If you open the contents view, it will stay empty. If you close the viewer (without closing the contents view), then the contents (or No TOC) will show upon the next start of the viewer. Another workaround is to switch to the embedded viewer, then switch back to the windowed viewer.

Addition:
### Expected behavior case 2
When I preview a pdf of a latex document, then switch to a tab of txs editor of another document and click view, then content view dock and all other pdf information docks of the pdf-viewer should show data of the second document.

### Actual behavior case 2
If any of the docks of the pdf-viewer is open, this works as expected. But if all docks had been closed while viewing the first document then data of the first document may be presented in the docks you open now while viewing the second document. It may even happen that de contents dock shows a mixture of the first and the second document:

grafik
page 3 is from first document, page 4 is the end of the contents of the second document

octaeder:
This PR fixes #2199. The issue showed that information presented in the docks of the windowed pdf-viewer may be missing (pdfOutline) or be of a previous document (pdfOutline, pdfInfo, pdfOverview, ...). These docks can be enabled and disabled via menu Window/Show. The issue depends of the visibility of the docks at time you open the viewer or switch to another document.

The small change in code ensures that all docks are assumed not beeing filled when loading a new document, so data will be retrieved as necessary. Disabeling and (re) enabeling a dock will not reload data. In case that the document has no structure (chapter, ...) the pdfOutline shows _No TOC_.

### DIFF
--- a/src/pdfviewer/PDFDocks.cpp
+++ b/src/pdfviewer/PDFDocks.cpp
@@ -55,6 +55,7 @@ PDFDock::~PDFDock()
 
 void PDFDock::documentLoaded()
 {
+	filled = false;
 	if (!isHidden()) {
 		fillInfo();
 		filled = true;


sunderme:
thanks

PR #2404, opened 2022-06-11
Fixed logical error that caused a new item to be created every time.

fjpereny:
Currently iteration of the loop will allocate new memory.  It should only do so when the existing item does not exist (null pointer).  The logic was previously there but not implemented correctly.


sunderme:
thanks

PR #2410, opened 2022-06-13
fix Update check (issue #2244)

### LINKED ISSUE #2244: Check for updates not correct in some case
With current txs 4.2.2 Qt5 on win10 I observe following: Open config dialog. It shows Update Level "Stable Releases". I change this to "Dev. Versions" and let immediately check for updates (do not close config dialog). Dialog says "txs is-up-to date". Of course this is not true. Close and reopen config dialog. I see Update Level "Dev. Versions" as expected. Check for updates now shows as expected:

grafik

If you now change to "Stable Release" without closing config dialog will show prev. image.

### Expected behavior
Check correctly for updates according to (changed) setup without closing config dialog.

octaeder:
The GUI has been slightly reworked as can be seen in #2244, because of some issue with displaying of non stable current version. Also alpha versions are no longer ignored.


src/updatechecker.cpp
```diff
@@ -121,6 +128,25 @@ void UpdateChecker::parseData(const QByteArray &data)
...
+//	+ "{\"ref\":refs/tags/4.0.0alpha7,"
+//	+ "{\"ref\":refs/tags/4.0.0alpha8,"
+//	+ "{\"ref\":refs/tags/4.0.0beta1,"
+//	+ "{\"ref\":refs/tags/4.0.0beta2,"
+//	+ "{\"ref\":refs/tags/4.0.0beta3,"
+//	+ "{\"ref\":refs/tags/4.0.0rc1,"
+//	+ "{\"ref\":refs/tags/4.0.0rc2"
+//	;
+//*************** T E S T **************
+
```
sunderme, line 149: this looks like it needs to be put in autotests.
See directory src/tests

src/updatechecker.cpp
```diff
@@ -41,8 +42,14 @@ void UpdateChecker::autoCheck()
...
 
-void UpdateChecker::check(bool silent)
+void UpdateChecker::check(bool silent, int currentComboBoxUpdateLevel)
 {
+	// catch value if possible, s. comment at start of checkForNewVersion
+	if (currentComboBoxUpdateLevel == -1)
+		comboBoxUpdateLevel = -1;
+	else
+		comboBoxUpdateLevel = currentComboBoxUpdateLevel;
+
```
sunderme, line 52: this makes no sense, it is basically `comboBoxUpdateLevel = currentComboBoxUpdateLevel;`

sunderme:
There is already a test for version comparison which now fails. Please check and fix either test or code whichever is wrong.

octaeder:
new PR #2420

PR #2420, opened 2022-06-20
fix update check

octaeder:
This PR fixes update check and enhances display of update check notification.
Related informations: Q&A #2413, Issue #2244

### Issues fixed

- Update check doesn't recognize released alpha versions, even when the update level is set to Development Versions in config dialog.
- Assume current version to be "4.3.0beta2-1-g5c925a387", a build of master one commit after 4.3.0beta2. Then txs lets revision be 1, so the internal version is 4.3.0beta1. Hence updater would display incorrectly a new 4.3.0beta2, even so the build is based on a commit after beta2. This could be irritating or misleading.

### Enhancements

- When txs is up-to-date then update notification now includes current version (as all other update notifications do).
- Update notifications now show type of versions (alpha, beta, rc, stable possible, but is not normally shown). This avoids messages like the following ones:

    grafik
    (0) in first row is not the revision which is 1 (4.2.3rc1), so suggesting a stable version is running. But it is a release candidate that could be updated with next stable version or later beta1.

    If update level is set to stable then update check shows:

    grafik
- If available, update notification also shows the commit number from the git revision. Ex. 4.3.0beta2-1 for 4.3.0beta2-1-g5c925a387 (s. above). This will not happen with officially released versions, since their git revisions lack that number (saying it is zero):

    grafik
- New function versionToString allows easy printing of version object's data in notifications:

    grafik

- Extraction of data strings per version into a list of data strings is now done via a very small parser. This avoids needless lines to be parsed after splitting the JSON string from github api at commas.
- More tests available.
 
### Overview of code changes

utilsversion_t.h:
- test of new function parseGitData (simple parsing of git data)
- new test of function parseVersionNumber (string to list)
- test of new function stringVersion2Parts (split version string into string parts)
- new test of function current version
- test of new function versionToString (convert version object to string)
- new test of function isValid (validation of version)

utilsVersion.h:
- removed gitRevisionToInt
- removed parseGitRevisionNumber
- added function parseGitData (simple parsing of git data)
- added function stringVersion2Parts (split version string into string parts)
- added function versionToString (convert version object to string)
- new value commitsAfter

utilsVersion.cpp:
- removed gitRevisionToInt
- removed parseGitRevisionNumber
- added function parseGitData (simple parsing of git data)
- added function stringVersion2Parts (split version string into string parts)
- amend function current
- new function versionToString (convert version object to string)
- amend/simplify operator >
- accept alpha as a version tag type (development) and allow types in any case
- new value commitsAfter

updatechecker.cpp:
- use simple parser for git data
- accept alpha as a version tag type (development) and allow types in any case
- amend update notification

### DIFF
--- a/src/tests/utilsversion_t.h
+++ b/src/tests/utilsversion_t.h
@@ -12,12 +12,141 @@ class VersionTest: public QObject{
 public:
 	VersionTest(bool executeAllTests) : allTests(executeAllTests) {}
 private slots:
+	void parseGitData_data() {
+		QTest::addColumn<QString>("gitData");
+		QTest::addColumn<QStringList>("list");
+		QTest::addColumn<bool>("valid");
+
+		// s. issue #2244
+		QTest::newRow("parseGitData")
+			<< QString( "[{\"ref\":\"refs/tags/3.1.2\",\"object\":{...}},")
+					+ "{\"ref\":\"refs/tags/4.0.0alpha1\",\"object\":{...}},"
+					+ "{\"ref\":\"refs/tags/4.0.0alpha2\",\"object\":{...}},"
+					+ "{\"ref\":\"refs/tags/4.0.0alpha3\",\"object\":{...}},"
+					+ "{\"ref\":\"refs/tags/4.0.0alpha4\",\"object\":{...}},"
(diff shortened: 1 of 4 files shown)


sunderme:
thanks

PR #2432, opened 2022-06-24
custom grid: always check standards

octaeder:
after confirming custom grid dialog (like after canceling) check for suitable standard settings, because you may set x=2 and y=2: Observed while testing new 4.3.0beta2.


sunderme:
thanks

PR #2443, opened 2022-06-30
pdf-viewer CrossCursor for Ctrl+Shift

octaeder:
In the pdf-viewer signal action "copy position to clipboard" (ctrl+shift then click) with cross cursor (s. [ Qt Cursor Shape enum](https://doc.qt.io/qt-6/qt.html#CursorShape-enum), value 2) instead of
- the zoomIn cursor while in magnifier mode, or
- the openHand cursor while in scroll mode,

the first being misleading and the second imprecise to use.

Refer to #2324 by @egidioln.


sunderme:
thanks

PR #2457, opened 2022-07-05
Install Win32 as Unix

podsvirov:
We use this to make [package](https://github.com/msys2/MINGW-packages/pull/11958) for [MSYS2](https://www.msys2.org/) distro.

### DIFF
--- a/texstudio.pro
+++ b/texstudio.pro
@@ -172,7 +172,7 @@ macx {
 }
 
 # ###############################
-unix:!macx {
+if(unix|win32):!macx {
     isEmpty( PREFIX ):PREFIX = /usr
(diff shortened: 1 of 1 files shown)


podsvirov:
Custom scripts that I can see (in `.github/scripts`) can make standalone packages. But package in distro is not some as standalone pakcage. Part of installation logic available in project file (`.pro`) and used for `Linux`. It can be reused for more platforms.
Suggested changes do not break some thing.
I just want to reuse install rules generated from project file like:

```console
make INSTALL_ROOT="${pkgdir}" install
```

sunderme:
obsolete with transition to cmake

podsvirov:
Wow, `transition to cmake` - good news! Thanks!

PR #2470, opened 2022-07-11
new zoom-in/out-magnifier cursors

octaeder:
The custom _zoom-in_ and _zoom-out_ cursors are scaled from size 16x16 to 32x32. This yields slightly unsharp and oversized images (s. below). Both do not optimally express the operation of changing the zoom level. Especially the zoom-in cursor looks more like some cross hair that suggests you should target at something on the screen (what makes no sense here). I present below new cursors (second row) which could replace these (first row):

grafik


sunderme:
I would prefer svg images

octaeder:
which free sw can be used for this without harm?

octaeder:
but original cursors are svg files.  But this doesn't prevent them from becoming unsharp after scaling.

sunderme:
> which free sw can be used for this without harm?

you can use inkscape for drawing svg

sunderme:
thanks

PR #2478, opened 2022-07-16
add zoom-in/-out-magnifier cursors for modern (fix)

octaeder:
Add to modern same zoom-in/-out-magnifier images from classic (s. PR #2470), but with "modern" colour. The new files with -magnifier names (same as for classic) fix the problem from #2470 that modern won't find these cursor files (fixed names in PDFDocument.cpp, l. 615f). Screen shots:

grafik

Old files (zoom-in/-out) are still needed in menu and zoom slider:

grafik    grafik


sunderme:
thanks

PR #2480, opened 2022-07-16
scroll hand cursor with marked hot spot

### LINKED ISSUE #2201: Show active point of "scroll hand" mouse pointer of pdf viewer
I use synctex very often to jump back from some text in the viewer to the position of that text in the tex file. But often synctex is not able to locate the correct position. Then I observed that increasing zoom of the preview say from 94% to 150% increased hit rate so much that synctex even marked single words in the tex file instead of only one or two lines containing the word. Poorly, with this zoom the page does not fit in the window.

Reasons why to show the hot spot of the _open hand cursor_ and move it outside of the hand:

1. precise click on a position would be easy
2. text below hot spot is visible for better accuracy
3. zoom in (double click) to a specific point (this is pointed out in PR #2468 that addresses the hot spot of the magnifier)

I prepared some sample cursors that I will show in the next comment.

Compare with gimp (be aware of the small text under the hand):

Textstudio:
      grafik

Gimp:
      grafik

octaeder:
Reasons why to show the hot spot of the _open hand cursor_ and move it outside of the hand (fixes #2201):

1. precise click on any position would be easy,
2. text below hot spot is visible for better accuracy for synctex and for
3. zoom in (double click) to a specific point (this is pointed out in #2468 that addresses the hot spot of the magnifier)

Comparison old vs. new:
grafik grafik

new file is focusedhand.svg  (size 128x128, replaces standard Qt openhand cursor) can be used for classic and modern


sunderme:
thanks

PR #2488, opened 2022-07-22
add completion files for (u)pTeX

mbertucci47:
cwls added for formats `ptex`, `platex`, `uptex`, and `uplatex`, the helper package `plext`, and those formats' base classes (`jarticle`, `ujarticle`, etc.)


sunderme:
thanks

PR #2496, opened 2022-07-25
resize rows of error table to contents and some more

octaeder:
This PR will fix small issues with the Error Log Table and adds small enhancements:

####  How to reproduce issue
1. Compile a file and preview pdf (embedded) and adjust width of pdf-viewer such that messages don't fit in one line, close txs. Open txs, open Error Log Table, messages have line breaks:
grafik

2. Enlarge width of editor (or double click on the "message" header section's right edge) so message needs one line. but height doesn't decrease:
grafik
Also when you decrease width, heigth remains constant, even if more space is needed (for 3 lines or more).

4. Compile. Messages now show in one line, height of rows now as needed, no way to come back to line breaks if needed (even if you decrease width of "message" header section). To read message up to the end you have to scroll to the right:
grafik
This should be the default look, also if you reopen txs (s. 1.)

5. The Log always indicates at first time that column File is sorted in descending order. But this is not true, files are listed in order they were encountered.

6. As the following image shows sorting by column Line is not correct. Due to the redundant string "line" in each field the entry is not treated as a number (two lines not correct, compare to the right after fix):
grafik grafik

#### Enhancements
1. It should be possible that line breaks (re)appear in last column by decreasing column width.
2. It should be possible to reorder columns, when you need to. Ex.: Column Type moved to the left (by dragging):
grafik
Note: How fare you have to drag is not that visible in default style (General Config), you may use Fusion style:
File header still white, not dragged far enough to switch:
grafik
File header gray, dragged far enough to switch:
grafik
3. Keep natural ordering of rows and show no sort indicator for File column. Further allow disabling sort by multiple clicks on the column header (cycles up, down, off), but this is available since Qt6.1 only.

#### Implementation
Use section resize mode "resize to content" for rows. No default heigth needed.
Use signal sectionResized to resize rows to content.
Make horizontal header sections movable.
After loading error table let last column stretch.
Set sort by column to -1 to suppress sort indicator.
Set sort indicator clearable to true (if Qt version is at least 6.1.0).

### DIFF
--- a/src/latexlog.cpp
+++ b/src/latexlog.cpp
@@ -44,7 +44,7 @@ QVariant LatexLogModel::data(const QModelIndex &index, int role) const
 	case 2: {
 		int line = log.at(index.row()).oldline;
 		if (line > 0)
-			return tr("line") + QString(" %1").arg(line);
+			return line;
 		return QString();
 	}
 	case 3:
--- a/src/latexlogwidget.cpp
+++ b/src/latexlogwidget.cpp
@@ -44,6 +44,11 @@ LatexLogWidget::LatexLogWidget(QWidget *parent) :
 	errorTable->setMinimumHeight(5 * rowHeight);
 	errorTable->setFrameShape(QFrame::NoFrame);
 	errorTable->setSortingEnabled(true);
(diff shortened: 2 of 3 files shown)


sunderme:
thanks

Issue #1935, opened 2021-11-12
"Fit to Text Width" does not update zoom when textwidth changes


krzgvtrs:
### Environment
- TeXstudio: 4.0.4 (git n/a)
- Qt: 5.12.8, compiled with Qt 5.12.8 R
- OS: Linux (Pop OS 20.04 LTS)
- TeX distribution: miktex

### Expected behavior
If the option "Fit to Text Width" is marked, the zoom of the viewer should always be in consonance with the text width.

### Actual behavior
If text width changes, the zoom keeps the same. The only way to make it update is to close and reopen texstudio or to change the viewer mode (from embedded to windowed or viceversa).

### How to reproduce
Make a new document and write only a word. Compile and view. Then add a paragraph. Compile and view. The zoom in the preview does not adapt to the new text width. 
Another way to reproduce is to have two documents with different margins and change the view from one to the other.

bryango:
A popup like the one below will appear when the viewer is re-calculating text width:

![texstudio-calculating-text-width](https://user-images.githubusercontent.com/26322692/181764512-5e49c5d6-356a-4349-b5e5-c75b5a3dc94a.png)

It steals the window focus so I couldn't keep working on the document. This is okay once in a while, but this happening after _every_ rebuild is super annoying, especially when doing successive rebuilds (e.g. when tweaking a TikZ diagram). 

I hence propose that we either:
- redesign the popup, making it less disruptive to the workflow, or
- simply revert this change e9cfcec7a8a73045882b6551d71deeb679401989, i.e. stop re-calculating text width after every rebuild. 

In fact, I would say that recalculating every time is unnecessary, since for most documents the width is more or less fixed. I have hence created a fork reverting this commit. This would of course generates a regression for @krzgvtrs. If you are okay with it, I can open a pull request and ask @sunderme to merge so that it becomes the default behavior. If not then I will keep this patch on my private branch. 

Personally I really really need the popup to go away because it has seriously affected my workflow. Sorry for the rambling.

PR #2503, opened 2022-07-29
Stop re-calculating text width after rebuilds

bryango:
Revert e9cfcec7a8a73045882b6551d71deeb679401989, unfix #1935. See: https://github.com/texstudio-org/texstudio/issues/1935#issuecomment-1199282889 for the explanations.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -705,7 +705,6 @@ void PDFWidget::setDocument(const QSharedPointer<Poppler::Document> &doc)
 	document = doc;
 	maxPageSize.setHeight(-1.0);
 	maxPageSize.setWidth(-1.0);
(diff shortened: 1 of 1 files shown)


sunderme:
I am not reverting the old fix.
You can come up with a better fix but giving up one functionality for another again is not the solution.

octaeder:
I cant't reproduce with the 4.3.0beta3 version for win10 Qt6 (maybe I need more details). I typically use inline preview (with pdflatex quite fast) for finetuning tikz pictures, which I think is very nice.

bryango:
> I am not reverting the old fix. You can come up with a better fix but giving up one functionality for another again is not the solution.

It's okay, I understand. I'll keep this in my private fork then. I couldn't come up with a better fix because I'm not familiar with C nor the codebase. I guess I'll keep patching the code for my own use then.

PR #2506, opened 2022-07-30
new cwls

mbertucci47:
New cwls for `boxdims`, `class-aalok`, the `exam-zh` bundle, and `nndraw`.
Updated cwls for `hyperref`, `nicematrix`, and `pbalance`.


sunderme:
thanks

PR #2527, opened 2022-08-11
fix update check if current version type is stable

octaeder:
This PR will fix a problem with current stable version 4.3.0. When updates are checked result is:

image
The second 0 is the revision that is printed after the version type (which should be "stable", s. next section).

Reason: Conversion of string "4.3.0" into a version object leaves type as null string instead of assigning "stable". This was intended to remember that the string was not the equivalent one "4.3.0stable0" (which would be accepted but is never used). But printing the version always suppresses stable0, so a null string for type is obsolete.

This is fixed and I added printing some debug information. Further I added a message for the case that the current version object isn't valid (note that type is emtpy in the message):

image
This is accompanied by a new auto test for this case.

Final test running txs now shows the message expected:

image

### DIFF
--- a/src/tests/utilsversion_t.h
+++ b/src/tests/utilsversion_t.h
@@ -210,6 +210,7 @@ private slots:
 		QTest::newRow("valid2.6") << "3.0.1" << "alpha" << 0 << 0 << true;
 		QTest::newRow("valid2.7") << "3.0.1" << "alpha" << -1 << 0 << false;
 		QTest::newRow("valid2.8") << "3.0.1" << "alpha" << 1 << -2 << false;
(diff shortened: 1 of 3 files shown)


sunderme:
there is still debug code in the PR ...

octaeder:
I thought this is ok, since it only runs if you have a debug build and run the update check.

sunderme:
I would prefer not to have debug code in it.
It makes things confusing if I look for something else.

sunderme:
and it messes up autotest output

sunderme:
I would prefer to remove the debug code completely as it degrades code readability

sunderme:
thanks

PR #2532, opened 2022-08-15
new cwls

mbertucci47:
New cwls for the `bithesis` bundle, `class-shtthesis`, the `tudscr` bundle, and `hereapplies`. Small updates elsewhere.


sunderme:
thanks

PR #2539, opened 2022-08-18
Make reading configuration window size more robust (Wayland issue)

hseliger:
Make reading configuration window size more rebust, so it can deal with incorrect x and y values for the top left corner of the window, as are being returned under **Wayland**. The changes makes sure that even if the position of the window is nonsense, the window size can get restored (as long as the window fits the screen).

Needs 3 temporary variables, as the second call setting `screen` delivers incorrect values for `width`, `height`, and `right` if the point coordinate is off the screen.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -4183,17 +4183,23 @@ void Texstudio::readSettings(bool reread)
     config->beginGroup("texmaker");
 
     QRect screen = QGuiApplication::primaryScreen()->availableGeometry();
-    int w = config->value("Geometries/MainwindowWidth", screen.width() - 100).toInt();
+	// Save the screen dimensions, as the next call will overwrite the correct values
+    int screen_w = screen.width();
+	int screen_h = screen.height();
+	int screen_r = screen.right();
+	// Read window size from config
+	int w = config->value("Geometries/MainwindowWidth", screen.width() - 100).toInt();
     int h = config->value("Geometries/MainwindowHeight", screen.height() - 100).toInt() ;
     int x = config->value("Geometries/MainwindowX", screen.x() + 10).toInt();
     int y = config->value("Geometries/MainwindowY", screen.y() + 10).toInt() ;
     screen = UtilsUi::getAvailableGeometryAt(QPoint(x, y));
     if (!screen.contains(x, y)) {
         // top left is not on screen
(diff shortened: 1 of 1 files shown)


src/texstudio.cpp
```diff
@@ -4183,17 +4183,23 @@ void Texstudio::readSettings(bool reread)
...
     if (!screen.contains(x, y)) {
         // top left is not on screen
+		//Problem: then the other coordinates are incorrect as well!! So we need to use width and height from above
         x = screen.x() + 10;
         y = screen.y() + 10;
-        if (x + w > screen.right()) w = screen.width() - 100;
-        if (y + h > screen.height()) h = screen.height() - 100;
+        if (x + w > screen_r) w = screen_w - 100;
+        if (y + h > screen_h) h = screen_h - 100;
     }
```
sunderme, line 4203: I don't see that the code changes anything. QRect screen is used directly or indirectly for the same function.

hseliger: Because when you call `screen = UtilsUi::getAvailableGeometryAt(QPoint(x, y));`, the original and at that point correct screen information gets replaced by data for the specific point at x/y (which in Wayland is off-screen because of the incorrect x/y saved in the settings). And since the point is off, at least under Wayland you get simply a 0 back for right(), height() and width(). I put in a bunch of debug lines to find that out.

So the only way to get the correct screen size and thus restore the correct window size from settings is to remember the screen right/width/height from the first call.

Without this change I always get a very small default size window, regardless of what is saved in the config.

sunderme: I see.
I have just commited a different approach on master branch which discards the check if screen is not valid (too small).
Please check.
The main idea of that branch is to adapt the window if it is on a different sized screen and your code assumes that the sizes are always the same...

hseliger: Thanks! This approach works for me! While, of course, the weird values for x and y still are in the config (should I find the time, I'll dig around in QT and file a report with them, but that may be beyond my time budget), now the window size does get restored correctly even if the top right coordinate is out of whack.

I'll go ahead and close this change request then.

hseliger:
Closing pull request as it is made obsolete by sunderme's alternative approach [#2539](https://github.com/texstudio-org/texstudio/pull/2539)

PR #2561, opened 2022-08-29
new and updated cwls

mbertucci47:
New cwls for `luaquotes`, `swungdash`, and `tikzpingus`.
Updated cwls for `ProfCollege`, `markdown`, `scripture`, and `simpleicons`.


sunderme:
thanks

PR #2562, opened 2022-08-31
new and updated cwls

mbertucci47:
New cwls for `exam-zh-textfigure`, `langnames`, and `proof-at-the-end`.
Updated cwls for `biblatex`, `covington`, `crefthe`, `exam-zh`, and `xgreek`.


sunderme:
thanks

PR #2572, opened 2022-09-06
more cwls

mbertucci47:
New cwls for `class-ltnews`, `democodelisting`, `democodetools`, and `subdepth`.
Updated cwls for `colorist`, `fduthesis`, and `minimalist` bundles.


sunderme:
thanks

PR #2580, opened 2022-09-13
Update usermanual_en.html

octaeder:
Updating Contents as follows:

- add some subsections not present in Contents
- indent subsections in Contents with respect to sections for better readability
- changing paragraphs of section "2.15 Special Commands" into subsections and add these to Contents to clarify what these special editing commands are


sunderme:
thanks

PR #2583, opened 2022-09-15
new and updated cwls

mbertucci47:
New cwls for `class-milog` and `textcsc`.
Updated cwls for `ProfCollege`, `langnames`, and other small updates/fixes.


sunderme:
thanks

PR #2593, opened 2022-09-18
Update usermanual_en.html

octaeder:
Only when you select 2 columns with one row (i.e. 2x1) grid page offset is set to 1 automatically.


sunderme:
thanks

PR #2601, opened 2022-09-20
Update usermanual_en.html

octaeder:
note: bookmarked text receives no updates


utilities/manual/usermanual_en.html
```diff
@@ -500,7 +500,8 @@ <h3 id="SECTION143">2.4.3 Using Bookmarks</h3>
 
 <p>You may choose the Bookmarks view of the left side panel, which shows the text for each line with a bookmark of all editors. With the ease of a mouse click you
 can jump to the text line in the editor. A tooltip shows the surrounding context of the line (s. left image below). The lines are ordered in the order you created
-the bookmarks. A context menu allows you to change the order of lines or remove bookmarks (s. right image below).</p>
+the bookmarks. A context menu allows you to change the order of lines or remove bookmarks (s. right image below). Bookmark text saved in the list of the view will
```
sunderme, line 503: "Bookmark text saved in the list of the view will not reflect later changes to the text in the editor." , not sure what that is supposed mean.

octaeder: The text of a line is copied to a list view entry (of the view Bookmarks) when you set a bookmark for that line. This copy of the text will not change when you change the text in the source (i.e. in the editor).

sunderme: "in the list of the view" does not sound right. Please rephrase.

octaeder: Suggestion:
The text of a line is copied to view Bookmarks when you set the bookmark. The copied text will not change when you edit the text later on.

sunderme: "to view bookmarks" is not a term used by users.
I would not necessarily mention this as it is quite evident.
basically the bookmark text is not updated as you type.

sunderme:
thanks

PR #2602, opened 2022-09-20
change order of items in bookmark contextmenu

### LINKED ISSUE #2598: About the Ordering of Bookmark Entries in the Bookmark Context Menu
Following image shows (from left to right) the _Goto Bookmark_ menu, the _Toggle Bookmark_ menu, and the context menu of the bookmark column left to the editor's text area:

grafik

The menus show Bookmark 0 as the last one (as if it where Bookmark 10), but ordering in the context menu is different. Not a big deal. But when you click into the bookmark column over and over this will create the sequence _Bookmark 1, 2, ..., 9, 0_ (continued by unnamed bookmarks):

grafik

So I suggest using the same order in the context menu as given by the _Toggle Bookmark_ menu.

### Environment
- TeXstudio: 4.3.1, Qt6, Win 10, miktex

Thx

octaeder:
same ordering as in menus _Goto Bookmark_ and _Toggle Bookmark_, i.e. Bookmark 0 at bottom
this PR resolves #2598

### DIFF
--- a/src/latexeditorview.cpp
+++ b/src/latexeditorview.cpp
@@ -3215,9 +3215,10 @@ void LatexEditorView::lineMarkContextMenuRequested(int lineNumber, QPoint global
 	act->setData(-1);
 	menu.addAction(act);
 
-	for (int i = 0; i < 10; i++) {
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #2607, opened 2022-09-23
Update usermanual Chapter 1 Configuring TeXstudio

octaeder:
Explanation: Configuration offers the options partitioned into groups (subsets).
Changes:
- added tabular overview with the available option groups as links
- new screenshot for each of the 15 groups replace old images
- new images show advanced options highlighted/framed
- reordered sections of chapter 1 to same order as the groups
- added sections for four missing groups


sunderme:
it looks like the picture contain marked areas but I don't find reference to that marked areas in the text ? Why the marks ?

octaeder:
after the new  table we find:

The following sections show screenshots of the configuration for these groups including advanced options. The advanced options are framed in blue (but only in the pictures). So you can easily recognize which options are not available in standard mode. Groups only present in advanced mode show a large blue frame surroundig all options.

octaeder:
immediately before 1.1

sunderme:
I read in the commentary that it is advanced/default. That is confusing.
I would suggest to stick with default.
In general I find the current structure of the manual a bit intimidating. 
You **can** configure a lot but normally you don't have to change a thing.
Apart from that, some text seems really to be outdated, e.g. `Also the log viewer can appear tabbed which allows faster change between error table, log view and previewer ... ` (if that can be changed, for sure not on the general page)

sunderme:
Maybe the time has come to switch to markdown. Any input from you @dbitouze ?

sunderme:
> after the new table we find:
> 
> The following sections show screenshots of the configuration for these groups including advanced options. The advanced options are framed in blue (but only in the pictures). So you can easily recognize which options are not available in standard mode. Groups only present in advanced mode show a large blue frame surroundig all options.

as logically as that seems, manuals are skimmed over. Images need to be self-explanatory as much as possible.

octaeder:
Many things are outdated and others even have not found its way in the manual. You find many examples in my changes which proof that for years txs was changed but not the manual. There are images (not in this chapter) I wonder why there is a chameleon head drawn, must be very old and is random. Main target for last days was replacing images and reordering of chapter 1. Reordering alone took a lot of time.

grafik

Decision for advanced settings images was that we can describe things when necessary. Otherwise we would have to redraw images from time to time. Look of the images would change more and more (not same settings, resolution, OS what so ever, different authors). Many of the options are not self-explanatory so the manual should help the reader. These images also help documenting how things look like.

Many things need not changed. In fact txs works quite good out of the box. So one could ask why chapter 1 starts with _Before using TeXstudio, you should configure the editor and latex related commands via the "Configure TeXstudio" command_. It seems to me that changing the commands is quite special. I always asked myself why a manual should start with a large chapter about configuration. Was not my decision, I kept is as it was.

I've seen that the log text is wrong, but I don't want to mix to many different actions in one PR. I don't see which comment is confusing. If you start reading introduction to chapter 1 things should be clear. After that there is an easy way to jump where you like. So we could even move these parts further back. In addition I could add a short note to each caption explaining the fact.

Not sure why markdown is relevant here, but I still have a current LaTeX version.

sunderme:
in context of html, markdown is convenient.
I will probably switch to markdown the coming days and then reorder at least the start of the document, moving configuration to the back.
pandoc conversion seems to work pretty well, it can also generate the numbering/toc.
latex would in principle also work, but markdown easily allows including of html code when desired and the target *is* html and not pdf.

octaeder:
changes concerning blue markings:
- the images for groups that are only available in advanced mode no longer have a blue marking at all (these had a big one that surrounded all options). Changed this because in this case the section titles already state "(Advanced Options only)" (be aware that this addition had never been added in the contents).
- added to the captions "(options marked blue n.a. in standard mode)" as an explanation. This is not done for groups not having advanced options or having advanced options only (s. above).

Since you plan to move chapter 1 to the back I see most of your concerns resolved and we could keep the enhancements made in this PR.

dbitouze:
> Maybe the time has come to switch to markdown. Any input from you @dbitouze ?

A nice option would be for the manual to be generated thanks to the documentation generator [Sphinx-doc](https://www.sphinx-doc.org/en/master/). The input files can be [reStructuredText](https://docutils.sourceforge.io/rst.html) (OK, not very handy) or Markdown and, thanks to Pandoc, probably many others (including LaTeX). Sphinx can generate a series of HTML files, a PDF file (via LaTeX), man pages and much more.

An example of such a documentation is [the one of Furo](https://pradyunsg.me/furo/), precisely a theme for Sphinx-doc.

octaeder:
and who's going to work on the content? I think my solution here is quite acceptable and a good base to continue working on the manual.

sunderme:
This will not be merged, primarily as the manual is now written in markdown.
But also because images and especially marks on images should enhance the information flow.
Anything which is just visible when the user opens the program is of little value, marks to focus and help to find the important point are helpful.
Marking all the differences between advanced/default config mode is something for a specification but not for a manual.

PR #2609, opened 2022-09-24
fix alignement of option in insert graphic wizard

### LINKED ISSUE #2538: Extra spacer in Graphics Wizard
### Environment
- TeXstudio: 4.3.1 (git 4.3.0-17-gd6f13e0)
- Qt: Using Qt Version 5.11.1, compiled with Qt 5.11.1 R
- OS: Linux Mint

### Description
In the Insert Graphics Wizard I found that there is an extra spacer in front of the Center Horizontally checkbox, as we can see in the image below.

Seleção_256

I'm not sure if that alignment is the desired one but in my opinion, removing that spacer would look better, as below.

Seleção_257

octaeder:
This PR fixes #2538.

Change: Removed spacer in front of the checkbox of option "Center Horizontally":

before:
image

after:
image

Note:
A center env is used when only option Center Horizontally is set.
Are both options set (default) then \centering is used within a figure env.

### DIFF
--- a/src/insertgraphics.ui
+++ b/src/insertgraphics.ui
@@ -190,22 +190,6 @@
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout_2">
      <item>
-      <spacer name="horizontalSpacer_2">
-       <property name="orientation">
-        <enum>Qt::Horizontal</enum>
-       </property>
-       <property name="sizeType">
-        <enum>QSizePolicy::Fixed</enum>
-       </property>
-       <property name="sizeHint" stdset="0">
-        <size>
-         <width>8</width>
-         <height>20</height>
-        </size>
-       </property>
-      </spacer>
-     </item>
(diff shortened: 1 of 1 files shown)


sunderme:
to be clear, this is not a fix but a different design choice from what the original author made.
I will accept this as it is minor enough.

octaeder:
You said this change was from 2011. But since we don't know the author I couldn't trie to ask him if this change is ok. I traced back the file history in github but the history ended 2017 (a large change from Tim), so I could not find the change. But I have seen a change stating something like errors from qt in the xml. In this case this would be a fix. Maybe this is not the case here. Another argument why I changed this: A short look at the image reveals that indentation is used for grouping things. Ex:
- Caption
  - short
  - long
 - Label
The issue of one user shows that it is not clear what this indentation should mean. The GUI looks rather inconsistent at that point.

PR #2612, opened 2022-09-28
new cwls

mbertucci47:
New cwls for `chordbox`, `class-udes-genie-these`, `dirtytalk`, `jobname-suffix`, `neo-euler`, `nestquot`, `onedash`, and `quoted`.


sunderme:
thanks

PR #2617, opened 2022-10-01
usermanual typos: changed "an" to "a"

octaeder:
Several cases where a word starting with a consonant follows the word "an".


sunderme:
thanks

PR #2618, opened 2022-10-01
usermanual typos: changed "a" to "an"

octaeder:
Some cases where a word starting with a vowel follows the word "a".


sunderme:
thanks

Issue #2608, opened 2022-09-24
Overhauling user manual


sunderme:
The user manual seems to be more and more outdated.
I have started to convert the manual to markdown and reordered it. (https://github.com/texstudio-org/texstudio/blob/master/utilities/manual/usermanual_en.md)
Help is welcome, especially for the parts which are annotated **todo** .
Target audience are users who are getting started and are looking for new features or trying to solve issues.

sunderme:
there are a few points i want to achieve and maybe someone can help.
* TOC should be beside the text in html, similar to https://pandoc.org/MANUAL.html
* tooltips on top of images would be great to allow context sensitive information.

octaeder:
destroyed my bookmark chapter, please fix:
![image](https://user-images.githubusercontent.com/102688820/192116413-cd8ac816-d938-4752-b204-e1b0e0d5bc71.png)

sunderme:
https://github.com/texstudio-org/texstudio/pull/2607#issuecomment-1257010002 , furo looks interesting. 
For now the usermanual_en.md is written as markdown. Let's see what is the best way to get a html out of it.

octaeder:
your images look quite oversized. Ex. quickstartwizard is 1050 pixels wide, more than half of a FullHD display. Image info shows 144dpi. It should be rendered at about 96dpi (67%).

dbitouze:
> [#2607 (comment)](https://github.com/texstudio-org/texstudio/pull/2607#issuecomment-1257010002) , furo looks interesting. For now the usermanual_en.md is written as markdown. Let's see what is the best way to get a html out of it.

Here is a proof of concept with (a minimal configuration of) `Sphinx-doc` as documentation generator and  `usermanual_en.md` as source file (as it is: could be improved):

- for the HTML output (with `Furo` as theme): https://dbitouze.gitlab.io/texstudio-manual/usermanual_en.html
- for the PDF output: [texstudio.pdf](https://github.com/texstudio-org/texstudio/files/9641733/texstudio.pdf)

dbitouze:
> `usermanual_en.md` as source file (as it is: could be improved)

I meant, from the `Sphinx-doc` point of view.

I also forgot to give the link to the source: https://gitlab.com/dbitouze/texstudio-manual

sunderme:
I tried to use sphinx here. Apparently it is a bit tricky on ubuntu (20.04) as the extensions are not present and it failed.
[...]

PR #2619, opened 2022-10-01
usermanual manipulating tables w. multicursors

octaeder:
Ctrl+Shift instead of Ctrl+Alt, s. [#2608 comment](https://github.com/texstudio-org/texstudio/issues/2608#issuecomment-1263527690) and [block cursors](https://texstudio-org.github.io/editing.html#block-cursor)

### DIFF
--- a/utilities/manual/build/html/editing.html
+++ b/utilities/manual/build/html/editing.html
@@ -458,7 +458,7 @@ current row. If already a command <code class="docutils literal notranslate"><sp
 command is placed.</p></li>
 <li><p>Remodel the table after a template,see <a class="reference internal" href="#using-table-templates"><span class="std std-doc">below</span></a>.</p></li>
 </ul>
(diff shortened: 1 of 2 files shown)


sunderme:
`ctr+alt` is also correct, especially with the instruction to drag the mouse. ctrl+shift makes a box select, ctrl+alt add cursors. As it is described here, I assume ctrl+alt is the intention of the original author.

octaeder:
> `ctr+alt` is also correct, especially with the instruction to drag the mouse. ctrl+shift makes a box select, ctrl+alt add cursors. As it is described here, I assume ctrl+alt is the intention of the original author.

You can use ctrl+alt and click for each cursor you want to create. But dragging the mouse with ctrl+alt is odd. All cursors are in the same column (maybe except the current cursor before clicking). Probably something is selected in one row when you try to move side wards. Please try yourself.

sunderme:
there are three ways of generating the block cursor, all are described in the last section, one in the table section, that should suffice.

Issue #2629, opened 2022-10-09
For pdf-viewer move/add "Fit to Text Width"


octaeder:
### Current behavior:
While using pdf-viewer I couldn't remember the number n in the shortcut Ctrl-n for "Fit to Window". So I looked at the toolbar and counted icons from left to right:
![grafik](https://user-images.githubusercontent.com/102688820/194734333-d54b1510-71ef-4446-9547-31aae51f34b8.png)

So I assumed n=4, but actually Ctrl+4 is "Fit to Text Width" as can be seen from the view menu:
![grafik](https://user-images.githubusercontent.com/102688820/194734370-06060714-3f3e-43a3-ab3f-efdccc458e43.png)

Thus I would like to interchange the two entries in the menu and the two icons in the toolbar for actions "Fit to Window" and "Fit to Text Width". This would give consistent ordering with respect to shortcuts and avoids changing shortcuts.

Additionally the context menu of the pdf doesn't show action "Fit To Text Width" at all:
![grafik](https://user-images.githubusercontent.com/102688820/194734526-9690c57a-581d-4a76-b70c-5cf84f094153.png)

### Expected (images from a change I can provide as PR):
![grafik](https://user-images.githubusercontent.com/102688820/194734601-5b9cc928-c273-49e5-91e2-1efbb3628ac8.png) ___ ![grafik](https://user-images.githubusercontent.com/102688820/194734621-508c3461-9522-49d6-82d1-81337d2c30d2.png) ___ ![grafik](https://user-images.githubusercontent.com/102688820/194734646-91d2a769-4e6a-4d2e-b65d-60dfe58ae7f4.png)

sunderme:
don't change the order of the menu.
You can adapt the shortcuts if you deem it indispensable.

octaeder:
you are not talking about the missing "Fit to Text Width" in the context menu (3rd image). Not add it? (then of course in original order of the view menu I suppose)
you always have good reasons. can you let me know why not interchange these two items?

sunderme:
1) the order seems logical as is, the two width variants next to each other. Splitting it looks odd. (and changing menu position will challenge muscle memory if anyone is using it)
2) nobody has missed the option in the context menu. Unless you intend to use it, changes for a user base of 0 is a waste of time.
You can question if these options need to be in the context menu at all as they are more easily available in the toolbar, but there is no need to change that either.

octaeder:
[...]

PR #2631, opened 2022-10-10
pdf menu entry "Fit Text to Window"

octaeder:
this is a small cosmetic change discussed in #2629


sunderme:
I don't accept changing the order in the toolbar.

Issue #2637, opened 2022-10-15
New Icons & Icon Themes


geolta:
### Environment
- TeXstudio: 4.3.1
- Qt: 5.15.6
- OS: Fedora
- TeX distribution: TexLive

### Description
The icons currently used look quite old and have no really consistent style. The designers of LibreOffice created the incredible [Colibre Icon Theme](https://github.com/rizmut/libreoffice-style-colibre), I partially adapted for TexStudio for testing. I haven't changed all icons yet but here is a screenshot of what I tried in comparison to the old icons:
![grafik](https://user-images.githubusercontent.com/115890454/196006176-914e71e2-8271-4ff4-b15a-38e3e60fd31a.png)
![grafik](https://user-images.githubusercontent.com/115890454/196006183-0f4b1a09-9aa6-4942-ae33-92937f7821b7.png)

Not everything is perfect and there are still a few things to change. But what do you think about the idea of updating the icons in general?

Since different people always favour different styles, I would additionally suggest adding the option for icon themes so that everyone can choose the style he likes most.

sunderme:
I agree that a icon theme selection would be great.
If you provide an icon set, I can make it work.

geolta:
Thank you for your positive feedback!
I'm now working on the implementation of icon themes in the configure dialog as well as on further improving the Colibre icon theme for TXS.

Until now, the icon theme was combined with the setting for the color scheme, I separated this to offer the possibility to create icon themes without theming the entire application (but the detection of dark mode still works).

There are a few hard referenced icons that are not loaded via one of the `getRealIcon` functions in `src/utilsSystem.cpp`. Should we ignore them or try to integrate them in the theming as well? (They look like the person who wrote the code just forgot to use the functions but I'm not sure since I'm no that deep into the code...)

sunderme:
As you like. If there are new icons, i can probably integrate them, but it will take a bit of time.

PR #2645, opened 2022-10-25
Implementation of Icon Themes & New Colibre Icon Theme

geolta:
Referring to the issue #2637, the following main changes were done:

## Icon Theme Selection
A new setting, `Icon Theme` is now available in the preferences. A few small changes in the code were done to remove hard referenced icons (but not all spots are corrected, e.g. it is not easily possible to adapt the icons used in UI files).

But I didn't improve the current handling of icons in general like different icons for different resolutions, etc.

## New Colibre Icon Theme
The *Colibre* icon theme from LibreOffice (https://github.com/rizmut/libreoffice-style-colibre) was adapted for TXS and added to the icon theme selector, a dark mode is included.

grafik

## Changed default Main Toolbar Scaling
The default icon size of the main toolbar was changed from 22 to 24px. This has the background, that 24px is the standard size of icons; for both the classic and the new icons this leads to a significantly increased sharpness:

grafik


sunderme:
thanks

sunderme:
a name for the about dialog ?

geolta:
You can just put my username "geolta" into the about dialog, thanks :)

tmelorc:
@geolta Please, could you fix the icon for XY? It should be like TZ or AS.

geolta:
@tmelorc Do you mean this one?
grafik

octaeder:
I think that's the icon tmelorc was talking about. But the Beamer icon has also changed from BM, and TZ formerly was TI. So wouldn't it be consequent to discuss if this would have to be turned back either? I don't know... Maybe it's a matter of relearning.

tmelorc:
> @tmelorc Do you mean this one? grafik

Yes. This icon comes from the Icon Theme but I think it is not consistent with other icons for drawing packages, like AS or TZ. So, XY is much more related.

tmelorc:
> I think that's the icon tmelorc was talking about. But the Beamer icon has also changed from BM, and TZ formerly was TI. So wouldn't it be consequent to discuss if this would have to be turned back either? I don't know... Maybe it's a matter of relearning.

Since Beamer is for slides creation, I think that the presentation icon is consistent. But the new icon for XYpic does not reflect the package (in my opinion it is more related to some axis or nodes in bounding boxes, like on Inkscape, for example). Since other drawing packages (like TikZ) have the simple two letters icon, XY is still preferred, in my point of view.

geolta:
> Since Beamer is for slides creation, I think that the presentation icon is consistent. But the new icon for XYpic does not reflect the package (in my opinion it is more related to some axis or nodes in bounding boxes, like on Inkscape, for example). Since other drawing packages (like TikZ) have the simple two letters icon, XY is still preferred, in my point of view.

I think this makes sense. Of course, it would be best to have icons with more than just two letters for all drawing packages but I have no good idea for them... So I think, I will change the icon.

tvaillantdeguelis:
Colibre is a great icon theme! 

However, I think the "track change" icons need some improvements:
Colibre

They looked much more meaningful with the Oxygen Modern icon theme:
Oxygen Modern

geolta:
Yeah, that is true (for the replacement icon in particular); the ones for adding and removing are ok in my opinion. When I was creating these icons I was also thinking about a better alternative but had no good idea, because the "old" ones are also not perfect. Do you have an idea?

geolta:
> > Since Beamer is for slides creation, I think that the presentation icon is consistent. But the new icon for XYpic does not reflect the package (in my opinion it is more related to some axis or nodes in bounding boxes, like on Inkscape, for example). Since other drawing packages (like TikZ) have the simple two letters icon, XY is still preferred, in my point of view.
> 
> I think this makes sense. Of course, it would be best to have icons with more than just two letters for all drawing packages but I have no good idea for them... So I think, I will change the icon.

Oh, I see that I forgot to do this, I hope I can do it this weekend.

tvaillantdeguelis:
> Yeah, that is true (for the replacement icon in particular); the ones for adding and removing are ok in my opinion. When I was creating these icons I was also thinking about a better alternative but had no good idea, because the "old" ones are also not perfect. Do you have an idea?

I'm colorblind. All new icons (expected the second one) look the same to me. I think the shape of the icon should tell us what it is intended to do.

geolta:
Ok, that's something I was not aware of, but TXS should be a software for everyone, you're absolutely right!

What do you think about these two suggestions?

Version | Add | Remove | Replace | Alert
---|---|---|---|---
No. 1 | add | remove | replace | highlight
No. 2 | add | remove | replace | highlight

* Don't know why they are displayed in different sizes

tvaillantdeguelis:
Thank you for the suggestions! I would recommend to use blue and red instead of green and red. Green and red look very similar for most colorblind people. If you prefer to keep green and red, please consider using different lightness for those two colors. For example #07EF6F and #EF0716.

I've a personal preference for version No. 2.  I'd suggest to try with a blue "+" and a red "-". I'd also try with much bolder "+" and "-" symbols. For the _replace_ icon, could it make sense to use the exact same icons used for _add_ and _remove_ since, I guess, _replace_ is simply a combination of both?

PR #2665, opened 2022-11-08
new+updated cwls

mbertucci47:
New cwls for `euler-math`, `luacas`, `luaoptions`, and `songproj`.
Updated cwls for `ProfCollege`, `ProfLycee`, `mathastext`, and other small updates.


sunderme:
thanks

PR #2686, opened 2022-11-19
show Advanced Options notifications only one time

### LINKED ISSUE #2684: show Advanced Options notifications only one time
In the config dialog you can silently disable Advanced Options. Turning back to Advanced Options always needs to confirm this notification:

image

This can be annoying, so I suggest a second option at the bottom of the General options (turned on by default):

image

May I implement this?

octaeder:
This PR closes #2684. The first time a user checks "Show Advanced Options" (in General options) he will be presented a notification about advanced options. Txs already remembers this (s. Config%20Riddled below). But due to this PR txs will never popup this notification again when Config%20Riddles is true.

see ini-file:
```
Interface\Config%20Show%20Advanced%20Options=false
Interface\Config%20Riddled=true
```

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -868,7 +868,7 @@ void ConfigDialog::advancedOptionsToggled(bool on)
 
 void ConfigDialog::advancedOptionsClicked(bool on)
 {
-	if (on) {
+	if (on & !riddled) {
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #2695, opened 2022-11-21
new and updated cwls

mbertucci47:
New cwls for `crossrefenum`, `lgrmath`, and `oz`. Updated cwl for `pixelart` and other small fixes.


sunderme:
thanks

PR #2699, opened 2022-11-24
cwls for tikzlibraries

mbertucci47:
Cwls added for tikz libraries shipped with tikz/pgf and the `tikzmark` library.

New cwl for `mbboard`.

Updated cwls for `ProfLycee`, `tikz`, and other small fixes.


sunderme:
thanks

PR #2714, opened 2022-11-27
Editor action "move all others to other view"

### LINKED ISSUE #2713: Editor: add action "move all others to other view
Today this can only be done in two steps: "move all to other view" and then "move to other view".

grafik

I'll create a PR right now.

octaeder:
This PR closes #2713. Added action can be seen in editor tabs context menu:

grafik

### DIFF
--- a/src/editors.cpp
+++ b/src/editors.cpp
@@ -396,15 +396,18 @@ void Editors::tabBarContextMenu(const QPoint &point)
 	if (!editorUnderCursor) act->setEnabled(false);
 	connect(act, SIGNAL(triggered()), SLOT(moveToOtherTabGroup()));
 
-	act = menu.addAction(tr("Move all to other view"));
-	act->setData(QVariant::fromValue<TxsTabWidget *>(tabGroup));
-	if (!editorUnderCursor) act->setEnabled(false);
-	connect(act, SIGNAL(triggered()), SLOT(moveAllToOtherTabGroup()));
-
-	act = menu.addAction(tr("Move all others to other view"));
-	act->setData(QVariant::fromValue<LatexEditorView *>(editorUnderCursor));
-	if (!editorUnderCursor) act->setEnabled(false);
-	connect(act, SIGNAL(triggered()), SLOT(moveAllOthersToOtherTabGroup()));
+	if (tabGroups[0] == tabWidgetFromEditor(editorUnderCursor)) {
+		act = menu.addAction(tr("Move all others to other view"));
+		act->setData(QVariant::fromValue<LatexEditorView *>(editorUnderCursor));
+		if (!editorUnderCursor) act->setEnabled(false);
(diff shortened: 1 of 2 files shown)


sunderme:
hmmm, I am not a fan of filling menus/context menus with fringe functionality.
"move all others" is bascially "move all" and then "one back".
What can make some sense is "move all others" on the left view and "move all" on the right view.

octaeder:
Open a dozen documents (which may even have lengthy names or shrink window size) in the left view. Open context menu on some tab from the left and choose "move all to other view". How do you bring back this tab? This is likely not possible with two clicks. Reason is that after moving all tabs to the right view the last tab moved will be active. Thus some of the first tabs will not be visible (tabs move out to the left). You will have to scroll to the left or open a document list. And for this task you have to remember the title of the document (sorry is some of the documents are untitled) or its position. Thus you impose some restrictions to the user, consume user's memory and have a varying menu entry (same list position but changing name and function) which is irritating. What do you prefer? Let me know so I can do the changes if necessary.

sunderme:
I stick with my proposal.
move all on the left view is basically never useful, on the right "move all" serves to close the dual view set-up.

octaeder:
Further detail: Only change the menu of the left side (view) to show "Move all others to other view" replacing "Move all to other view", while right side continues to show "Move all to other view":

left view:
grafik

right view:
grafik

src/editors.cpp
```diff
@@ -396,10 +396,18 @@ void Editors::tabBarContextMenu(const QPoint &point)
...
 	connect(act, SIGNAL(triggered()), SLOT(moveToOtherTabGroup()));
 
-	act = menu.addAction(tr("Move all to other view"));
-	act->setData(QVariant::fromValue<TxsTabWidget *>(tabGroup));
-	if (!editorUnderCursor) act->setEnabled(false);
-	connect(act, SIGNAL(triggered()), SLOT(moveAllToOtherTabGroup()));
+	if (tabGroups[0] == tabWidgetFromEditor(editorUnderCursor)) {
+		act = menu.addAction(tr("Move all others to other view"));
+		act->setData(QVariant::fromValue<LatexEditorView *>(editorUnderCursor));
+		if (!editorUnderCursor) act->setEnabled(false);
```
sunderme, line 402: this is dead code (never true)
I would suggest to handle the whole part (move ...) only if editorUnderCursor != nullptr

sunderme: whole part-> ll. 399-410

octaeder: So you don't want to have disabled entries that I see here?
grafik

sunderme: the logic removes all the other entries , see some lines below, so I would stick with that logic.

octaeder: grafik

sunderme:
thanks

Issue #2707, opened 2022-11-27
Make the left and right parts of the Config Dialog resizable with a splitter


octaeder:
Green arrow points to the splitter handle (for a larger version click on the image):

![SplitterInAction](https://user-images.githubusercontent.com/102688820/205197284-170eadfd-7480-48a4-a8de-ec20334b18e4.gif)

I'm working on that...

Thx

sunderme:
the page selection should never collapse, icons only can be minimum width.

PR #2715, opened 2022-11-29
Improvements to config dialog pages layout

octaeder:
This PR mainly addresses following items:

- Implementing a splitter (this solves #2707, where you can find a small animation). For a more pleasing look space in the middle (between page list and page area) is decreased from 36px to 9px, so the splitter handle here can be located easily.
- Resolving that some of the pages use a different right margin as the majority does, and that the language checking page is in addition moved a little bit downwards (s. images).

The list below states changes done in the pages (reducing left margin for all pages is not listed).

Note: The splitter is set such that the list of pages has the same width as it is in the reference image to reduce jitter. The first image (with the arrow) in #2707 shows the initial width when opening the config dialog.

1. General: right margin of page (s. animation)
General
2. Commands:
3. Build: right margin of page, no scroll area for three lines at bottom ("Additional Search Paths") (s. animation)
Build
4. Shortcuts:
5. Menus:
6. Toolbars:
7. GUI Scaling: fit group box, changed group box title (s. animation)
GUI-Scaling
8. Editor: moved checkbox "folding" below "Show Line Numbers" (s. animation)
Editor
9. Adv. Editor:
10. Syntax Highlighting:
11. Completion:
12. Language Checking: upper, and right margin of page (s. animation)
SpellCheck
13. Preview: moved two checkboxes to the left, changed group box title (s. animation)
Preview
14. Internal PDF Viewer:
15. SVN/GIT: new group box (s. animation)
SVN-GIT

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -13,7 +13,7 @@
   <property name="windowTitle">
    <string>Configure TeXstudio</string>
   </property>
-  <layout class="QGridLayout">
+ <layout class="QGridLayout">
    <property name="leftMargin">
     <number>9</number>
    </property>
@@ -32,7 +32,25 @@
    <property name="verticalSpacing">
     <number>6</number>
    </property>
-   <item row="0" column="0">
+<item row="0" column="0">
+<widget class="QSplitter" name="mainSplitter">
+ <property name="childrenCollapsible">
+  <bool>
+   true
+  </bool>
+ </property>
+ <property name="handleWidth">
+  <number>9</number>
+ </property>
+ <widget class="QWidget" name="leftPart">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Expanding" vsizetype="MinimumExpanding">
+       <horstretch>2</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+  <layout class="QVBoxLayout">
+   <item>
     <widget class="QListWidget" name="contentsWidget">
      <property name="sizePolicy">
       <sizepolicy hsizetype="Preferred" vsizetype="Expanding">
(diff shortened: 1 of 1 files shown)


sunderme:
Commands: top/bottom looks cut.
grafik
build (&general): left part is to wide and stays wide on cost of right. Is different from before.
grafik

octaeder:
That's what I see:

grafik

What style settings do you use? Which Qt? For reference: TeXstudio 4.4.2 (git 4.4.1-3-ga9d878dca), Using Qt Version 6.4.1, compiled with Qt 6.4.1 R.

The width of the select page part is the same as I see from new installed 4.4.1 Win10. Be aware that there is currently no splitter (is expected to be done with the other issue). Reference (old) images show current state of 4.4.1, so no difference I can see.

octaeder:
Here 4.4.2 (mine) is on top of 4.4.1: Only difference is the smaller gap between page list and page, so page list is wider:

grafik

Maybe I can figure out which value influences the width of page list.

sunderme:
have you tried to resize the dialog and see what happens ?

octaeder:
yes, but no bad observations. Can it be that a cpp program defines size of the dialog and of the page list?

octaeder:
Following image shows the width when the list part starts shrinking on further reduction of the dialog width:

grafik

This new version can reach a smaller width because the part right to the list is smaller (less spacing). In fact the old space in the middle consists of two column separators of width 18px each. This is possible because there is a second but empty column of width 0 (well it contains a vertical spacer of width 0 otherwise this would not be possible). I removed that empty column (and thus one column separator) and reduced the width for column separators to 12px. This is the new space in the middle.

sunderme:
you image above shows the issue. right pane is reduced in size, starting with scrolling, while the left pane is still at max width.
You can set scaling ratios between different elements and my guess is that you removed the definition for the right-hand side.
This all formular stuff, no cpp involved.

octaeder:
I reworked my first post incl. images and finished work.

sunderme:
looks good.
Thanks.

PR #2720, opened 2022-12-03
insert inline math mode as $$ from menu

### LINKED ISSUE #2719: line math mode
Hello. I want to suggest that the line math mode $ ... $ be written $$ instead of $ $ (no space between the first $ and the second $).

I also want to suggest that the shortcut Ctrl+Tab be available to activate the line math mode option.

octaeder:
this PR resolves #2719 

Notes to be considered:
- \[  ]\ and \( \) are inserted with two spaces
- Maybe the menu should show $$ instead of $...$

### DIFF
--- a/uiconfig.xml
+++ b/uiconfig.xml
@@ -227,7 +227,7 @@
 
 <menu id="main/math" text="&amp;Math">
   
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #2722, opened 2022-12-04
add menu LaTeX/Horizontal Spacing

octaeder:
This PR closes #2687.

Additionally:
- added \vfill to menu Vertical Spacing

Question: Where are the info texts displayed or are they only documentation? Hoped they would be tooltips.

PR #2730, opened 2022-12-05
fix newline in central toolbar

octaeder:
name had been changed while introducing horizontal spacing menu in uiconfig.xml (#2722)


sunderme:
thanks

PR #2737, opened 2022-12-06
Localize dateTime display for last update check

### LINKED ISSUE #2731: Localize DateTime display for last update
The General config page shows date and time of last update check.

not typically german/english:
grafik
grafik

look at this:
grafik
grafik

So I suggest following change:
`` QLocale::QLocale().toString(lastCheck, QLocale::ShortFormat)`` 
instead of
``lastCheck.toString("d.M.yyyy hh:mm")``

Shall I?

octaeder:
This PR closes #2731.


sunderme:
thanks

PR #2741, opened 2022-12-07
fix min. height for config dialog

octaeder:
This PR addresses some new and old issues regarding the minimum height achievable for the config dialog.
I observed that for the current dev version the height of the config dialog is almost fixed:

smallShrinkAmount

While investigating I found that current version 4.2.1 shows some strange behaviors. First is that all pages have the same minimum height except Commands page having a larger one. But you will only notice if you try to increase the height of the window because then heigth makes a step up:

CommandsNeedMoreHeight

Second the Editor page schrinks the height of some widgets when height is minimal, but this is less obvious:

EditorShrinksWidgets

Now Editor page has a scrollbar so this no longer happens. And now all pages have the same minimum height. Both can be seen in the following images:

EditorScrollbar

BuildHeightFixed

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -7,13 +7,13 @@
     <x>0</x>
     <y>0</y>
     <width>1116</width>
-    <height>674</height>
+    <height>300</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #2747, opened 2022-12-09
Narrow language picker

octaeder:
For a quick overview of this PR concerning config dialog you may have a look at the images/animated gifs below (showing before/after, style set to Fusion, Classic scheme).

Often there is a label to the left that is followed by a widget allowing corresponding settings. The distance between the both shouldn't be too large because the eye can follow the line easier and recognize that both information are related to each other.  As an example have a look at Build page below. There we see a column of some check boxes to the right. The combo box and the spin box below are aligned left to the check boxes which gives us the impression that they are related. So we might not take care of the two labels far to the left.

Further there are cases where we find large distances leaving big spaces (as an ex. s. Preview page below). It happens often that these spaces shrink or expand when width of the config window is resized. While moving objects can be irritating in it self it also can lead to loss of vertical alignment of widgets. As an example try with Language Checking page where it seems that the widgets of the first group box are left aligned with those of the second group box. But they are no longer when you change the width.

Last but not least there are horizontal sequences of more than two check boxes. In general there should be a noticeable space between the text and the check box (the square to be checked) of the next option. This space should be slightly larger than is the space between the check box and the following text. Thus the options can be distinguished easily without being scattered.

Many of the combo boxes expand to the right so they are very wide. As an example take the language picker of General page. When you click on it then a huge panel opens that takes about three quarters of the page width while the text list takes only a small amount of it. As you can see in the images below I reduced the width and it looks fine now. Even worse is the font combo box that stretches far to the right. But I couldn't fix that.

I want to note one drawback of oversized combo (and spin) boxes. When scrolling a page (with the wheel) it becomes likely that such a combo box moves under the mouse pointer. But then the page stops scrolling and the combo box is being scrolled. Thus you might change settings without notice. If the pointer is at the lower edge of a scrollable area then you can't even see that a combo box might rise up (same at top). Even if you use the wheel while the pointer is over the scrollbar you may not notice that your hand slowly moves the pointer into the page. But saying all that I have no nice solution for that.

01_General

03_Build

04_Shortcuts

07_GuiScaling

12_SpellCheck

13_Preview

14_PdfViewer


octaeder:
I'll try with a new PR, the merge commits are to much.

PR #2748, opened 2022-12-09
Add a workflow to build for Windows ARM64

minnyres:
Use MSVC to cross compile on Windows x64 host. Official prebuilt Qt library is used. Poppler is built from source and its dependencies are built via vcpkg. TexTablet is rebuilt with .Net 4.8 targeting ARM64.


minnyres:
the arm64 build works well on my tablet PC

屏幕截图 2022-12-09 225455

sunderme:
I have noticed a few points:
1) msys is installed but not really used for anything
2) the result should be an installer (nsis) and not a zip
3) as all depencies are built next to the actual binary, this is not suited as continous deployement. Maybe for releases but not for every push.

PR #2750, opened 2022-12-10
narrow General config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page General of Config Dialog:

01_General

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -192,7 +192,7 @@
                     </property>
                    </widget>
                   </item>
-                  <item row="10" column="1" colspan="1">
+                  <item row="10" column="1" colspan="3">
                    <widget class="QComboBox" name="comboBoxLanguage">
                     <property name="sizePolicy">
                      <sizepolicy hsizetype="Expanding" vsizetype="Preferred">
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -164,17 +164,7 @@
                  <property name="title">
                   <string>Appearance</string>
                  </property>
-                 <layout class="QGridLayout">
-                  <item row="4" column="1">
-                   <widget class="QCheckBox" name="checkBoxUseTexmakerPalette">
-                    <property name="text">
-                     <string>Ignore Most System Colors</string>
-                    </property>
-                    <property name="advancedOption" stdset="0">
-                     <bool>true</bool>
-                    </property>
-                   </widget>
-                  </item>
+                 <layout class="QGridLayout" columnstretch="0,0,1">
                   <item row="0" column="1" colspan="3">
                    <widget class="QComboBox" name="comboBoxInterfaceStyle">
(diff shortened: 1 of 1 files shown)


sunderme:
all drop-down above are over the complete width, why should the language selector stick out ?

sunderme:
the shift left of the widgets is okay (still not necessary)

octaeder:
did you read #2747 ?

sunderme:
please, don't refer to a bunch of text and expect me to pick up what part of that you are referring to.
I don't read anything there that states that there is a deeper necessity to keep that drop-down small.

octaeder:
Many of the combo boxes expand to the right so they are very wide. As an example take the language picker of General page. When you click on it then a huge panel opens that takes about three quarters of the page width while the text list takes only a small amount of it.

octaeder:
Even worse is the font combo box that stretches far to the right.

octaeder:
but the latter I did'n fix

sunderme:
yes, that is basically true for all combo-boxes in the in the config.
but why is that bad ? odd maybe.

octaeder:
1. not all combos are full width, so this is not a one exception (s. edit, adv. edit). 2. in most cases the presented list is quite short. So the area is not so large. But the language list combo is really large (with fusion you see the full list). And the fonts display even expands over the right edge of the window.

image

image

This not nice at all.

octaeder:
and if you look close you will notice that the appereance of Font combo boxes (General and Edit) looks different from all others. That's wired.

Why not full width for combos in general? As I pointed out in the first PR there are scrolling issues: When scrolling a page (with the wheel) it becomes likely that such a combo box moves under the mouse pointer. But then the page stops scrolling and the combo box is being scrolled. Thus you might change settings without notice.

octaeder:
even so I don't plan to shrink all combos.

octaeder:
The large extent of the combos could be bad for the user because he has to focus far to the right for moving the pointer to where he can open the list, and then back far the left to read the entries of the list. Much distracting eye movement. But this is not such a problem but only if you are an experienced user who knows that this is not really necessary. One can click anywhere in the combo an the list is opened (at least for fusion and default style). But this is not true for the font combos. So in most cases I agree that the combo can extend to the right edge, but sometimes it should not.

sunderme:
I would much prefer to actually address weaknesses of txs and not pure cosmetics.
the config dialog may be better with small screens and possibly with large screen/fullscreen dialog (and no, I am not talking about moving stuff a few pixels, look at the commands pane).

The fontcombobox seems special, but that is probably a qt issue.
So, there is no argument to just shrink one combobox. The scrolling issue does not go away by moving comboboxes around, even if you do it with all of them.

So, acceptable change: shift the left edge of the combobox.

sunderme:
thanks

PR #2751, opened 2022-12-10
narrow Build config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page Build of Config Dialog:

03_Build


sunderme:
not necessary, not harming, so I accept it.

PR #2752, opened 2022-12-10
narrow Shortcuts config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page Shortcuts of Config Dialog:

04_Shortcuts

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -1098,6 +1098,9 @@ Then you can select a new shortcut by one of the following ways:
               <string>Close Element (Esc)</string>
              </property>
              <layout class="QHBoxLayout" name="horizontalLayout_5">
+              <property name="spacing">
+               <number>9</number>
+              </property>
               <item>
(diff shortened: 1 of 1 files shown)


sunderme:
I am not exactly sure, why it should be "narrow".
Basically every where in that dialog a grid based placement is used for check boxes, why not here ?

octaeder:
It is a grid. But now with a defined spacing that is independent of the window geometry. Open Config window maximized on a 24" monitor (not a smart phone) and the distance gets so large that you don't see them as a group of options. From my point of view this is already the case when the window opens in default size.

sunderme:
I would much prefer to actually address weaknesses of txs and not pure cosmetics.
the config dialog may be better with small screens and possibly with large screen/fullscreen dialog (and no, I am not talking about moving stuff a few pixels, look at the commands pane).

This is a design decision change.
The previous author laid it out this way, I don't see a compelling reason to override that decision.

PR #2753, opened 2022-12-10
narrow GUI Scaling config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page GUI Scaling of Config Dialog:

07_GuiScaling


sunderme:
thanks

PR #2754, opened 2022-12-10
narrow Spell check config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page Language Checking of Config Dialog:

12_SpellCheck

Loss of vertical alignment of widgets where it seems that the widgets of the first group box are left aligned with those of the second group box. But they are no longer when you change the width. So moving left to the labels as all others are.


sunderme:
thanks

PR #2755, opened 2022-12-10
narrow Preview config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page Preview of Config Dialog:

13_Preview

Why using a grid that splits the pane in two halfs?


sunderme:
naja

PR #2756, opened 2022-12-10
narrow PDF viewer config

octaeder:
This PR is a cut down of PR #2747 where you can find full details. Change affects page Internal PDF Viewer of Config Dialog:

14_PdfViewer

No need to align with checkboxes.


sunderme:
again, completely unnecessary.

PR #2757, opened 2022-12-10
Update Level as an advanced Option

octaeder:
Make Update Level (combobox) an advanced option.

Further suggestions:
1. Normalize distance between checkbox (automatic checking) and spinbox (cycle time) --compare with distance between date/time and check button in current version, 3rd image-- and so it also looks more like one text
2. Increase max. number in spinbox from 99 to 999 (this allows checking once a year)
3. More compact arrangement, of course combobox could also extend to the right

Without advanced options:
image

Advanced options checked (combobox and spinbox have same extent to the right independent of language translation):
image

Compare to current:
image

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -483,111 +483,175 @@
                  <property name="title">
                   <string>Update</string>
                  </property>
-                 <layout class="QGridLayout" name="gridLayout03">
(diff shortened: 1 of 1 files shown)


sunderme:
no
Beta testing is already too low. Everybody should be aware of that possibility.

Issue #2672, opened 2022-11-13
Drag-Dropping of Tables from LibreOffice Calc / Excel


geolta:
Creating tables in LaTeX is always a bit annoying if it is based on table already existing in a spreadsheet application. Similar to drag-dropping images I would suggest to add drag-dropping and copy-pasting tables to TXS. 

I could implement this.

sunderme:
there is some functionality with the tabular wizard, but it is not very extensive

geolta:
Yeah, I'm more thinking about just converting the table to LaTeX format or in other words separate cells by `&` and newlines by `\\` without starting the wizard. Maybe also putting this into a tabular environment by default.

PR #2761, opened 2022-12-11
#2672 – Add support for drag-dropping & copy-pasting spreadsheets from LO Calc

geolta:
This implements the functionality to drag-drop and copy-paste fragments of spreadsheets from LibreOffice Calc to TexStudio. Support for Excel is currently not supported, as well as complex formatting.


sunderme:
I`d assume that `\begin{tabular}{cols}...\end{tabular}` is also automatically inserted ?

geolta:
No, the tabular environment is not inserted. I did it this way to make it easier to combine multiple table fragments as it is usually done (at least in my experience where you join several parts of a table).

sunderme:
In my opinion, that is probably a bit unexpected for the user.
I will look into it.

geolta:
Hi, do you already had a look into it?

Thanks :)

sunderme:
My idea would be to check if content into an existing table and act accordingly.
I have not yet started to implement that.
For now, I will prepare the next release.

sunderme:
implemented for paste 
if no table is already present, call table waizard

PR #2765, opened 2022-12-12
add completion support for pgfplots and tcolorbox libraries

mbertucci47:
Makes completion possible for pgfplots and tcolorbox libraries, so `\usepgfplotslibrary{clickable}` will load `pgfplotslibraryclickable.cwl` and `\tcbuselibrary{minted}` will load `tcolorboxlibraryminted.cwl`.

I had trouble building (perhaps the [Ubuntu instructions](https://github.com/texstudio-org/texstudio/wiki/Compiling#ubuntu) could be updated) so wasn't able to test, but based on the analogous [commit](https://github.com/texstudio-org/texstudio/commit/74b4042daef8e2df5614354748a15a16b1671baf) for tikzlibraries I think it should be okay.

### DIFF
--- a/completion/pgfplots.cwl
+++ b/completion/pgfplots.cwl
@@ -789,7 +789,7 @@ enable tick line clipping#true,false
 # no new user commands
 
 ## pgfplotslibrary.code.tex
-\usepgfplotslibrary{library list%keyvals}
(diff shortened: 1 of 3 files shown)


sunderme:
thanks

sunderme:
ubuntu instructions should still work.
What fails ?

PR #2772, opened 2022-12-14
updated cwls

mbertucci47:
New cwls for `bibcop`, `diadia`, and `scikgtex`.

Updated several cwls to account for tcolorbox library completion.


sunderme:
thanks

PR #2773, opened 2022-12-15
add completion localization for `tabularray` libraries

mbertucci47:
In line with recently added support for `\usetikzlibrary` and so on, makes `\UseTblrLibrary{amsmath}` load `tabularraylibraryamsmath.cwl`.


sunderme:
thanks

this seems to be about the limit for "special" treatment of these commands. It needs to be generalized after that.

PR #2787, opened 2022-12-19
fix side panel title when changing grid size in config

### LINKED ISSUE #2743: Structure Pane with false title when changing Symbol Grid size
- open side pane and choose Bookmarks view
- go to GUI Scaling in the Config Manager
- Change value of Symbol Grid with the slider (if reset would change the value then reset has the same effect)

The side pane has switched to the Symbols view so life monitoring of symbol's size is possible. But the title is still Bookmarks (s. image). In fact it should be Symbols. This does not change on closing the Config in any way.
Choosing any of the side pane views (incl. Symbols) will result in a correct title.

grafik

Txs 4.4.1, Qt 6, Win 10

octaeder:
This PR fixes #2743. When changing the grid size in the config we first have to check if the symbols view is not hidden in the left panel. If not we can switch to that view and update the side panel title appropriate.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -11059,7 +11059,11 @@ void Texstudio::changeSymbolGridIconSize(int value, bool changePanel)
 	int iconWidth=qRound(value*scale);
 
 	if (changePanel) {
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #2791, opened 2022-12-21
Port txs from Phonon to QtMutlimedia framework

### LINKED ISSUE #2790: Port txs from Phonon to QtMutlimedia framework
As was mentioned in #2784, cmake build stack will not support phonon. Phonon framework was substituted by QtMutlimedia one that present in Qt6.
So, we need to migrate from Phonon to QtMutlimedia.

idamir:
close #2790


idamir:
Cannot reproduce the test fail on a local workstation.

sunderme:
thanks

PR #2796, opened 2022-12-23
manual: update and fix typo

muzimuzhi:
Remarks
- I believe the update is mainly caused by newer versions of Sphinx (5.2.1 to 5.3.0) and Furo (2022.09.15 to 2022.12.07)
- Somehow I find it needs four backslashes in markdown to produce one in html.


sunderme:
thanks

Issue #2798, opened 2022-12-23
Strings in some tooltips are not translated


dbitouze:
### Environment
- TeXstudio: TeXstudio 4.5.1 (git N/A)
- Qt: 5.15.2
- OS: Mageia 8
- TeX distribution: irrelevant

### Expected behavior
All strings are translated.

### Actual behavior
Some strings are not translated.

### How to reproduce
1. Options → Configure TXS → General → Language → fr (French)
2. Close and restart TXS
3. Move the cursor:
   1.  on the central toolbar over:
       -  the “new line” icon
       - the “math environment” icon
   1.  on the top toolbar over the (four) “easyReview” icons
5. The strings which appear in the corresponding tooltips are in English, not in French.

sunderme:
fixed

dbitouze:
Thanks! Wow, 61 new strings to translate :wink: 

For some of them, e.g. “The \title command declares text to be the title.”, I can't figure it out where they appears in TXS. Could you tell me it?

sunderme:
in theory tooltips in the latex menu

dbitouze:
Where is available these theory tooltips? Ah, found: in the bottom of the UI...

BTW, some of them are not coherent, e.g.:

- “The \quad command inserts a horizontal space of 1em (1em being the width of M).”
- “A \quad is a horizontal space with a size equal to the font size.”

octaeder:
The LaTeX reference txs is linking to says:

em, traditionally the width of the capital letter M, is often used for horizontal lengths. This is also often the size of the current font, e.g., a nominal 10pt font will have 1em = 10pt.

So I suggest changing both to

“The \quad command inserts a horizontal space of 1em (1em being the width of M or the font size).”

I think you will never see the tooltip because they need an icon. There is none defined so the txs icon would be used, for example when adding the entries to the central toolbar, which is not useful:

![grafik](https://user-images.githubusercontent.com/102688820/209454048-1b992bd9-5195-4054-885c-5aec13e9a7eb.png)

Even so you may want to tell what other cases you found.

Since I added the first description 3 weeks ago I could do that  (at least for the \quad) if you agree.

sunderme:
"or the font size" is more misleading than helpful.
Just take the first variant: ...(1em being...)

dbitouze:
Other not coherent (homogeneous) strings (“produces ... list” vs “is used to make ...list**s**”):

[...]

PR #2802, opened 2022-12-26
Update uiconfig.xml

octaeder:
see also [comment](https://github.com/texstudio-org/texstudio/issues/2798#issuecomment-1364602156) and following in issue #2798


sunderme:
thanks

PR #2806, opened 2022-12-27
Reduced precision in file time change filter

### LINKED ISSUE #2805: File change watcher filter is too sensitive
### Environment
- TeXstudio: 4.4.1 / 0d9b79b
- Qt: 6.4.1
- OS: Windows 10
- TeX distribution: tex-live

When working on a project that is stored in Nextcloud (possibly also happening with other cloud storage), when one saves a file, one immediately gets the "File Changed" Messagebox. One can either ignore the external change every time or just enable autoreload, which is inconvenient as the undo stack gets lost, even if there was no actual change to the file. I tried to debug why

https://github.com/texstudio-org/texstudio/blob/78ef1555f5bf9aaf0784f66cd82ef983fde898b1/src/qcodeedit/lib/qreliablefilewatch.cpp#L200

doesn't catch, by adding
```cpp
qDebug() << "PreMod:" << it->lastModified << ", NewMod: " << info.lastModified() << ", PreSz: " << it->size << ", PostSz: " << info.size();
```

and it shows
```
PreMod: QDateTime(2022-12-27 18:15:35.637 Mitteleurop�ische Zeit Qt::LocalTime) , NewMod:  QDateTime(2022-12-27 18:15:35.000 Mitteleurop�ische Zeit Qt::LocalTime) , PreSz:  31 , PostSz:  31
```
so apparently, the time gets truncated to the full seconds. Not sure who's to blame for that, but I think it would be good to not assume that the file system has sub-second timestamps. I therefore suggest to modify the filter condition to ignore the sub-second part of the timestamp.

orangelynx:
Changes condition

https://github.com/texstudio-org/texstudio/blob/78ef1555f5bf9aaf0784f66cd82ef983fde898b1/src/qcodeedit/lib/qreliablefilewatch.cpp#L200

to require a file time change greater than 1 (positive) second. This way, issues with cloud storage that may be backed by file systems with lower/higher mtime precision than the local file system are avoided. Fixes #2805. If the new time is older than the cached time, ignore the change, as we are by definition "newer".

### DIFF
--- a/src/qcodeedit/lib/qreliablefilewatch.cpp
+++ b/src/qcodeedit/lib/qreliablefilewatch.cpp
@@ -197,7 +197,9 @@ void QReliableFileWatch::sourceChanged(const QString& filepath)
 
 	qDebug("%s modified.", qPrintable(filepath));
 	QFileInfo info(filepath);
-	if(it->lastModified==info.lastModified() && it->size==info.size()){
+	// (issue #2805) when using e.g. cloud storage, clashes between mtime precision of different file-systems can occur.
+    // assume no more than 1 second precision. If new time is older than cached time, ignore.
+    if(it->lastModified.secsTo(info.lastModified()) < 1 && it->size == info.size()) {
 		qDebug("filtered");
 		return;
 	}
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

for future reference, please use an up to date fork for providing a PR.
This basically inhibits checking out a PR in a sensible way for testing on my side.

PR #2811, opened 2022-12-30
Dark Mode for Symbol Icons

groverlynn:
Add dark mode versions for symbol icons
Also, increase the size of svg files for high dpi screens


sunderme:
thanks

PR #2817, opened 2023-01-02
new cwls

mbertucci47:
New cwls for `gfdl`, `luacomplex`, `luagcd`, `luamaths`, `luaset`, `maze`, and `resmes`.

Updated cwls for `keyvaltable`, `simpleicons`, and other small updates.

Redundant entries for commands incorporated into `colortbl` removed from cwls.


sunderme:
thanks

PR #2818, opened 2023-01-02
Align Symbols at Baseline

groverlynn:
Align symbols at typesetting baseline (same height and same depth) whenever possible (unless the symbol is an icon). So the symbol glyphs are the same as in the pdf in terms of relative font sizes and vertical alignment. Meanwhile correct a couple of wrong glyphs or unicodes.

### DIFF
--- a/src/symbolpanel/symbollistmodel.cpp
+++ b/src/symbolpanel/symbollistmodel.cpp
@@ -85,14 +85,12 @@ SymbolItem loadSymbolFromSvg(QString fileName)
 	nl = root.elementsByTagName("desc");
 	if (!nl.isEmpty()) {
 		QDomNode n = nl.at(0);
-		item.packages = n.toElement().attribute("Packages");
-	}
-	nl = root.elementsByTagName("additionalInfo");
-
-	if (!nl.isEmpty()) {
-		QDomNode n = nl.at(0);
-		item.unicode = n.toElement().attribute("CommandUnicode");
-
+		QString pkg = n.toElement().attribute("Packages");
+		if (!pkg.isEmpty())
+			item.packages = pkg;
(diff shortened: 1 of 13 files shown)


sunderme:
looks good
gensymb-ng binary/gensymb-ng.o should not be part of the PR.

groverlynn:
> gensymb-ng binary/gensymb-ng.o should not be part of the PR.
@sunderme they are now omitted

src/symbolpanel/symbollistmodel.cpp
```diff
@@ -85,14 +85,12 @@ SymbolItem loadSymbolFromSvg(QString fileName)
...
-		item.packages = n.toElement().attribute("Packages");
-	}
-	nl = root.elementsByTagName("additionalInfo");
-
-	if (!nl.isEmpty()) {
-		QDomNode n = nl.at(0);
-		item.unicode = n.toElement().attribute("CommandUnicode");
-
+		QString pkg = n.toElement().attribute("Packages");
+		if (!pkg.isEmpty() && pkg != "")
```
sunderme, line 89: this is a bit redundant code 
isEmpty is by definition the same as `== ""`

sunderme:
let's start over.
github is not well equipped with reviewing 4000 files.
And submitting everything in one large chunk with one commit does not make things easier.
I need to review source code and I try out the svgs.
The svgs seem to be okay.
So, please, one PR for the changed cpp sources (and maybe the xml control files) and another one for the actual symbols.

First remark on source:
symbollistmodel.cpp, lines 138ff: 
use the `QChar::fromUcs4`and fall back on the other solution for qt5. I suspect that the former is a bit more complete.

groverlynn:
Split the PR as requested

sunderme:
thanks

Issue #2814, opened 2022-12-31
OSX: toolbutton with text show wrong text color


sunderme:
With style macOS and color scheme manhattan the labelled toolbars show wrong text color (black instead of white).
Other styles do not show this behavior.
### Environment
- TeXstudio: 4.5.1
- Qt: 6.4.1
- OS: OSX
- TeX distribution: N/A

### Expected behavior
Text color should be white.

### Actual behavior
the toolbutton with text show black text instead of white.
See also #2767

sunderme:
fixed by @groverlynn

Issue #2767, opened 2022-12-12
Toolbar illegible colour


Atcold:
### Environment
- TeXstudio: 4.4.1
- Qt: 6.4.1
- OS: Mac.
- TeX distribution: N/A

### Expected behaviour

The text on the toolbar should be readable (light colour on dark background).
(Also the ‘undo’ and ‘redo’ arrows could use a brighter blue, for visibility purposes.)

### Actual behaviour

<img width="1332" alt="image" src="https://user-images.githubusercontent.com/2119355/207153255-043e3815-2d60-4502-af69-2c598fd9e733.png">

sunderme:
looks rather like dark-mode is not detected.
please provide a complete image of txs.
Also, how was the dark-mode triggered.

Atcold:
My Mac is just set on dark.
This is how the interface looks.

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/2119355/207443543-dc240b52-7424-4ebd-bf39-11c51db10e1b.png">

sunderme:
Have you tried to use another icon theme ?

sunderme:
by the way, what is set in color scheme and style ?

sunderme:
looking at the picture again, txs uses dark-mode, the unexpected part is that the characters in the drop-down (like `\left(` or `part`) are still black.

geolta:
Hi, the screenshot indeed uses the icons for dark mode. The problem is though, that different dark modes use different shades of gray and therefore for some dark modes, the colors are too dark while for others, they are too light.

My original colors were chosen accordingly to what [rizmut](https://github.com/rizmut/libreoffice-style-colibre) the creator of the icon theme, used. For the creation of the dark mode icons, I used a script. Can you send the exact colors you changed, so I can adjust the script (I think, I will add it to the folder of the icons)?

sunderme:
@Atcold have you tried a current dev build to check if the colors are better ? (see https://github.com/texstudio-org/texstudio/actions/workflows/cd.yml )

Atcold:
> Have you tried to use another icon theme?

The icon theme does not affect the text colour.

> by the way, what is set in colour scheme and style?

- macOS
- Modern - dark
- Colibre

> […] different dark modes use different shades of gray

I'm confused. All Macs use the same dark mode.

> and therefore for some dark modes, the colors are too dark while for others, they are too light.

If the background is dark, the font should be bright.
Here it is back.

[...]

PR #2833, opened 2023-01-06
Fix Toolbar Color Bug under MacOS Dark Mode

groverlynn:
This bug (#2814 #2767) was actually fixed long time ago on Qt's original version of Manhattan Style.


sunderme:
thanks

by the way, it would be easier to see changes in the diff view if not arbitrary indentation was changed ...

groverlynn:
> thanks
> 
> by the way, it would be easier to see changes in the diff view if not arbitrary indentation was changed ...

Just tick `Hide Whitespace Changes`. I had to change to consistent tabs/spaces for all these switches in order to find where to apply the patch.

PR #2840, opened 2023-01-10
PDF pinch-zo-zoom fix for macOS

Iksas:
For some years, the pinch-to-zoom gesture in the integrated PDF viewer on macOS has been broken. Apparently the `QGestureEvent` returns incorrect coordinates, which causes the page to move sideways while a pinch gesture is performed. The bug still occurs in the [4.5.1beta2](https://github.com/texstudio-org/texstudio/releases/tag/4.5.1beta2) version, only on macOS.

This PR proposes a workaround for this issue. By only using the `QGestureEvent` to detect pinch gestures, while taking the coordinates from the `QTouchEvent`, correct zoom behavior can be achieved on macOS.

It has been successfully tested with macOS Ventura 13.1 and Qt 6.4.1. Please let me know your thoughts.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -576,6 +576,8 @@ PDFWidget::PDFWidget(bool embedded)
 	, docPages(0)
 	, saveScaleFactor(1.0)
 	, saveScaleOption(kFitWidth)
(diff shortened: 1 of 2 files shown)


sunderme:
looks good

thanks

PR #2842, opened 2023-01-10
updated cwls

mbertucci47:
New cwls for `lazylist` and `Scrabble`. Updated cwls for `ProfLycee`, `functional`, `marvosym`, `pst-flags`, `scripture`, and many other small fixes.


sunderme:
thanks

PR #2845, opened 2023-01-11
Packages Panel in QuickStart Wizard

### LINKED ISSUE #2831: Feature: A Package Panel (Tab) for the QuickStart Wizard
For all: I want to let you know about an almost finished work. If you like tell me which packages should I add to the packages list.

### Introduction

The following image shows the QuickStart Wizard as you can find it in txs. I added some marks for later reference:

grafik

Even so the tab claims "Class Options" there are other things included (marks 1 to 4). The first two marks show commonly used options (if you don't use LuaLaTeX) for packages but which integrate well into the tabs geometry. This is also true for the babel option/package. But the following line with three options (mark 4) are packages not options. And the first of them (AMS) in fact represents two packages, but which packages is not clear at first (amsmath and amssymb). This line of option doesn't fit well into the geometry and leads to a scrollbar (mark 5) because the "Other Options" widget has not enough space. As a side note: The geometry package has it's own tab.

### A new Packages Tab

So the idea was born to have a separate "packages" tab widget having a table of (about 30) useful packages with a small description that can help one figure out if this package could be of interest. The four packages (mark 4) are moved there. Some packages have options that are assumed to be useful and are also shown (and automatically used when creating the document). As before txs stores the package names selected for later use.

The following image shows the new QuickStart dialog right after starting it with the new Packages tab:

grafik

Activation of that tab also activates a tool tip for this tab widget:

grafik

### The List of Packages

The following images show the complete list of packages:

grafik
grafik
grafik

As mentioned in the tool tip you can select packages with check boxes or by selecting lines:

grafik

Thus you may even select all packages within a second. When you need line numbers for better orientation you may activate them with the check box at the bottom:

grafik

And here's the final document:

grafik

octaeder:
This PR closes #2831. This mainly adds a new tab to the QuickStart Wizard named Packages, It shows in a table a small selection of packages each with a short description. Also those packages which are currently selectable with a checkbox are removed from the first tab  (Class Options), as well as the combobox for the Input Encoding which is now set to utf8 in any case.

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -283,30 +283,29 @@ void QuickDocumentDialog::Init()
     }
 
 	QTableWidget *table = ui.tableWidgetPackages;
-	table->clear();
-	table->setColumnCount(2);
-	table->setHorizontalHeaderLabels( QStringList( { tr("Package"), tr("Description") } ) );
+	table->clearContents();
 	table->horizontalHeader()->setStretchLastSection(true);
-	table->setSelectionMode(QAbstractItemView::ExtendedSelection);
-	table->setSelectionBehavior(QAbstractItemView::SelectRows);
+	table->setSelectionMode(QAbstractItemView::NoSelection);
+//	table->setSelectionBehavior(QAbstractItemView::SelectRows);
 	table->verticalHeader()->hide();
 
 	//each QStringList holds 2 items: the name of the package, and a short package description. These constitute a row of the table of the packages tab
 	QList<QStringList> packages = QList<QStringList>()
-		<< QStringList( {"amssymb"     , "Mathematical symbols from AMS"} )
-		<< QStringList( {"graphicx"    , "Graphics package, easily include images (s. Insert Graphic Wizard)"} )
-		<< QStringList( {"hyperref"    , "Support for hyperlinks in your document"} )
-		<< QStringList( {"mathtools"   , "Extension package to amsmath incl. fixes for bugs in amsmath, loads amsmath"} )
-		<< QStringList( {"amsthm"      , "Define your theorem like env., has to be loaded after amsmath"} )
-		<< QStringList( {"nameref"     , "Reference to names of chapters, sections, ..., loaded by hyperref"} )
-		<< QStringList( {"thmtools"    , "Extention package to amsthm"} )
-		<< QStringList( {"xcolor"      , "Sophisticated package for colors, with table option use colors in tables"} )
+		<< QStringList( {"amssymb"     , tr("Mathematical symbols from AMS")} )
+		<< QStringList( {"graphicx"    , tr("Graphics package, easily include images (s. Insert Graphic Wizard)")} )
+		<< QStringList( {"hyperref"    , tr("Support for hyperlinks in your document")} )
+		<< QStringList( {"mathtools"   , tr("Extension package to amsmath incl. fixes for bugs in amsmath, loads amsmath")} )
+		<< QStringList( {"amsthm"      , tr("Define your theorem like env., has to be loaded after amsmath")} )
+		<< QStringList( {"nameref"     , tr("Reference to names of chapters, sections, ..., loaded by hyperref")} )
+		<< QStringList( {"thmtools"    , tr("Extention package to amsthm")} )
+		<< QStringList( {"xcolor"      , tr("Sophisticated package for colors, with table option use colors in tables")} )
 		;
 	//add user given packages
 	for (const QString& package:otherPackagesList){
 		packages << QStringList( { package, "" } );
 	}
 	//setup packages table 
+	table->setRowCount(packages.size());
 	int row=-1;
 	int tableHeight = table->horizontalHeader()->height() + 2;
 	for (const QStringList& data:packages){
@@ -317,24 +316,20 @@ void QuickDocumentDialog::Init()
 		QTableWidgetItem *itemPkgName = new QTableWidgetItem(pkgName);
 		QTableWidgetItem *itemPkgDescription = new QTableWidgetItem(pkgDescription);
 
-		itemPkgName->setFlags(Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsUserCheckable);
-		itemPkgDescription->setFlags(Qt::ItemIsEnabled | Qt::ItemIsSelectable);
+		itemPkgName->setFlags(Qt::ItemIsEnabled | Qt::ItemIsUserCheckable);
+		itemPkgDescription->setFlags(Qt::ItemIsEnabled);
 
-		table->setRowCount(row+1);
 		table->setItem(row,0,itemPkgName);
 		table->setItem(row,1,itemPkgDescription);
 		tableHeight += table->rowHeight(row);
 
 		bool found = packagesUsed.contains(pkgName);
 		itemPkgName->setCheckState( found ? Qt::Checked : Qt::Unchecked );
-		updateSelected(row);
 	}
 	int height = ui.GridPackages->geometry().height();
 	// prevent ever increasing window height when constantly adding user defined packages
 	if (tableHeight < height || height == 0) ui.tableWidgetPackages->setMinimumHeight(tableHeight);
 	ui.tableWidgetPackages->setMaximumHeight(tableHeight);
-	connect(ui.tableWidgetPackages, SIGNAL(itemSelectionChanged()), SLOT(updateCheckState()));
-	connect(ui.tableWidgetPackages, SIGNAL(cellClicked(int,int)), SLOT(updateSelected(int)));
 
 	configManagerInterface->linkOptionToDialogWidget(&document_class, ui.comboBoxClass);
 	configManagerInterface->linkOptionToDialogWidget(&typeface_size, ui.comboBoxSize);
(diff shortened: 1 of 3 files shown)


src/quickdocumentdialog.cpp
```diff
@@ -303,15 +282,65 @@ void QuickDocumentDialog::Init()
         item->setCheckState(Qt::Unchecked);
     }
 
+	QTableWidget *table = ui.tableWidgetPackages;
+	table->clear();
+	table->setColumnCount(2);
+	table->setHorizontalHeaderLabels( QStringList( { tr("Package"), tr("Description") } ) );
```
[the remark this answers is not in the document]
octaeder: no

```
<item row="1" column="0">
 <widget class="QTableWidget" name="tableWidgetPackages">
  <property name="sizePolicy">
   <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
   </sizepolicy>
  </property>
 </widget>
</item>
```

src/quickdocumentdialog.cpp
```diff
@@ -303,15 +282,65 @@ void QuickDocumentDialog::Init()
...
+		QString pkgName = data[0],
+				pkgDescription = data[1];
+
+		QTableWidgetItem *itemPkgName = new QTableWidgetItem(pkgName);
+		QTableWidgetItem *itemPkgDescription = new QTableWidgetItem(pkgDescription);
+
+		itemPkgName->setFlags(Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsUserCheckable);
+		itemPkgDescription->setFlags(Qt::ItemIsEnabled | Qt::ItemIsSelectable);
+
+		table->setRowCount(row+1);
```
sunderme, line 323: probably not the most efficient way to build a table, but in this context it does not matter.

octaeder: What's the point? Always setting RowCount?

src/quickdocumentdialog.cpp
```diff
@@ -303,15 +282,65 @@ void QuickDocumentDialog::Init()
...
+	table->setColumnCount(2);
+	table->setHorizontalHeaderLabels( QStringList( { tr("Package"), tr("Description") } ) );
+	table->horizontalHeader()->setStretchLastSection(true);
+	table->setSelectionMode(QAbstractItemView::ExtendedSelection);
+	table->setSelectionBehavior(QAbstractItemView::SelectRows);
+	table->verticalHeader()->hide();
+
+	//each QStringList holds 2 items: the name of the package, and a short package description. These constitute a row of the table of the packages tab
+	QList<QStringList> packages = QList<QStringList>()
+		<< QStringList( {"amssymb"     , "Mathematical symbols from AMS"} )
```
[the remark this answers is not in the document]
octaeder: ok, is prepared

src/quickdocumentdialog.cpp
```diff
@@ -303,15 +282,65 @@ void QuickDocumentDialog::Init()
...
+		table->setRowCount(row+1);
+		table->setItem(row,0,itemPkgName);
+		table->setItem(row,1,itemPkgDescription);
+		tableHeight += table->rowHeight(row);
+
+		bool found = packagesUsed.contains(pkgName);
+		itemPkgName->setCheckState( found ? Qt::Checked : Qt::Unchecked );
+		updateSelected(row);
+	}
+	int height = ui.GridPackages->geometry().height();
```
sunderme, line 332: this looks odd.
Why is it necessary ?

src/quickdocumentdialog.cpp
```diff
@@ -491,46 +520,81 @@ void QuickDocumentDialog::addUserPaper()
...
 	if (dialog.exec() && !newoption.isEmpty()) {
-		otherBabelOptionsList.append(newoption);
+		otherOptionsList.append(newoption);
 		Init();
 	}
 }
 
-void QuickDocumentDialog::addUserOptions()
+void QuickDocumentDialog::addUserPackages()
 {
```
sunderme, line 557: this dialog is a bit underwhelming, especially as txs has a list of all available packages.
But maybe a task for a future PR.

src/quickdocumentdialog.cpp
octaeder: You mean referencing updateSelected? This ensures that the line having checkState set appears selected. Otherwise one could click in a row (say in second column) which turns the row selected (blue) but the checkbox isn't selected or in a checkbox but row is not selected. Thus this is done to avoid visual ambiguity.

src/quickdocumentdialog.cpp
octaeder: I know

src/quickdocumentdialog.cpp
```diff
@@ -491,46 +520,81 @@ void QuickDocumentDialog::addUserPaper()
...
 	UniversalInputDialog dialog;
 	dialog.addVariable(&newoption, tr("New:"));
 	if (dialog.exec() && !newoption.isEmpty()) {
-		otherOptionsList.append(newoption);
+		otherPackagesList.append(newoption);
 		Init();
 	}
 }
+//when package selections change
+void QuickDocumentDialog::updateCheckState()
```
sunderme, line 567: I am not sure what exactly the logic is supposed to do, but basically if you select a row by clicking on a description text, all other selections/checked states are removed.
Let's stick with Qt logic, i.e. you click on the box to select something to be inserted ...

src/quickdocumentdialog.cpp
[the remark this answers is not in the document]
sunderme: shouldn't it be set there ?

src/quickdocumentdialog.cpp
sunderme: yes, especially as you know the number from the start.

src/quickdocumentdialog.cpp
sunderme: also the height stuff.
There should be automatisms for that.

octaeder: Maybe, I used hours to figure out which way could work at to try them. But nothing helped. I think the main reason for this is your which not to see a scrollbar at the beginning. If I don't specify the height, then Qt will distribute space between table and what is below the table. Now one could think change ratio but if the table widget height then is larger as needed an empty space appears after the rows but still inside the frame of the table widget. I have seen this when setting the spacer size policy to minimum. And as a side effect when clicking in this extra space all rows get unselected.

src/quickdocumentdialog.cpp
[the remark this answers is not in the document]
octaeder: arraydialog.ui hasn't it (l. 35, obviously because we don't no the columns), bibtexdialog.ui has it (l. 55). So I'll move it, even so I think that it makes translations more complicate.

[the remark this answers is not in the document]
octaeder: ok, is prepared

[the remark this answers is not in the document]
octaeder: I was mislead. This doesn't work, reason is calling tableCear() in the init function which is called in each addXxx function. So I returned to my solution.

src/quickdocumentdialog.cpp
sunderme: Have you tried a vertical spacer below the table ?

src/quickdocumentdialog.cpp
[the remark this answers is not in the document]
sunderme: `...->clearContent()`

src/quickdocumentdialog.cpp
octaeder: I you don't have a spacer then the table widget fits exactly the header plus rows, but it is vertically centered in the tab which looks random. So we need a spacer (I already mentioned above) below the table widget that moves the table upwards. To do so the spacer needs a policy of minimumExpanding. But this means no other than that the spacer and the table take their space in a specific ratio, normally 1:1 (to prevent this I set the minimum and maximum height of the table to the same value). But this will introduce a scrollbar because the table needs more space. Of course we could now try to approximate the real number needed as ratio by rational numbers (you need to use integers for specifying proportions). But after adding one more row (or just changing to windows style which has different decorations not counted into the heights) this will break again. I don't see how we could break out of the circle.

src/quickdocumentdialog.cpp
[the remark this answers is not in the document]
octaeder: Ah, good hint, that helps. Ok, prepared.

src/quickdocumentdialog.cpp
octaeder: You may try what happens if you remove my connected signals (I started development without these). But I found that checking more than three items is cumbersome since the checkboxes have a small size and it's slow (remember the project offering all known packages). Selecting by dragging the pointer over the rows or clicking into the (large) rows is quick and easy. Since this is a table (not a list) you have to use modifiers as it is known from windows file explorer or calc/excel. Thus clicking one row results in a selection containing only this row. Right, if you slightly miss a checkbox you want to click on then this row will be the only one selected. But you can redo your selection quickly and take care the next time. Maybe we could add the tooltip text before the table so help is immediately visible (s. image)?

grafik

If the wizard starts with initial selections (maybe a longer table with a scrollbar) the selected lines make it easier to recognize which rows are selected than the checkboxes on their own would do. And you can go easily through the descriptions of the selected packages without moving your eyes from left to right.

I also investigated the case when I keep the selection theme but remove the checkboxes at all. But at the first clicks I was not convinced that this is enough feedback for everyone to know that one already selected packages. So I added the checkboxes again which give a pretty clear feedback. But I don't use them for selecting for the reasons said, but you can if you want to.

This image may clarify why I wanted to synchronize row selections and checkStates:

grafik

Compare to this (I just expanded the previous selection, no signals):

grafik

So I would recommend allowing selections.

octaeder: You may try what happens if you remove my connected signals (I started development without these). But I found that checking more than three items is cumbersome since the checkboxes have a small size and it's slow (remember the project offering all known packages). Selecting by dragging the pointer over the rows or clicking into the (large) rows is quick and easy. Since this is a table (not a list) you have to use modifiers as it is known from windows file explorer or calc/excel. Thus clicking one row results in a selection containing only this row. Right, if you slightly miss a checkbox you want to click on then this row will be the only one selected. But you can redo your selection quickly and take care the next time. Maybe we could add the tooltip text before the table so help is immediately visible (s. image)?

grafik

If the wizard starts with initial selections (maybe a longer table with a scrollbar) the selected lines make it easier to recognize which rows are selected than the checkboxes on their own would do. And you can go easily through the descriptions of the selected packages without moving your eyes from left to right.

I also investigated the case when I keep the selection theme but remove the checkboxes at all. But at the first clicks I was not convinced that this is enough feedback for everyone to know that one already selected packages. So I added the checkboxes again which give a pretty clear feedback. But I don't use them for selecting for the reasons said, but you can if you want to.

This image may clarify why I wanted to synchronize row selections and checkStates:

grafik

Compare to this (I just expanded the previous selection, no signals):

grafik

So I would recommend allowing selections.

sunderme:
thanks

PR #2847, opened 2023-01-12
Improve pinch-to-zoom performance on macOS

Iksas:
While testing the changes introduced in #2840 on multiple machines, two issues were discovered and fixed in this PR:

 ### Degraded performance while zooming

This issue was fixed by taking the zoom scale from the `QGestureEvent` instead of the `QTouchEvent`. This way, no more calculations are performed to determine the zoom scale. The `QTouchEvent` is now used only for the zoom coordinates:

- `QTouchEvent` <= unreliably generated
  - Zoom coordinates

- `QGestureEvent`
  - ~~*Zoom coordinates*~~ <= broken on macOS
  - Zoom scale
  - Zoom event detection

The performance with this fix is identical to the release version of Texstudio.

### Qt sometimes stops generating `QTouchEvent`s

Texstudio would sometimes zoom to the wrong coordinates when no `QTouchEvent`s were generated. This was fixed by setting the `Qt::WA_AcceptTouchEvents` attribute on the `PDFWidget`.

All changes affect the macOS version only.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -576,8 +576,8 @@ PDFWidget::PDFWidget(bool embedded)
 	, docPages(0)
 	, saveScaleFactor(1.0)
 	, saveScaleOption(kFitWidth)
-	, pinchStartedScaleFactor(1.0)
-	, pinchGestureDetected(false)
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

PR #2852, opened 2023-01-16
Fix a cwl typo

muzimuzhi:
`\sbox` takes two margs, rather than a marg followed by an oarg.
See for example https://latexref.xyz/_005csbox-_0026-_005csavebox.html .


sunderme:
thanks

PR #2855, opened 2023-01-17
CWLs!

ponte-vecchio:
G'day,

This MR adds some 20 cwl files for [pdfMsym](https://www.ctan.org/pkg/pdfmsym) package as well as a large number of packages available in [KTUG](http://www.ktug.org/xe/) [repos](https://mirror.ischo.org/KTUG/texlive/texmf-dist/).

Since this is my first MR, I'll keep this MR a draft so that the files can be reviewed


completion/ifpxltex.cwl
```diff
@@ -0,0 +1,48 @@
...
+TeX
+XeLaTeX
+XeTeX
+#endkeyvals
+
+#keyvals:\IfpxlTeX#c, \IfpxlTeXpxl*#c, \pxlRequireTeX#c
+p
+x
+l
+l,p
```
sunderme, line 35: why are comma separated values here ?
if you define p,x&l previously, "l,p" are valid as any comma separated combination.

completion/jiwonlipsum.cwl
```diff
@@ -0,0 +1,18 @@
...
+# ponte-vecchio 2023-01-17 for v0.6
+
+#keyvals:\usepackage/jiwonlipsum#c
+nopar
+numbers
+#endkeyvals
+
+\jiwon
+\jiwon[%<num%>]
+\jiwon[%<char%>]
```
sunderme, line 11: do you really want to define the same command twice, only with different argument name ?

completion/ksbaduk.cwl
```diff
@@ -0,0 +1,103 @@
...
+posmark
+imageback
+imagefile=%<file%>
+textcmds
+numberfont=%<font%>
+tmarkfont=%<font%>
+#endkeyvals
+
+## 2.2 Global options
+\BadukpanSize{%<number%>}%definition
```
sunderme, line 18: I doubt that %definition is doing anything here.
Not sure what the intention is ?

completion/colorjamo.cwl
```diff
@@ -0,0 +1,12 @@
+# colorjamo package
+# ponte-vecchio 2023-01-17 for v.0.3
+
+#include: luacolor
+#include: transparent
```
mbertucci47, line 5: The cwl format is not very flexible when it comes to spaces. This, and every other `#include:`, needs to be
```
#include:luacolor
#include:transparent
```

completion/graphicsonthefly.cwl
```diff
@@ -0,0 +1,107 @@
...
+#endif
+
+#ifOption:animate
+#include: animate
+\animatedgifonthefly[%<options%>%keyvals]{%<fps%>}{%<image name%>}{%<start%>}{%<end%>}{%<url%>}
+\animatedgif{%<fps%>}{%<image name%>}{%<start%>}{%<end%>}{%<url%>}
+\usegifonthefly[%<options%>%keyvals]{%<image name%>}{%<url%>}
+#endif
+
+#keyvals:\animatedgifonthefly#c, \usegifonthefly#c
```
mbertucci47, line 30: Similarly the space needs to be removed here, it should be `#keyvals:\animatedgifonthefly#c,\usegifonthefly#c`

completion/graphicsonthefly.cwl
```diff
@@ -0,0 +1,107 @@
...
+all
+curl
+nocoalesce
+convert
+#endkeyvals
+
+## Animations
+#ifOption:all
+#include: animate
+\animatedgifonthefly[%<options%>%keyvals]{%<fps%>}{%<image name%>}{%<start%>}{%<end%>}{%<url%>}
```
mbertucci47, line 18: The `%keyvals` in `%<options%>%keyvals` is superfluous since `%<options%>` alone is specially recognized for keyvals

completion/grruby.cwl
```diff
@@ -0,0 +1,29 @@
...
+
+\begin{grruby}
+\end{grruby}
+\begin{grrubypars}
+\end{grrubypars}
+
+
+#ifOption:expos=true
+\grexpos
+\grexpos[%<options%>]#true,false
```
mbertucci47, line 24: `#true,false` is only for use in keyvals

completion/hanjacnt.cwl
```diff
@@ -0,0 +1,175 @@
...
+shi
+bae
+chu
+man
+eok
+cho
+#endkeyvals
+
+## 2.6 Counters
+\HANJA{%<counter%>}#d*
```
mbertucci47, line 79: Do these define a command? Otherwise the `#d` does not belong here, and the trailing `*` doesn't do anything (I think). If you're trying to mark them unusual then you'd use `#*` here (or `#*d` if that were the correct spec)

completion/kocircnum.cwl
```diff
@@ -0,0 +1,146 @@
...
+Raise=%<dim%>%l
+font=%<font options%>
+fontplus=%<font options%>
+base
+reset
+#endkeyvals
+#endif
+
+#ifOption:hz
+shape=%<shape%>#circle,circ,rectangle,rect,oval,ball
```
mbertucci47, line 112: Missing a `#keyvals:` here

completion/hanjacnt.cwl
sunderme: correct

completion/ksmisc.cwl
```diff
@@ -0,0 +1,13 @@
...
+#include:ksjosaref
+#include:kstextks
+#include:kslinematters
+#include:preparefont
+#include:singlepageempty
+
+\ks_pop_left:NN %<tlvar1%> %<tlvar2%>#S
+\ks_pop_right:NN %<tlvar1%> %<tlvar2%>#S
+\ks_clist_set_from_tl:NN %<clistvar%> %<tlvar%>#S
+\ks_seq_set_from_tl:NN %<seqvar%> %<tlvar%>#S
```
mbertucci47, line 13: There's not currently any support for expl3 syntax so these lines don't do anything

completion/pdfMsym.cwl
```diff
@@ -0,0 +1,196 @@
...
+\varwidecheck{%<arg%>}#m
+\varwidetilde{%<arg%>}#m
+
+## 2.5 Extendable Operators
+\suum#m
+\prood#m
+
+# 3. Defining your own symbols
+### Pre-defined lineheads
+\@rarrow#*
```
mbertucci47, line 160: Any lines containing an `@` are ignored so you don't need to list these internal commands

completion/ifpxltex.cwl
ponte-vecchio: Thanks for pointing this out! Reflected in the new commit.

completion/jiwonlipsum.cwl
ponte-vecchio: Changed to `jiwon[%<num/char%>]`

completion/colorjamo.cwl
ponte-vecchio: Thanks for pointing this out! Used `grep` to look for all instances of `#include: <pkg>` and fixed them in the new commit.

completion/ksbaduk.cwl
ponte-vecchio: Removed in the new commit

completion/graphicsonthefly.cwl
ponte-vecchio: Reflected in the new commit with all other cwl files, grep saves the day :)

completion/graphicsonthefly.cwl
ponte-vecchio: Reflected in the commit, changed all `%<options%>%keyvals` to just `%<options%>`

completion/grruby.cwl
ponte-vecchio: Made a new keyval list in the new commit

completion/hanjacnt.cwl
ponte-vecchio: It converts the value of the counter being used to Hanja/Hangul, so probably not a `d` here. Reflected in the new commit.

completion/ksmisc.cwl
ponte-vecchio: Removed!

completion/ksbaduk.cwl
```diff
@@ -15,10 +15,10 @@ tmarkfont=%<font%>
 #endkeyvals
 
 ## 2.2 Global options
-\BadukpanSize{%<number%>}%definition
-\BadukpanColor{%<color%>%color}%definition
-\BackgroundColor{%<color%>%color}%definition
-\NumberFont{%<font%>%font}%definition
+\BadukpanSize{%<number%>}
+\BadukpanColor{%<color%>%color}
+\BackgroundColor{%<color%>%color}
```
mbertucci47, line 20: These should just be
```
\BadukpanColor{color}
\BackgroundColor{color}
```
though it would be nice to be able to specify color args this way

completion/pdfMsym.cwl
ponte-vecchio: Removed!

completion/ksbaduk.cwl
ponte-vecchio: Ah, sad

completion/kocircnum.cwl
ponte-vecchio: Added!

completion/kswrapfig.cwl
```diff
@@ -0,0 +1,38 @@
...
+CaptionName=#figure,table
+#endkeyvals
+
+#ifOption:usebox
+\ksinsbox[%<position%>%keyvals]{%<number%>}{%<cmd%>}
+#keyvals:\ksinsbox#c
+r
+l
+c
+#endkeyvals
```
mbertucci47, line 38: Missing `#endif` and option name does not match option given in `#keyvals:\usepackage/kswrapfig#c`

completion/pdfMsym.cwl
```diff
@@ -0,0 +1,161 @@
+# pdfMsym package
```
mbertucci47, line 1: the file is `pdfmsym.tex` so even though this cwl would have to be manually turned on in the settings (only .sty and .cls files are picked up automatically), I think it would be better to name it `pdfmsym.cwl`

completion/kswrapfig.cwl
```diff
@@ -0,0 +1,38 @@
...
+Pos=#r,l
+Width=%<dim%>%l
+InPos=#r,c,l
+Sep=%<dim%>%l
+Indent=%<dim%>%l
+Lower=%<dim%>%l
+Caption={%<text%>}
+LastLineSkip=%<dim%>%l
+FirstLineSkip=%<dim%>%l
+List=%<env name%>#itemize,enumerate
```
mbertucci47, line 20: The correct syntax is either `List=%<env name%>` or `List=#itemize,enumerate`

completion/kswrapfig.cwl
```diff
@@ -0,0 +1,38 @@
...
+VAdjust=%<dim%>%l
+UseBox=%<bool%>#true,false
```
mbertucci47, line 22: This should just be `UseBox#true,false`

completion/grruby.cwl
```diff
@@ -0,0 +1,34 @@
...
+rubystyle=#default,gromob,XXruby,ruby,progress,rup,lw,rw,user
+rubycolor=%<color%>#%color
+expos=#true,false
+pdflatex
+#endkeyvals
+
+\grfoo %<hangul%>(%<hanja%>)
+\grfoo %<hangul%>(%<hanja%>){%<footnote%>}
+\grrubystyle{%<options%>}#default,gromob,XXruby,ruby,progress,rup,lw,rw,user
+\grrubycolor{%<options%>%color}
```
mbertucci47, line 14: These are not correct for reasons previously stated

completion/kswrapfig.cwl
ponte-vecchio: This is an immutable list in fact, the documentation specifically lists those two as examples, but it can be any environment name. Should I remove them?

completion/grruby.cwl
```diff
@@ -0,0 +1,34 @@
+# grruby package
+# ponte-vecchio 2023-01-17 for v0.7
+
+#keyvals:\usepackage/grruby#c
+rubystyle=#default,gromob,XXruby,ruby,progress,rup,lw,rw,user
+rubycolor=%<color%>#%color
```
mbertucci47, line 6: This should just be `rubycolor=#%color`

completion/grruby.cwl
```diff
@@ -0,0 +1,34 @@
...
+expos=#true,false
```
mbertucci47, line 7: For boolean keys you should just write `expos#true,false`

completion/kswrapfig.cwl
mbertucci47: Then just `List=%<env name%>`

completion/grruby.cwl
ponte-vecchio: Does this work for when the default is `expos=false`, and you must pass `expos=true` in order to enable some features cf. just passing `expos` to assume `true`?

mbertucci47: Almost every way to define boolean keys in latex views `expos` and `expos=true` the same. That's true here since it's done with l3keys. So the special `expos#true,false` makes it such that the completer will only complete the key to `expos`, not `expos=` as it would with `expos=#true,false`. The only time it makes sense to write `key=#true,false` is when the default is true, which IMO is bad practice on the part of the package writer

ponte-vecchio: Resolved! just went with `expos#true,false`

completion/grruby.cwl
ponte-vecchio: Done

completion/grruby.cwl
ponte-vecchio: Fixed! Thanks :)

sunderme:
thanks

PR #2860, opened 2023-01-20
An Editor option which swapps line numbers column and line marks column

### LINKED ISSUE #2220: Enhance creating bookmarks

octaeder:
This PR adds a checkbox option to the editor config page. This option allows the user to have the line numbers of the editor left to the line marks. Thus both columns have constant distance to the editor window independent of the  total number of lines of the editor. This PR closes #2220 where this was discussed.

Note: Changing this option takes only effect when opening new editors or when restarting txs. This seems to be ok since this option is intended to be changed only once. Even so I tried to code that all editors (visible or invisible) are changed immediately. But I didn't succeed.

Comparison Line Marks to the left (default) vs. Line Numbers to the left:

linesAndMarks

Option:

grafik

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -1670,6 +1670,16 @@ Then you can select a new shortcut by one of the following ways:
                     </item>
                    </widget>
                   </item>
+                  <item row="11" column="2">
+                   <widget class="QCheckBox" name="checkBoxLineNumbersFirst">
+                    <property name="toolTip">
+                     <string>This moves the column with the line numbers from the right to the left of the column with the line marks.</string>
+                    </property>
+                    <property name="text">
+                     <string>Line numbers to the left of line markers</string>
+                    </property>
+                   </widget>
+                  </item>
                   <item row="15" column="0" colspan="2">
                    <widget class="QCheckBox" name="checkBoxInlineCheckNonTeXFiles">
                     <property name="text">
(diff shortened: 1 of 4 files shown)


sunderme:
In my opinion this is not a (good) solution for #2220 .
As stated in that discussion, the point is unclear visualization for where to click.
Having gray bars left **and** right of the line numbers is not really an improvement.
The solution is tolerance of the click, i.e. clicking on the line number panel also sets bookmarks.

octaeder:
I agree that lines and colors are not a good solution. So I didn't do so.
A click on a line number sets the cursor at the start of the line. Maybe not only me would miss this.
This PR gives the user an option to choose what he likes. And the solution gives orientation without dropping functionality.
In addition the large and variable distance between the marks and numbers is removed.
You could add your suggestion as an option, too.

sunderme:
basically anything behind an option is dead code

So, my suggestion would be to change the cursor, as qtcreator does it.
grafik
Then we don't need to change the layout.

sunderme:
This is implemented now, as it is a one-liner.

Issue #995, opened 2020-03-24
On win10 the save and cancel buttons are not visible.


bernhard-kleine:
### Environment
TeXstudio 2.12.22 (git 2.12.22)
Using Qt Version 5.12.1, compiled with Qt 5.12.1 R
Win10,
texlive
### Expected behavior
When Opening the Option->Configure Texstudio, the screen should make the "ok" or "save" and "cancel" visible at the bottom of the screen. 
### Actual behavior
As of now, Options -> configure the screen immediately changes to 100+ % of fullscreen, 
in the bottom of the screen some button seem to exist, but they cannot be clicked even if you know which button to expect.  

### How to reproduce
Should work

gaplin:
Same behaviour on newer versions:

### Environment
- TeXstudio: 4.4.1
- Qt: 6.4.1
- OS: Windows 11
- TeX distribution: miktex

octaeder:
My txs build of current master with Qt 6.4.2, Win10 opens a config dialog of size 1378x864 (same as with 4.4.1). I think this should be small enough for current displays (but not for HD ready). Screenshot scaled to 50%:

![grafik](https://user-images.githubusercontent.com/102688820/213323729-26c73e5e-e12a-42ce-9273-e7758c4ad1b3.png)

gaplin:
On my laptop there is a 125% display scaling enabled in windows settings by default. If I disable it, the dialog is fully visible.
I think that 125% scaling on laptops is quite common so making this window a little smaller might be a good idea.

octaeder:
When I scale the given size of the dialog by 125% the result is 1733x1080. So it should still fit on a FullHD screen. Can you explain why the ok and cancel buttons are not visible on your screen?

Maybe it could be useful to know following about your display: Number of pixels in horizontal and vertical direction, the screen format and the length of the diagonal.

gaplin:
It's typical 1920x1080(16:9) 15,6 inch screen.
I'll attach screenshot how it looks. Buttons are hidden behind windows task bar.

![image](https://user-images.githubusercontent.com/50521366/213588591-9a3db41c-a5d8-4baa-8b87-1c0e725f1204.png)

octaeder:
I can't figure out how the size is set. There seems to be code in configdialog.cpp (l. 600ff) which tries to reduce the window size. But program control returns before any change is done. Reason is that the size used for comparison is less than the screen size. But even forcing the code to set the dialog size to a smaller value doesn't work.
[...]

PR #2863, opened 2023-01-21
#995 fix - Removed resizing from configManager. Changed size of dialo…

gaplin:
[#995](https://github.com/texstudio-org/texstudio/issues/995) fix.

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -597,17 +597,22 @@ ConfigDialog::ConfigDialog(QWidget *parent): QDialog(parent,Qt::Dialog|Qt::Windo
     connect(ui.tbRevertPDF, SIGNAL(clicked()), this, SLOT(revertClicked()));
 
 	// limit dialog size
-	QRect screen = QGuiApplication::primaryScreen()->geometry();
+	QRect screen = QGuiApplication::primaryScreen()->availableGeometry();
 	if (!screen.isEmpty()) {
 		int nwidth = width(), nheight = height();
 		if (nwidth > screen.width()) nwidth = screen.width();
 		if (nheight > screen.height()) nheight = screen.height();
 		if (nwidth == width() && nheight == height()) return;
-		resize(nwidth, nheight);
-		move(frameGeometry().right() > screen.right() ? screen.left() : x(),
-		     frameGeometry().bottom() > screen.bottom() ? screen.left() : y());
+		resize(nwidth - 32, nheight - 32);
 	}
 }
+
+void ConfigDialog::moveIfExceedsScreen() {
+	QRect screen = QGuiApplication::primaryScreen()->availableGeometry();
+	move(frameGeometry().right() > screen.right() ? screen.left() : x(),
+		   frameGeometry().bottom() > screen.bottom() ? screen.top() : y());
+}
+
 
 ConfigDialog::~ConfigDialog()
 {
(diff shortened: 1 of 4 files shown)


src/configdialog.cpp
```diff
@@ -597,18 +597,23 @@ ConfigDialog::ConfigDialog(QWidget *parent): QDialog(parent,Qt::Dialog|Qt::Windo
...
+	QRect screen = QGuiApplication::primaryScreen()->availableGeometry();
 	if (!screen.isEmpty()) {
 		int nwidth = width(), nheight = height();
 		if (nwidth > screen.width()) nwidth = screen.width();
 		if (nheight > screen.height()) nheight = screen.height();
 		if (nwidth == width() && nheight == height()) return;
-		resize(nwidth, nheight);
-		move(frameGeometry().right() > screen.right() ? screen.left() : x(),
-		     frameGeometry().bottom() > screen.bottom() ? screen.left() : y());
+		resize(nwidth - 32, nheight - 32);
```
sunderme, line 606: numbers coming out of the blue which are supposed to work over different styles/OS don't give much confidence ...

sunderme:
and why not going for maximized state when detecting small screen size ?

gaplin:
This fixes in my opinion most popular cases(windows laptops with fullHD screens and 125% scaling enabled by default).
Not maximizing is not fixing the issue. You can still minimze and get the same problem so i think it is better to do it this way.

src/configdialog.cpp
gaplin: It's change from one numbers to another:
image
Above ones were causing the issue.

gaplin:
I've created new [branch](https://github.com/gaplin/texstudio/tree/%23995-buttons-not-visible-in-config-dialog) with another solution. This time with [adjustSize](https://doc.qt.io/qt-6/qwidget.html#adjustSize).

But this solution behaves poorly with low resolutions. Here ss with 800x600 ubuntu:
image

Ss with first solution:
image

sunderme:
so, there is no final solution yet and I wait ?

gaplin:
I would say this is the final solution. 

AdjustSize solution changes size of the dialog for all users and it's not good for lower resolutions.

This solution doesn't have that problem since it's making this window only a little bit smaller for those users.

Alos considering that config dialog should be big because of its content I would just go with this one.

sunderme:
I can't really test. Let's hope for the best.
Thanks.

src/configdialog.cpp
octaeder: @gaplin nheight-32 is used because the window decoration has a large height mainly due to the title bar. So I wonder why you use nwidth-32. This leaves many space to the left and right of the dialog because the left and right decorations are typically thin lines.

We should not use magic numbers (i.e. 32) if possible. If we do so we could add an explanation as a comment:

``// subtract 32 because the window decoration has to be taken into account, but we can only estimate the values``

I tried show() as explained in the [qt forum](https://forum.qt.io/topic/41135/solved-qdesktopwindow-availablegeometry-not-excluding-taskbar-qt-5-2-windows-7). For me it did not work but maybe you can figure out.

gaplin: The -32 on width is for side taskbar. I didn't think about it in the way you desribed(title bar). 
I thought window as a whole was possibly misplaced in both axis.

I'll take a look.

octaeder: availableGeometry is designed to calculate the size of the difference between the screen area and the taskbar (have a look [here](https://doc.qt.io/qt-6/qscreen.html#availableGeometry-prop)).

gaplin: Ye, so the availableGeometry simply shows size of maximized window.

Following hint with show() method I got rid of this 32 const. I'll create another PR.

gaplin: New PR: #2868

Issue #2816, opened 2023-01-01
ntheorem's Math Definitions in English and German


octaeder:
Some time ago I added the math definitions (aka ntheorem's standard environments) to the math menu (#2222). But there are only the English variants of them (Corollary, Proof, ... that print in English). This decision was the result of a discussion about too many entries in the menu list. So the German variants (Korollar, Beweis that print in German) were dropped.

I'm currently experimenting with a different approach that is similar to the solution for the \left/\right brackets commands combined with parentheses of all kind (s. ![grafik](https://user-images.githubusercontent.com/102688820/210185609-6a891d96-224d-44ce-9b5d-6f7c35d66e49.png) in the side panel). That means there could be a tags xml file that contains the sections English and German with tags of the env names in English and German respectively. This would give following options: 1) The tags list would be integrated in the side panel (like left/right). 2) The desired tags (either German or English) could be easily added to customizable toolbars which offer quick access (s. image below).

So those writing English documents as well as those writing in German could easily take advantage of the ntheorem package.

This solution would need a new icon for the side panel (maybe in 3 variants for the icon sets) and maybe the Math definitions menu should be removed.

![grafik](https://user-images.githubusercontent.com/102688820/210185803-d8667608-339a-41a2-a736-1d50cc180161.png)

A slightly different solution could be to keep the Math definitions menu and add only the German tags to the xml file. That's the state of my current build (and I reuse the icon for left/right):

![grafik](https://user-images.githubusercontent.com/102688820/210186328-7d5a6516-6703-4b78-84f6-d720b18dd370.png)

Any suggestions/ideas on my proposal?

sunderme:
the maybe most important question: are **you** going to use the ntheorem via the tags or menus ?

if yes, and be honest about that, I would prefer no change in the math menu (removal of the menu) but rather a xml tags.
The probably nicer solution would be a dynamic solution to add the additional entries to the math menu if the user wants them.

octaeder:
I mainly use the side panel with the first four views. I tried before to avoid adding the definitions tags to the side panel. But the code depends on the xml being there. I would prefer the math toolbar as the place (see my picture). Automatic insertion of the code is more reliable than typing. If you don't need a label use completion. My concern is rather the less experienced user, the power user may even remove the menus via config dialog. I also thought about an option. But where should it be in the config? Menus page seems to fit best, but I don't think that we can put it there. For now I'll stop working on that issue for a while...

sunderme:
[...]

PR #2864, opened 2023-01-22
manual: Define and Load menus

octaeder:
Update manual as suggested in #2816.

### DIFF
--- a/utilities/manual/source/background.md
+++ b/utilities/manual/source/background.md
@@ -117,11 +117,6 @@ editor when this item is selected. You may add the `info` attribute which holds 
 
 Be aware that TeXstudio adds some of its menus in a different way.
 
-```note
-Watch out for menu definitions on [texstudio-macro]{https://github.com/texstudio-org/texstudio-macro} which can be downloaded with
-the macro editor.
-```
-
 ## Description of the cwl format
 
 cwl stands for completion word list and is a file format originally used
--- a/utilities/manual/source/background.md
+++ b/utilities/manual/source/background.md
@@ -82,6 +82,41 @@ Note: The most common tests are run automatically, if there were changes
 to the executable (i.e. TXS has been compiled since the last run).
 Furthermore all tests are run once a week.
 
+## Menu Definitions
+
+Menu definitions are used to set-up menus. You can [configure the Menu]{configuration.md#configuring-the-menu-(advanced-option)} to
+have menus defined by you. Of course the existing standard menu is defined in this way. The definition is located in the source
+file `texstudio/uiconfig.xml`. This shows an excerpt:
+
+```xml
+<TexMakerXUI>
+<menu id="main/latex" text="&amp;LaTeX">
+  <insert id="documentclass" text="\documentclass" insert="\documentclass[%&lt;options%|%>]{%&lt;class%>}"/>
+  <insert id="usepackage" text="\usepackage{}" insert="\usepackage[%&lt;options%>]{%&lt;package%|%>}"/>
+  ...
+  
+  <menu id="sectioning" text="&amp;Sectioning">
+    <insert id="part" text="part" insert="\part{%&lt;title%>}" icon="part"/>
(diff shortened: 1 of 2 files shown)


octaeder:
Is there a place where anyone could make his definitions available to everyone as with the macros? First texstudio-macro came into mind, but this may collide with the download feature of the macro editor.

utilities/manual/source/configuration.md
```diff
@@ -797,6 +797,16 @@ The "show all" check box allows to see the complete menu structure of the main w
 Changes to the Slots are discouraged unless you really know what you are doing. No support for changes here is given.
 ```
 
+You can use command `app.loadManagedMenu()` in [script macros](advanced.md#script-macros) to load a file that contains your definitions for menus
```
sunderme, line 800: And now it would be nice if the general style of the manual would be followed.
Everything is expressed neutrally, and all of a sudden "you can"...

utilities/manual/source/background.md
```diff
@@ -82,6 +82,41 @@ Note: The most common tests are run automatically, if there were changes
 to the executable (i.e. TXS has been compiled since the last run).
 Furthermore all tests are run once a week.
 
+## Menu Definitions
+
+Menu definitions are used to set-up menus. You can [configure the Menu]{configuration.md#configuring-the-menu-(advanced-option)} to
```
sunderme, line 87: `Menu definitions are used to set-up menus.` this does not explain anything.

utilities/manual/source/background.md
```diff
@@ -82,6 +82,41 @@ Note: The most common tests are run automatically, if there were changes
 to the executable (i.e. TXS has been compiled since the last run).
 Furthermore all tests are run once a week.
 
+## Menu Definitions
```
sunderme, line 85: Too early. cwl definition is way more important, so place it after that.

utilities/manual/source/background.md
```diff
@@ -82,6 +82,41 @@ Note: The most common tests are run automatically, if there were changes
...
+
+</menu>
+</TexMakerXUI>
+```
+
+The `\<TexMakerXUI\>` tag is mandatory and contains one ore more `\<menu\>` tags. Each menu must have a unique `id` attribute. The
+menu with id \"main/latex\" is the \"LaTeX\" menu (s. text attribute) in the menu bar of TeXstudio, while the menu with id
+\"sectioning\" is a submenu of the LaTeX menu. The character & marks the character which can be used for quick access to the menu.
+
+Most menus not only have submenus but also have items you can select. These are defined with `\<insert\>` tags. As with the menu tags
```
sunderme, line 114: "Most menus not only have ..." ? This is a manual ...

utilities/manual/source/background.md
```diff
@@ -82,6 +82,41 @@ Note: The most common tests are run automatically, if there were changes
...
+these must have a unique id and a text attribute. In addition we have an `insert` attribute which holds the text to be inserted in the
+editor when this item is selected. You may add the `info` attribute which holds text that can be presented by TeXstudio here and then.
+
+Be aware that TeXstudio adds some of its menus in a different way.
```
sunderme, line 118: How does that matter ?

utilities/manual/source/background.md
```diff
@@ -82,6 +82,41 @@ Note: The most common tests are run automatically, if there were changes
...
+    ...
+  </menu>
+
+  ...
+
+</menu>
+</TexMakerXUI>
+```
+
+The `\<TexMakerXUI\>` tag is mandatory and contains one ore more `\<menu\>` tags. Each menu must have a unique `id` attribute. The
```
sunderme, line 110: There is no mention that this is xml, so reference to a tutorial for that ?

Probably a better way to explain part, is to quote them and explain them directly afterwards, possibly with an image what the end result looks like.

sunderme:
Probably better to merge this and then I can adapt.

octaeder:
All my experience with updating this manual told me to not longer touch it, especially after your tremendous rework and introducing to generate it. Since you asked to do so I tried again. And again nothing more than critics on my work. Please don't ask again.

PR #2868, opened 2023-01-22
#995 using frameGeometry to calculate available space for configDialog.

gaplin:
Using frameGeometry to calculate difference between size of the dialog with and without frame to get rid of 32 constant.
Doing this only for vertical part since only that one is affected.

Modified solution from [qt forum post](https://forum.qt.io/post/228552).

This PR resolves https://github.com/texstudio-org/texstudio/issues/995.


octaeder:
This PR solves #995.

gaplin:
"Solves" didn't work, resolves did.

octaeder:
> "Solves" didn't work, resolves did.

You are right, sorry. All possibilities can be found [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). Thanks.

PR #2876, opened 2023-01-23
Update symbols

groverlynn:
- Fix a bug where `<` is omitted from unicodes
- Correct wrong unicodes
- Delete duplicates and put paired symbols together

### DIFF
--- a/src/symbolpanel/symbollistmodel.cpp
+++ b/src/symbolpanel/symbollistmodel.cpp
@@ -13,7 +13,7 @@
  */
 
 SymbolListModel::SymbolListModel(QVariantMap usageCountMap, QStringList favoriteList) :
-    m_darkMode(false)
+	m_darkMode(false)
 {
 	foreach (const QString &key, usageCountMap.keys()) {
 		usageCount.insert(key, usageCountMap.value(key).toInt());
@@ -123,7 +123,7 @@ void SymbolListModel::loadSymbols(const QString &category, const QStringList &fi
 			symbolItem.packages = img.text("Packages");
 			symbolItem.unicode = img.text("CommandUnicode");
 			symbolItem.iconFile = fileName;
-            symbolItem.icon = QIcon(fileName);
+			symbolItem.icon = QIcon(fileName);
 		}
 		if (!symbolItem.unicode.isEmpty()) {
 			// convert to real unicode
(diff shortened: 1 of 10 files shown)


sunderme:
can you describe "bug where < is omitted from unicodes" ?

groverlynn:
> can you describe "bug where < is omitted from unicodes" ?

@sunderme  When the unicode character(s) for a symbol contains `<` (here, only several instances with the literal "<", but could also apply to "<⃥" or "<⃒"), inserting the unicode from the symbol list gives wrong character (empty). `<` should be escaped as `&lt;` to get the correct unicode character(s).

sunderme:
please state an exact test case, where the current behavior leads to unwanted behavior.
Your change fails when using "insert as unicode" on "<".
grafik

sunderme:
I would suggest that you take out the change on src/symbolpanel/symbollistmodel.cpp.
Especially as that is something different from the rest.

groverlynn:
Ok... escaped `<` too early. Now it works.

Screenshot 2023-01-26 at 02 17 17 (2)

sunderme:
toHtmlEscaped() seems to be the simpler solution.

src/symbolpanel/symbollistmodel.cpp
```diff
@@ -86,10 +86,10 @@ SymbolItem loadSymbolFromSvg(QString fileName)
 	if (!nl.isEmpty()) {
 		QDomNode n = nl.at(0);
 		QString pkg = n.toElement().attribute("Packages");
-		if (!pkg.isEmpty())
+		if (!pkg.isEmpty() && pkg != "")
 			item.packages = pkg;
 		QString cmdUnicode = n.toElement().attribute("CommandUnicode");
-		if (!cmdUnicode.isEmpty())
+		if (!cmdUnicode.isEmpty() && cmdUnicode!="")
```
sunderme, line 92: `&& cmdUnicode!=""`is redundadnt, please remove

src/symbolpanel/symbollistmodel.cpp
```diff
@@ -86,10 +86,10 @@ SymbolItem loadSymbolFromSvg(QString fileName)
 	if (!nl.isEmpty()) {
 		QDomNode n = nl.at(0);
 		QString pkg = n.toElement().attribute("Packages");
-		if (!pkg.isEmpty())
+		if (!pkg.isEmpty() && pkg != "")
```
sunderme, line 89: `&& pkg!=""`is redundant, please remove

sunderme:
thanks

PR #2877, opened 2023-01-23
Symbols update (.svg files)

groverlynn:
Based on #2876


sunderme:
my guess is that the svg are still generated from control files.
Are they changed ?

sunderme:
okay, they are in the other commit.

sunderme:
thanks

PR #2878, opened 2023-01-23
named placeholder in inline math

octaeder:
This is an update for issue #2719. At a glance: Using $$ may result in wrong coloring of the parser. So @sunderme introduced a space as a placeholder. Named placeholder seem to be more common. @dbitouze suggested maths as placeholder. This PR will implement this suggestion.

### DIFF
--- a/uiconfig.xml
+++ b/uiconfig.xml
@@ -246,7 +246,7 @@
 
 <menu id="main/math" text="&amp;Math">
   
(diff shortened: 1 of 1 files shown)


sunderme:
I stick with the current solution. Closer to how it used to be for years.

octaeder:
ok

PR #2882, opened 2023-01-24
updated cwls

mbertucci47:
New cwl for `expkv-pop` and changed `class-xput` to `class-pagelayout`.

Updated cwls for `babel`, the `expkv` bundle, `tkz-euclide`, and several other small fixes.


sunderme:
Thanks

PR #2889, opened 2023-01-26
rectangular magnifier with shadow

### LINKED ISSUE #2885: Add shadow option for rectangular magnifier
There are two circular and one rectangular shapes for the magnifier to choose from in the config. The circular one can be with or without shadow, but the rectangular one is always without shadow. I want to add the shadow option for the rectangular magnifier. Sreenshot from the POC:

grafik

All who would prefer the rectangular magnifier for practical reasons but find the shadow design of the circular one more attractive.

### Any Other info
Questions below

octaeder:
This PR resolves #2885. By this the configuration of the magnifier's shape, border, and shadow are fully separated by providing a checkbox for the shadow.  In addition the user can select the shape of a rectangle (formerly called square) with a shadow:

grafik

grafik
#### Backward compatibility
Default is _Circle with shadow_ as before. Old values are mapped like this:
```
Square (Rectangle)    --> Rectangle with shadow
Circle with shadow    --> Circle with shadow
Circle without shadow --> Circle without shadow
```
First case is exceptional because a shadow is added. But this is a very, very small change of the GUI. All other cases make no change.
On the other hand: A combobox with item _Rectangle with shadow_ appended would give a random item sequence, and anyone would ask, "border is not part of the list, so why is shadow?"

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -318,15 +318,14 @@ void PDFMagnifier::reshape()
 	if (!globalConfig || globalConfig->magnifierShape == oldshape) return;
 
 	switch (globalConfig->magnifierShape) {
-    case 2: [[clang::fallthrough]];
 	case 1: { //circular
-	        int side = qMin(width(), height());
+		int side = qMin(width(), height());
 		QRegion maskedRegion(width() / 2 - side / 2, height() / 2 - side / 2, side, side, QRegion::Ellipse);
 		setMask(maskedRegion);
 		break;
 	}
-	default:
-		setMask(QRect(0, 0, width(), height())); //rectangular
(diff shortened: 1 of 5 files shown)


src/pdfviewer/PDFDocument.cpp
```diff
@@ -341,15 +340,23 @@ void PDFMagnifier::setImage(const QPixmap &img, int pageNr)
 	update();
 }
 
-void PDFDraggableTool::drawCircleGradient(QPainter& painter, const QRect& outline, QColor color, int padding)
+void PDFDraggableTool::drawGradient(QPainter& painter, const QRect& outline, QColor color, int padding, int magnifierShape)
```
sunderme, line 343: enum ... magnifierShape

sunderme:
thanks

PR #2890, opened 2023-01-27
fix some units of measurement in config dialog

octaeder:
This PR fixes some issues with units in the config dialog. Looking through the pages I found following items affected from units:

1.  General: Check every 7 days
2. Completion: **Tab Bar Size 100 %** --> % character without space in all other cases
3. Preview: **Auto Update Delay: 300** -->Unit "ms" missing
4. PDF Viewer:
   - Screen Resolution 94 dpi
   - Magnifier Size 400 px
   - **Laser Pointer Size 25** --> Unit "px" missing
   - Highlight Duration 2000 ms
   - Cache Size 512 MB

The three bold values above are fixed:

grafik

grafik

grafik


sunderme:
thanks

PR #2894, opened 2023-01-28
new cwls

mbertucci47:
New cwls for `PixelArtTikz`, `TangramTikz`, `cals`, `class-univie-ling-poster`, `codeanatomy`, and `phoenician`.

Several cwls updated to use new `%file` specifier. Other small fixes.


completion/runcode.cwl
```diff
@@ -53,48 +51,40 @@ inline
...
 \inlnMatlab[server spec]{code%definition}
 \inlnMatlab[server spec]{code%definition}[vbox|inline]
 
-\runR{source file}{output file}
-\runR{source file}{output file}[run|cache]
-\runR[server spec]{source file}{output file}
-\runR[server spec]{source file}{output file}[run|cache]
-\runR{file}{output file}#Si
-\runR[server spec]{file}{output file}#Si
+\runR{source file%file}{output file}
```
sunderme, line 74: shouldn't output file also be declared as file ? This is basically for the syntaxchecker to leave that argument alone

sunderme:
thanks

completion/runcode.cwl
mbertucci47: Yes that's true. I'll add it to the next PR. This `%file` is a good addition since it allows things like being able to Ctrl+Click the arg in `\lstinputlisting{myfile.tex}` without adding the definitions, packages, etc. in myfile.tex to the completion in the original document.

PR #2895, opened 2023-01-28
#2892 QuickStart Wizard - Add Babel language to class options

### LINKED ISSUE #2892: In Wizards → Quick Start, the `⟨language⟩` should be given at the `\documentclass` level
# Describe the feature and the current behavior/state

When a `⟨language⟩` is specified thanks to the `babel` class option in Wizards → Quick Start, this `⟨language⟩` option is given at the package level: `\usepackage[⟨language⟩]{babel}`.

It is advised to give it at the document class level instead: `\documentclass[⟨language⟩]{article}`. Indeed, hence all packages sensible to the language(s) (for instance, `varioref`, `cleveref`, etc.) are aware of the underlying `⟨language⟩` and no need to repeat the `⟨language⟩` at the package level of all these packages.

# Who will benefit with this feature?

Everybody relying on Wizards → Quick Start.

octaeder:
This PR fixes #2892.

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -93,8 +93,10 @@ QuickDocumentDialog::~QuickDocumentDialog()
 
 QString QuickDocumentDialog::getNewDocumentText()
 {
+	bool babel = ui.comboBoxBabel->currentText() != "NONE";
 	QString opt = "";
 	QString tag = QString("\\documentclass[");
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

sunderme:
the `% language support` is rather annoying.

sunderme:
the produced document does not need to explain latex to anyone. it is a starting point for a document.

PR #2896, opened 2023-01-28
add tooltips to items in comboboxes in toolbars

### LINKED ISSUE #2881: Drop-down menu in toolbar should display tooltips/info
Example: Open menu math equations. The items there have the info attribute set in uiconfig.xml. It is presented in the statusbar (for ex. _The gather* environment is an unnumbered gather environment_) when you hover over the items with the cursor.
Add the menu to the math toolbar and repeat the procedure with these items in the toolbar. No info is  shown.

All those who want to use toolbars for quick access to items and need only small hints to decide which item to use.

### Any Other info
No idea how this is done in the menus and how this can be ported to the toolbars.

octaeder:
This PR resolves #2881.
Inserted additional parameter infos to createComboToolButton.
Removed several lines for debugging.

Note: I observed that separator lines from the source menu (s. math/equations between multline* and cases as an example are not copied to the menu in the combobox (s. image). But this is not new behavior.

image

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -786,19 +786,20 @@ void Texstudio::updateToolBarMenu(const QString &menuName)
 					REQUIRE(combo);
 
 					QStringList actionTexts;
+					QStringList actionInfos;
 					QList<QIcon> actionIcons;
 					int defaultIndex = -1;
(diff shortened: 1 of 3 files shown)


sunderme:
thanks

PR #2902, opened 2023-01-28
Reimplement rendering of rectangular gradients

Iksas:
A shadow for rectangular magnifiers has been added in #2889, however the shadow couldn't render correctly because the implementation of rectangular gradients in the `drawGradient` method was incorrect (only a single linear gradient was used).

This PR fixes this issue by reimplementing rectangular gradients in `drawGradient`. Rectangular gradients are now rendered by combining four `QLinearGradient`s with four `QRadialGradient`s, as pictured below.

rectangular-shadow-composition

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -342,6 +342,8 @@ void PDFMagnifier::setImage(const QPixmap &img, int pageNr)
 
 void PDFDraggableTool::drawGradient(QPainter& painter, const QRect& outline, QColor color, int padding, int magnifierShape)
 {
+	painter.save();
+
 	if(magnifierShape == PDFDocumentConfig::Circle) {
 		QRadialGradient gradient(outline.center(), outline.width() / 2.0 , outline.center());
 		color.setAlpha(0);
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

sunderme:
reverted as it fails to build on qt5

Iksas:
Qt5 is broken on my machine, so I can't test. I won't have time to fix this until later next month, unfortunately.

octaeder:
I think his is a matter of taste. I wouldn't call it incorrect. While you are trying to draw a symmetric shadow, my intention was to have an asymmetric shadow as if light source would be closer to upper left corner. At the top the gradient lights up the shadow.

image

Iksas:
Thank you for clarifying. My only issue with the above screenshot is that only the top edge has a gradient, while the left edge doesn't have a gradient at all. This looks unnatural to me.

I agree that asymmetric shadows are nicer, and have changed the shadow accordingly in #2904.

octaeder:
As can be seen in the cut out image above the left edge also shows fading before reaching the top. By symmetry this is the same on the right (yes, it's only a gradient south to north). But your latest try in #2904 looks smooth.

PR #2904, opened 2023-01-28
Make rectangular shadows work with Qt5

Iksas:
This PR tries to make the changes in #2902 work with Qt5 by removing calls to the `QPainterPath::clear()` method.

Additionally, a 3D-effect has been added to the rectangular shadow:
rect-3d


sunderme:
thanks

PR #2914, opened 2023-02-01
fix unintended loss of math left/right commands

octaeder:
I checked against source of txs 4.4.1 that there are no other unintended deletions in uiconfig.xml.


sunderme:
thanks

PR #2917, opened 2023-02-02
update unofficial latex2e manual (version January 2023)

octaeder:
Source: [https://ctan.org/pkg/latex2e-help-texinfo](https://ctan.org/pkg/latex2e-help-texinfo)
Extracted from this package:
- latex2e.html
- latex2e.css (unchanged)
- latex2e-figures/

added image folder since the image list.png is used (and in txs 4.4.1 missing) in section *This shows the horizontal and vertical distances*.

Note: As a reference s. #2271


muzimuzhi:
The unofficial latex2e manual also has an online version https://latexref.xyz/. Consider adding it to the menu?

sunderme:
it is not enough just to dump the files into the repository (especially newly added images)
**All** installers need to be adapted that the images are copied ...

sunderme:
I will take a look after the next release

octaeder:
I assume there are reasons (law) speaking against using the internet link. So this is not an option, right?

sunderme:
apart from linking to things outside our control, that manual has been part of txs for years and internet may not always be freely available.

octaeder:
When the image is missing then behavior should be the same as before: Replacement text *latex2e-figures/list* is shown. So the latest document could be rolled out and installers could be changed later.

sunderme:
done

Issue #2212, opened 2022-04-04
What happened to 4.2.3


BeerStud:
### Environment
- TeXstudio: 4.2.2
- Qt: 6.2.3
- OS: Windows 10 PRO
- TeX distribution: Miktex

### Description
The download page lists 4.2.2 as the newest Windows version to download. However, the changelog lists Texstudio 4.2.3. So, what happened to 4.2.3?

sunderme:
4.2.3 is not yet released

PR #2918, opened 2023-02-02
Add Link to Release Notes to About Dialog

octaeder:
This PR adds a link to the About dialog that opens the (local) file CHANGELOG.txt.
This gives quick access to the changes in the current version (and most people won't find the file in install folder).
Disadvantage of Change Log from [https://www.texstudio.org/](https://www.texstudio.org/) is that the latest version gives a preview to the next version which is irritating (s #2212).

### DIFF
--- a/src/aboutdialog.cpp
+++ b/src/aboutdialog.cpp
@@ -11,13 +11,18 @@
 
 #include "aboutdialog.h"
 #include "utilsVersion.h"
+#include "utilsSystem.h"
 
 AboutDialog::AboutDialog(QWidget *parent)
 	: QDialog(parent)
 {
 	ui.setupUi(this);
 	ui.textBrowser->setOpenExternalLinks(true);
-    ui.textBrowser->setHtml(QString("<b>%1 %2</b> (git %3)").arg(TEXSTUDIO,TXSVERSION,TEXSTUDIO_GIT_REVISION ? TEXSTUDIO_GIT_REVISION : "n/a") + "<br>" +
+	ui.textBrowser->setOpenLinks(false);
+	QString logfile = findResourceFile("CHANGELOG.txt");
+	if (logfile != "") logfile = " <a href=\"file:///" + logfile + "\">" + tr("Release Notes") + "</a>";
+	connect(ui.textBrowser, SIGNAL(anchorClicked(QUrl)), this, SLOT(anchorClicked(QUrl)));
+	ui.textBrowser->setHtml(QString("<b>%1 %2</b> (git %3)").arg(TEXSTUDIO,TXSVERSION,TEXSTUDIO_GIT_REVISION ? TEXSTUDIO_GIT_REVISION : "n/a") + logfile + "<br>" +
                             tr("Using Qt Version %1, compiled with Qt %2 %3").arg(qVersion(),QT_VERSION_STR,COMPILED_DEBUG_OR_RELEASE) + "<br><br>" +
 	                        "Copyright (c)<br>" +
 	                        TEXSTUDIO + ": Benito van der Zander, Jan Sundermeyer, Daniel Braun, Tim Hoffmann<br>" +
@@ -58,4 +63,9 @@ void AboutDialog::largeLogo()
 	dlg->layout()->addWidget(label);
 	dlg->setWindowTitle("TeXstudio");
 	dlg->exec();
+}
+
+void AboutDialog::anchorClicked(QUrl url)
+{
+	QDesktopServices::openUrl(url);
 }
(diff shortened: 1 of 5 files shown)


sunderme:
not sure that I want that there.
Probably better to incorporate the changelog into the manual

octaeder:
Maybe the local link could be added to About dialog. I can look at this.
But adding the link to the manual may make it necessary to use the internet link what would imply above disadvantage.

muzimuzhi:
> But adding the link to the manual may make it necessary to use the internet link what would imply above disadvantage.

@octaeder Manual is also available locally, for example on macOS menu Help - User Manual... will open
```
file:///Applications/texstudio.app/Contents/Resources/html/getting_started.html
```

octaeder:
Firefox Browser, Thunderbird, and Github Desktop are examples which all show the links to the release notes in the about dialog. So I've made it like this (the file is searched to get the right link):

image

Nonetheless it makes sense to also have a link in the manual that leads to the release notes.

src/aboutdialog.cpp
```diff
@@ -11,13 +11,18 @@
...
 #include "utilsVersion.h"
+#include "utilsSystem.h"
 
 AboutDialog::AboutDialog(QWidget *parent)
 	: QDialog(parent)
 {
 	ui.setupUi(this);
 	ui.textBrowser->setOpenExternalLinks(true);
-    ui.textBrowser->setHtml(QString("<b>%1 %2</b> (git %3)").arg(TEXSTUDIO,TXSVERSION,TEXSTUDIO_GIT_REVISION ? TEXSTUDIO_GIT_REVISION : "n/a") + "<br>" +
+	ui.textBrowser->setOpenLinks(false);
```
sunderme, line 21: any reason why not to do it like the txs link, some lines further down ?
```
tr("Project home site:") + " <a href=\"https://texstudio.org/\">https://texstudio.org/</a><br><br>" +
```

octaeder: You mean why I use setOpenLinks(false) and use the signal anchorClicked? Well, I tried the simple approach. But after many tries I ended up with reading the [qt docs](https://doc.qt.io/qt-6/qtextbrowser.html#openExternalLinks-prop). Reason: Links starting with https are considered external links, while those starting with file are internal links. This means that the textBrowser itself opens the file as a html page:

image

You can test this easily by setting lines 21 (setOpenLinks) and 24 (connect) as comments (image shows result).

Even if changelog.txt where correct html, I think no one want's to read the news in such a small window.

Btw: There is a context menu on the image that opens the image in a larger version. I would suggest trying to open the larger version when clicking on the image without using a context menu. What do u think? Maybe in a second PR.

sunderme:
The point is that the current about dialog already has a link to an external website without the hustle of "anchorClicked(QUrl url);". So, there is a simpler solution.

octaeder:
[QDesktopServices Class](https://doc.qt.io/qt-6/qdesktopservices.html)
The [openUrl](https://doc.qt.io/qt-6/qdesktopservices.html#openUrl)() function is used to open files located at arbitrary URLs in **external applications**. For URLs that correspond to resources on the local filing system (where the URL scheme is "file"), a suitable application will be used to open the file; otherwise, a web browser will be used to fetch and display the file.

[QTextBrowser Class](https://doc.qt.io/qt-6/qtextbrowser.html)
openExternalLinks : bool
Specifies whether [QTextBrowser](https://doc.qt.io/qt-6/qtextbrowser.html) should automatically open links to external sources using [QDesktopServices::openUrl](https://doc.qt.io/qt-6/qdesktopservices.html#openUrl)() instead of emitting the [anchorClicked](https://doc.qt.io/qt-6/qtextbrowser.html#anchorClicked) signal. Links are considered external if their scheme is neither file or qrc.

openLinks : bool
This property specifies whether [QTextBrowser](https://doc.qt.io/qt-6/qtextbrowser.html) should automatically open links the user tries to activate by mouse or keyboard.

With this in mind can you proof your conjecture?

It works with https://texstudio.org because in this case QTextBrowser uses openUrl() by default.
But this is not true for the file link. It is opened by the QTextBrowser itself (s. openLink, s. image above).
My code changes behavior of the QTextBrowser to use openUrl().

To proof your conjecture a working example with a file link would suffice.

sunderme:
qt accepts markdown and then it looks fine.
Implemented.

sunderme:
this also allows putting it into the manual which in term is also present online in a much nicer presentation.
by the way: CHANGELOG was moved to `utilities/manual/source/CHANGELOG.md`

PR #2926, opened 2023-02-04
A few more fixes to get Haiku build on par (these have been sitting a…

Begasus:
…round for a while)

### DIFF
--- a/src/bidiextender.cpp
+++ b/src/bidiextender.cpp
@@ -4,7 +4,7 @@
 
 //mostly taken from biditexmaker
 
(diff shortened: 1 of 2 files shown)


Begasus:
With the texlive packages available for Haiku I've gone ahead to add the latest release for texstudio (4.5.1 atm), thought about adding a small part of the patch I'm using for Haiku, latest recipe for this can be found at: https://github.com/haikuports/haikuports/tree/master/app-office/texstudio

sunderme:
qmake is deprecated. it does not work anymore with qt6.

Begasus:
Yes saw it when the build started, on the next version bump I'll look into using cmake for the build. (probably some other patches will be involved then, but I'll deal with that when it comes around), thanks!

PR #2928, opened 2023-02-05
changelog: add links for faster reference

octaeder:
Now that the changelog is markup we can handle links to the resources.


octaeder:
Please make html

sunderme:
thanks

PR #2929, opened 2023-02-05
Replace old icons

geolta:
I have replaced a few old icons in the `images` folder and removed the `images/qcodeedit` folder since it is not used anywhere (please double check this since I'm not that experienced with the code and only did a full-text-search across the project). Additionally, the icons in the tables in the manual are now the default colibre icons.

In the `images` folder are a few icons where I think, it would make sense to move them to `images-ng` (but I didn't do that in this PR).


sunderme:
looks okay, but until now, colibri were only used when activated in options ?

geolta:
No, they were enabled by default.

sunderme:
yes, active by default but can be turned off. The new changes do not offer that possibility?

sunderme:
as communication is delayed and several changes are mixed together, let's give a summary

I would suggest splitting the PR into several parts
1) changing the icons in editing.md is fine and will get accepted
2)Colibre icons for insertgraphics.ui/spellergraphics.ui are okay. It looks like it is cut to the right a little, but okay.
As there is no mechanism to select icon themes, this is fine.
3)images/...png are removed and replaced with Colibre svg files:
a) the images folder are legacy images, images-ng contains svg and that should remain this way.
b) as it is possible to still use the old icon theme, this should be true for the toc icons as well. 
So, I suggest moving the new ones to images-ng/colibre and have the old ones for other icon themes. Possibly the icon selection function needs to be adapted.

geolta:
All right, should meet your wishes now :)

sunderme:
looks good, thanks

PR #2936, opened 2023-02-07
add simple navigation to About dialog

octaeder:
Clicking on the changelog link loads a second page. This PR allows to come back to the first page by clicking the "Back" button. You can still close the dialog immediately with the ESC key.

Btw: Using QTextBrowser::backward() to load the first page is not possible (no history) because the first page is set by setHtml() and not loaded via setSource(). So setHtml() is used again.

### DIFF
--- a/src/aboutdialog.cpp
+++ b/src/aboutdialog.cpp
@@ -26,8 +26,7 @@ AboutDialog::AboutDialog(QWidget *parent)
         }
         changelogPath="file://"+changelogPath;
     }
-	ui.textBrowser->setOpenExternalLinks(true);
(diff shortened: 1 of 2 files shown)


sunderme:
i like the "ok" button better.

octaeder:
so return to first page with "ok"?

sunderme:
close with okay. Why would you go back in the first place. It is not like you browse through it extensively...

octaeder:
it feels a little bit strange when I follow a link and can't go back. Even if it is not used often. But that's the same with the image...

sunderme:
I close this, as I prefer the current solution.

PR #2946, opened 2023-02-09
add Special option to Adv. Editor for Overscrolling Last Line to Top

### LINKED ISSUE #2944: Editor vertical overscroll
Normally scrolling down in a txs editor is possible until the editor shows the last line at the bottom (*no* overscroll). With Adv. Editor option "vertical overscroll" set one can even scroll down until the last line reaches mid (*half*) of the editor. Notepad++ for example allows scrolling down until the last line reaches *top* of the editor:

image (image from a proof of concept (poc) build of txs)

Could the behavior of txs with overscroll option be changed to be the same as with notepad++, or could it be added as a third option (overscroll: *no*/*half*/*top*)?

Anyone who likes the behavior of notepad++ (and maybe other editors).

### Any Other info
Currently none

octaeder:
This PR closes #2944 by adding a new Option to Adv. Editor Config. This option allows the editor to scroll down over the last line until the line reaches the top of the editor. Please check the option text and the tooltip.

Please make CHANGELOG.html

### Description of Commits
1. reorder some Special options as depicted in following image (row numbers to the left):

    image

    The final result to the right shows 3 options concerned with parentheses/brackets, two with scrolling and the rest with mouse actions (including option Triple-Click Selection below lower edge of the image). Not visible options to the right are not touched.
2. Since the new option should be in row 6 rows 3 to 6 are renumbered from 2 to 5.
3. Add new option and optimize vertical scrollbar settings (discussed below):

    image

    The new option is not set by default. This together with the usage of two checkboxes ensures that the old setting can be reproduced exactly as it was. Option *Vertical Overscroll* shows following tooltip:
    image
    Be aware that the two options concerning monitoring of open files use two related checkboxes in a similar way as it is here.

### Enhancing Vertical Scroll Bar Settings
The old formulae used have two drawbacks: 1. Page scrolling may omit rows and 2. Scrolling doesn't exactly stop in the middle of the editor.

1. Page scrolling may omit rows
    Old formula ``setPageStep(qCeil(1.* viewportSize.height() / ls))`` applies qCeil to the number of lines visible in the editor. This number is not an integer when the lower edge of the editor covers some fraction of a line. Thus this line is part of the pageStep and the line after it will be the line presented at the top after scrolling one page.
    Example. In this case the editor shows almost nothing of line 14:

    image

    Now scroll down one page with the old formula:

    image

    No one (except me) can tell what is written in line 14. And this will happen every time scrolling down one page. The new formula ``setPageStep(qFloor(viewportSize.height() / ls + 0.3))`` guarantees that the covered line will be presented again at top after scrolling as long as more than 30% of the line are covered by the edge:

    image

    Now we know the contents of line 14. The following image shows the exact limit when line 14 will not be presented again at top:

    image

2. Scrolling doesn't exactly stop in the middle of the editor
    In this case (the dashed line divides the editor into equal parts)

    image

    overscrolling should show the last 7 lines. This is not the case when using the old formula

    image

    but with the new one:

    image

    When the number of visible lines is odd then there is one line in the middle of the editor. So overscrolling should show the last line at this place. But with the old formula this place is empty, but not with the new one. I leave it to the reader to check this.

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -2227,9 +2227,9 @@ Then you can select a new shortcut by one of the following ways:
                    </widget>
                   </item>
                   <item row="7" column="0" colspan="2">
-                   <widget class="QCheckBox" name="checkOverwriteOpeningBracketFollowedByPlaceholder">
+                   <widget class="QCheckBox" name="checkBoxSmoothScrolling">
                     <property name="text">
-                     <string>Overwrite Opening Bracket Followed by a Placeholder</string>
+                     <string>Smooth Scrolling</string>
                     </property>
                    </widget>
                   </item>
@@ -2241,9 +2241,9 @@ Then you can select a new shortcut by one of the following ways:
                    </widget>
                   </item>
                   <item row="4" column="0">
-                   <widget class="QCheckBox" name="checkBoxMouseWheelZoom">
+                   <widget class="QCheckBox" name="checkOverwriteOpeningBracketFollowedByPlaceholder">
                     <property name="text">
-                     <string>Mouse Wheel Zoom</string>
+                     <string>Overwrite Opening Bracket Followed by a Placeholder</string>
                     </property>
                    </widget>
                   </item>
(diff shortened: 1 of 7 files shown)


sunderme:
Looking at the code, it is not worth to have an extra option  for "overscrolling to the top"
Just take that as new behaviour for overscrolling.

octaeder:
new PR follows

PR #2948, opened 2023-02-10
updated cwls

mbertucci47:
New cwl for `sacsymb`. Updated cwls for `ProfLycee`, `ResolSysteme`, `class-altacv`, `optex`, and `tcolorbox` and its libraries.


sunderme:
thanks

PR #2949, opened 2023-02-11
Overscroll2top

### LINKED ISSUE #2944: Editor vertical overscroll

octaeder:
This PR closes #2944 by changing interpretation of option "Vertical Overscroll" in Adv. Editor config. This will allow the editor to scroll down over the last line until the line reaches the top of the editor. The old behavior of stopping in the middle of the editor is dropped as discussed in #2946.

Please make CHANGELOG.html

### Description of Commits
1. reorder some Special options as depicted in following image (row numbers to the left) to fit better in terms of content:

    image

    The final result to the right shows 3 options concerned with parentheses/brackets, two with scrolling and the rest with mouse actions (including option Triple-Click Selection below lower edge of the image). Not visible options to the right are not touched.
2. change behavior of the option.
3. changelog

### Enhancing Vertical Scroll Bar Settings
The old formulae used have two drawbacks: 1. Page scrolling may conceal rows and 2. Scrolling doesn't "exactly" stop in the middle of the editor (this is not discussed here since this is no longer of relevance).

#### Page scrolling may omit rows

Old formula ``setPageStep(qCeil(1.* viewportSize.height() / ls))`` applies qCeil to the number of lines visible in the editor. This number is not an integer when the lower edge of the editor covers some fraction of a line. Thus this line is part of the pageStep and the line after it will be the line presented at the top after scrolling one page.

Example. In this case the editor shows almost nothing of line 14:

image

Now scroll down one page with the old formula:

image

No one (except me) can tell what is written in line 14. And this will happen every time scrolling down one page. The new formula ``setPageStep(qFloor(viewport()->size().height() / ls))`` guarantees that the covered line will be presented again at top after scrolling:

image

Now we know the contents of line 14.

### DIFF
--- a/src/qcodeedit/lib/qeditor.cpp
+++ b/src/qcodeedit/lib/qeditor.cpp
@@ -3072,14 +3072,17 @@ void QEditor::selectOccurence(bool backward, bool keepMirrors, bool all)
 void QEditor::setVerticalScrollBarMaximum()
 {
 	if (!m_doc) return;
-	const QSize viewportSize = viewport()->size();
-	int viewportHeight = viewportSize.height();
-	if (flag(VerticalOverScroll))
-		viewportHeight /= 2;
-    const qreal ls = m_doc->getLineSpacing();
+	const qreal ls = m_doc->getLineSpacing();
+	const int totalLines = qRound(m_doc->height() / ls);
+	const int viewLines = qFloor(viewport()->size().height() / ls);
 	QScrollBar* vsb = verticalScrollBar();
-    vsb->setMaximum(qMax(0., 1. + (m_doc->height() - viewportHeight) / ls));
-    vsb->setPageStep(qCeil(1.* viewportSize.height() / ls));
+
+	if (flag(VerticalOverScroll)) {
+		vsb->setMaximum(totalLines - 1);    // allow last line scroll to top
+	} else {
+		vsb->setMaximum(qMax(0, totalLines - viewLines));    // disallow overscroll
+	}
+    vsb->setPageStep(viewLines);
 }
 
 /*!
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -2227,9 +2227,9 @@ Then you can select a new shortcut by one of the following ways:
                    </widget>
                   </item>
                   <item row="7" column="0" colspan="2">
-                   <widget class="QCheckBox" name="checkOverwriteOpeningBracketFollowedByPlaceholder">
+                   <widget class="QCheckBox" name="checkBoxSmoothScrolling">
                     <property name="text">
-                     <string>Overwrite Opening Bracket Followed by a Placeholder</string>
+                     <string>Smooth Scrolling</string>
                     </property>
                    </widget>
                   </item>
@@ -2241,9 +2241,9 @@ Then you can select a new shortcut by one of the following ways:
                    </widget>
                   </item>
                   <item row="4" column="0">
-                   <widget class="QCheckBox" name="checkBoxMouseWheelZoom">
+                   <widget class="QCheckBox" name="checkOverwriteOpeningBracketFollowedByPlaceholder">
                     <property name="text">
-                     <string>Mouse Wheel Zoom</string>
+                     <string>Overwrite Opening Bracket Followed by a Placeholder</string>
                     </property>
                    </widget>
                   </item>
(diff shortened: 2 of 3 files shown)


src/qcodeedit/lib/qeditor.cpp
```diff
@@ -3072,14 +3072,17 @@ void QEditor::selectOccurence(bool backward, bool keepMirrors, bool all)
 void QEditor::setVerticalScrollBarMaximum()
 {
 	if (!m_doc) return;
-	const QSize viewportSize = viewport()->size();
-	int viewportHeight = viewportSize.height();
-	if (flag(VerticalOverScroll))
-		viewportHeight /= 2;
-    const qreal ls = m_doc->getLineSpacing();
+	const qreal ls = m_doc->getLineSpacing();
+	const int totalLines = qRound(m_doc->height() / ls);
```
sunderme, line 3076: it should probably not be qRound.
qCeil ?

sunderme: definitely different from old, so please use that (or functional **identical** )

src/qcodeedit/lib/qeditor.cpp
```diff
@@ -3072,14 +3072,17 @@ void QEditor::selectOccurence(bool backward, bool keepMirrors, bool all)
...
 {
 	if (!m_doc) return;
-	const QSize viewportSize = viewport()->size();
-	int viewportHeight = viewportSize.height();
-	if (flag(VerticalOverScroll))
-		viewportHeight /= 2;
-    const qreal ls = m_doc->getLineSpacing();
+	const qreal ls = m_doc->getLineSpacing();
+	const int totalLines = qRound(m_doc->height() / ls);
+	const int viewLines = qFloor(viewport()->size().height() / ls);
```
sunderme, line 3077: different from old code ?
And also dangerous to accidentally take integer calculus
(that is why the calculations start with 1. (1.0) to force the compiler to calculate float)

src/qcodeedit/lib/qeditor.cpp
octaeder: from a numerical point of view height() = ls * totalLines. Due to rounding errors calculating totalLines from this equation will likely not give an exact integer but a real that is very close to an integer. qRound would fix this. Adding 0.5 and dropping the fractional part (convertion to int) should do the same. But original solution adds 1. for some reason I don't understand. I revert this.

src/qcodeedit/lib/qeditor.cpp
octaeder: To  be consequent ``1. + (m_doc->height() - viewportHeight) / ls`` should
be changed to ``1. + 1. * (m_doc->height() - viewportHeight) / ls`` . Really?
Variable ``ls`` is not an integer so no integer division happens.

sunderme: Okay, I have looked it up, this should be fine.
But I have seen cases, where it makes a difference.

sunderme:
thanks

src/qcodeedit/lib/qeditor.cpp
octaeder: Observations:
1. ``m_doc->height()`` and ``ls`` have fractions. But interesting enough the result of ``m_doc->height() / ls`` is a real that represents an exact integer. Let x be the result, then the following shows that x, qFloor(x), and qCeil(x) are the same (i.e. x is an integer) even with more than 786 thousand lines. This is independent of ``ls`` being 1, 1.25, or 17.4375.

    Examples:
    ```
    ls: 10.75  height() : 494.5  height()/ls : 46 46 46

    ls: 53.6875  height() : 2469.63  height()/ls : 46 46 46

    ls: 1.34375  height() : 1.05677e+06  height()/ls : 786433 786433 786433

    ls: 17.4375  height() : 1.37134e+07  height()/ls : 786433 786433 786433
    ```
    The exact calculation of the last line for overscrolling to the top ``totalLines = qRound(m_doc->height() / ls)`` does not rely totally on this observation by using qRound.

3. ``viewport()->size().height()`` is an integer (pixels), but dividing it by ``ls`` gives a real:
    ```
    ls: 17.4375  height(): 479  height() / ls: 27.4695
    ```
    The integer part of this result is used for pageStep.

PR #2950, opened 2023-02-12
Add synctex.gz to CleanDialog::defaultExtensions

Alexander-Wilms:


muzimuzhi:
I once opened a similar PR (without sorting), see #934. It's closed since one of the maintainers suggested

> We intentionally did not include synctex.gz. As opposed to other auxiliary files it is needed beyond the compilation process. As soon as you delete it, syncing between pdf and tex does not work anymore in TeXstudio. Therefore, I prefer to not delete this by default.

Alexander-Wilms:
I figured one would delete the auxiliary file once one is done with a document. I deleted the synctex.gz file for a document opened in TeXstudio and syncing still worked. So I assume it's loaded into memory after the compilation is finished. Only when you close and reopen TeXstudio the menu option to sync is no longer available. But it's not a big deal one way or another.

sunderme:
Noting has changed for Tim's statement. 
That "as long as txs is open" is also not completely true. Synctex is rather loaded when opening a new pdf.
So, switching between documents, possibly opening manuals etc, all invalidate the loaded synctex.

PR #2952, opened 2023-02-12
new and updated cwls

mbertucci47:
New cwls for `class-g-brief`, `class-g-brief2`, `europs`, and `lualinalg`. Updated cwls for `ResolSysteme`, `mismath`, `pxpic`, and other small fixes.


sunderme:
thanks

PR #2953, opened 2023-02-12
fix a cwl typoe

muzimuzhi:
tcolorbox choice option `/tcb/lowerbox` accepts `ignored`, not `ignore`.

Screenshot from the tcolorbox manual:
image


sunderme:
thanks

Issue #2975, opened 2023-02-24
Reintroduce templates


dbitouze:
With TeXstudio 4.5.2 (git 4.5.1-118-g9c7c37a15), almost all the templates (File → New From Template...) have disappeared: in my case, only the “Article (French) for LuaLaTeX” is still here.

It would be nice to reintroduce the old templates.

sunderme:
works fine here

octaeder:
that' what I see with my Win10 build (TeXstudio 4.5.2 (git 4.5.1-70-gd194111e3)):

![grafik](https://user-images.githubusercontent.com/102688820/221288190-0ced6042-6c4a-4bcf-afbe-e4377a6dcddf.png)

Btw: The text in the middle may need a spacer at the bottom.

muzimuzhi:
fine on macOS (texstudio 4.5.1)
![image](https://user-images.githubusercontent.com/6376638/221292170-12d7f998-fa01-409c-9a25-d2afd4216c31.png)

sunderme:
win build is local cmake build or one from github ?

dbitouze:
Here is what I can see:

![Screenshot_20230224_220821](https://user-images.githubusercontent.com/1032633/221292792-6cc4374e-0996-410a-ba4b-bd49baaede2e.jpg)

octaeder:
> win build is local cmake build or one from github ?

local cmake

sunderme:
@dbitouze also linux local build ?

dbitouze:
> @dbitouze also linux local build ?

I don't see what you mean by “local build”.

sunderme:
a) are you still running linux ?
b) do you build texstudio yourself (with build.sh or directly cmake) or do you download builds from github ?

sunderme:
copy from the source the templates folder to the build folder
(cmake --build . -t install probably works as well, but install txs ususally in /usr/local )

dbitouze:
> a) are you still running linux ? b) do you build texstudio yourself (with build.sh or directly cmake) or do you download builds from github ?

Still running GNU/Linux (Mageia 8), build thanks to `BUILD.sh` (sorry, I added these details afterwards in a previous comment).

octaeder:
![grafik](https://user-images.githubusercontent.com/102688820/221295875-ea9f39d1-fcfa-4b37-9b29-3e2a3ebbd468.png)
copy translation helped

dbitouze:
[...]

PR #2976, opened 2023-02-24
insert spacer below template's meta info

octaeder:
This issue was mentioned in issue https://github.com/texstudio-org/texstudio/issues/2975#issuecomment-1444482797

before fix:
grafik
note: dialog window shrinked for screenshot, so vertical spacing less than in standard dialog

after fix:
grafik

### DIFF
--- a/src/templateselector.ui
+++ b/src/templateselector.ui
@@ -122,6 +122,22 @@
         </property>
        </widget>
       </item>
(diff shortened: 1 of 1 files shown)


sunderme:
the set-up looks deliberate. 
I will not change that.

Issue #2974, opened 2023-02-24
TEXMFHOME not searched by texstudio after homebrew texlive install (after switching to Ventura?)


gpasa:
I was previously on texlive under MacOS Macports, and using texstudio without any problem. I have a bunch of custom packages in ~/Library/texmf and when I build my documents on the command line in terminal everything is fine. I have migrated to homebrew and moved away from MacPorts. Everything works fine from the terminal, but I cannot build my documents from texstudio anymore, as the packages in ~/Library/texmf are not found, although they are found when building from the cli. Does anybody know what the problem is ?

I tried to texhash, but in the terminal it does not check TEXMFHOME (which I set in .profile). Everything seems fine, but from texstudio pdflatex does not find the packages in my home folder ~/Library/texmf. This drives me mad!

I already tried to "Try to set the texmf folder with tlmgr conf texmf TEXMFHOME ~/Library/texmf/" (had to add conf to tlmgr/config first), but still gets ! LaTeX Error: File `mymath.sty' not found.

From the Terminal everything is fine. TEXMFHOME is fine too, and the commands run flawlessly. The .profile is correctly set up. It's only from Texstudio that ~/Library/texmf isn't searched for. For now, I got it work by launching texstudion from the command line in a Terminal. I think that (as samcarter_is_at_topanswers.xyz said, that GUI apps do not have the correct environment for TEXMFHOME, or something related). BTW: I got this problem after upgrading (sic! ) to Ventura (maybe it's related...) – 

I could check that if I start the app by double-clicking its icon or from the Spotlight, there is a problem with the search path. But if I lauch it by double-clicking the executable (after Dispalying the content of the package, and clicking on the executable in MAcOS folder) then the path is initialized correctly and everything is fine. Seems like opening the app as usual does not set the paths correctly. – 

I could figure out that texstudio started from Spotlight or by double-clicking the icon, does not have its environment variables set properly (I think it does not read the environment variables in the same way as when started from a terminal window). In particular TEXMFHOME is not initialized from direct icon launch. But everything is ok from the command line. How can I correct this behavior?

sunderme:
are you sure that there is only *one* latex installation and the same pdflatex is used ?

gpasa:
yes. I simply started a user command "echo $TEXMFHOME", and when launched from texstudio icon launched it displays nothing, but if launched on texstudio started from the cli, then it displays the TEXMFHOME folder.

muzimuzhi:
[...]

PR #2978, opened 2023-02-26
docs: delete a stray word

muzimuzhi:
caught when commenting in https://github.com/texstudio-org/texstudio/issues/2974#issuecomment-1445388199


sunderme:
thanks

PR #2981, opened 2023-03-01
Add search field to menu bar in macOS

Iksas:
The `Help` menu bar item is instantiated with the string `&Help` to enable the Alt+H keyboard shortcut in Windows.

Usually macOS automatically adds a search field to the `Help` menu, but it doesn't in TexStudio because it gets confused about the prefixed ampersand.

This PR removes the ampersand when building for macOS, which enables the menu bar search field:

help_menu_search


sunderme:
thanks

PR #2982, opened 2023-03-02
Stop displaying the pointing hand cursor on top of tools

Iksas:
When dragging a tool (e.g. the magnifier) over a hyperlink, the pointing hand cursor would be displayed on top of the tool until the mouse button was released. This was caused by a missing check in one of the two `PDFWidget::updateCursor` methods.

This PR fixes this issue.


sunderme:
thanks

PR #2987, opened 2023-03-04
add Min/Max/Close buttons to Usermenudialog (macro editor)

octaeder:
so no need to change the window size with the mouse

grafik

This makes reading and editing macros easier. There will be support for Win10 features such as maximum vertical stretching with a double click on the upper or lower border, or special window sizing by dragging the window to the screen borders or screen corners.


sunderme:
thanks

PR #2988, opened 2023-03-05
tooltips for macros in treewidget (Edit Macro dialog)

octaeder:
This PR helps identifying the macro files used. Thus one can figure out easily the pathname of the macro file. This is useful in occasions where you want to edit the macro in a different editor or need to knwo which file you want to upload to the repository. Counting lines to figure out the number used in the filename may not succeed for example after inserting, deleting, or reordering items in the treewidget, or because you not even know where macros are stored. Furthermore the tooltip shows either the repository url for macros just downloaded or "new" for new macros. Thus you can see which macros are not saved.

Examples:
grafik

The 4<sup>th</sup> item is currently saved in the 11<sup>th</sup> file (start counting at 0). This happenes after dragging that item to this place. After closing the dialog renumbering to 3 will appear (and all macros below are renumbered also).

grafik

Be aware that the tooltip information is not stored in the files. After closing and reopening the Edit Macros dialog all these files (downloaded or new) will show a pathname with an ordering number as a tooltip.

### DIFF
--- a/src/macrobrowserui.cpp
+++ b/src/macrobrowserui.cpp
@@ -65,7 +65,7 @@ QList<Macro> MacroBrowserUI::getSelectedMacros()
                 QString macroJson=cache.value(url);
                 if(!macroJson.isEmpty()){
                     Macro m;
-                    m.loadFromText(macroJson);
+                    m.loadFromText(macroJson,url);
                     lst << m;
                 }
             }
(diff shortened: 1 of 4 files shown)


sunderme:
This is a **1** user function which probably just confuses everyone else.
Not sure why I would want to know the file location in a tooltip either.
If I wanted to edit directly the files, `grep/findstr` is your friend to find the correct one.

PR #2994, opened 2023-03-06
let utf8 be default for quick letter

### LINKED ISSUE #2991: Wizards → Quick Letter → Encoding: `utf8` instead of `latin1`
Nowadays, UTF8 is the encoding to use. Hence it would be nice for the default Encoding in Wizards → Quick Letter to be `utf8` instead of `latin1`.
Everyone who is using this wizard

octaeder:
This PR resolves #2991


sunderme:
thanks

PR #3000, opened 2023-03-06
remove label "LaTeX Content" in Edit Macros dialog

### LINKED ISSUE #2999: Remove Label "LaTeX Content" from Edit Macros dialog
As pointed out in #2990 the label is not really suitable in all cases. A simple solution to this issue is removing the label. As a side effect the (script) editor gets a little bit wider. I think that the UI will look cleaner then. Click on image for full size.

Implemented solution includes a frame surrounding the editor widget:
grafik

This first solution had no frame surrounding the editor widget:
grafik

This shows the state before the change:
grafik

octaeder:
This PR resolves #2999

### DIFF
--- a/src/usermenudialog.ui
+++ b/src/usermenudialog.ui
@@ -235,7 +235,7 @@
       </layout>
      </widget>
      <widget class="QWidget" name="verticalLayoutWidget_2">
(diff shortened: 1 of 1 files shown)


sunderme:
Thanks

PR #3001, opened 2023-03-07
fix status bar of Macro Editor behaves weird

octaeder:
This shows the status bar of the latex editor:
grafik

Open the Macro Editor and observe the surprising large distance between column number and mode indicator INSERT in the status bar compared to the latex editor:
grafik

After 10s the indicator is shifted 6pix to the right, then after another 50s shifted back 6px. This behavior is quite distracting (image not in real time):
usermacroStatusbar
Notes: If you open the Macro Editor again you need to edit the macro to restart the timer. Every time something is edited the timer turns back to zero and a shift of the indicator may appear.

Reason is a not visible timer positioned left to the mode indicator that is counting minutes and seconds. After 10 seconds one digit more is needed, after another 50 seconds one digit is removed because the seconds turn back to 0. If you have time to wait 10min, 100min and so on you will even observe larger translations of the indicator to the left.


sunderme:
thanks

PR #3007, opened 2023-03-12
new and updated cwls

mbertucci47:
New cwls for `alchemist` and `class-hitec`. Moved `\pdfinfo` with the other pdftex-specific commands. Other cwl updates and fixes.

Removed `#d`/`#N` spec from many `\renew...` commands. Since they just change the definition of existing commands, treating them like `\newcommand...` only pollutes the completion list.


sunderme:
thanks

PR #3014, opened 2023-03-14
add tabstops to Quick Start dialog

octaeder:
This PR improves usage of the tab key and allows pressing the ok button if you just want to accept current settings.
The focus now is set to the title field because you likely use same settings for most of your documents created with the QuickStart wizard. So you can just enter title and author, and then confirm anything else by pressing enter key.

The tab key now walks line by line from left to right, touches ok and cancel, and finally the tabwidget item (where you can switch tabs with the left and right arrow keys). Thus it's easy having quick access to any of the fields you want to change.

### Comparison to previous builds
In **4.5.1** and **current master** the Class tabwidget item has focus. It's _not_ possible to confirm all settings by just pressing ok key because the add Class dialog catches this key (note blue framed widget). Tab key cycles to the Author field, then to Doc Class, then it's add button, then to add button for Language, to Size combo, and so on in a random walk.

image

More: One method and one widget were renamed to more suitable names, a comment and an almost useless focus policy were removed. Ok/Cancel button are now constructed in the same way as in the QuickBeamer ui file.

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -59,8 +59,7 @@ QuickDocumentDialog::QuickDocumentDialog(QWidget *parent, const QString &name)
 	setWindowTitle(tr("Quick Start"));
 
 	//Package Tab
-	ui.pushButtonPackages->setFocusPolicy(Qt::NoFocus);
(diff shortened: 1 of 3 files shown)


sunderme:
thanks

PR #3015, opened 2023-03-15
add \maketitle to quickStart document

octaeder:
To take advantage of the user given data for title and author a \maketitle is necessary in the LaTeX document.
The command is added only if a title is given (otherwise maketitle complains).


sunderme:
thanks

PR #3020, opened 2023-03-16
Fix search field in macOS menu bar

Iksas:
#2981 was supposed to add a search field to the macOS menu bar. However, the fix in #2981 was only tested with an old version of Qt and the `BUILD.sh` script, and didn't work with the CD workflow. Sorry about that.

This PR adds the creation of an empty file called `empty.lproj` to the CD workflow, which fixes the search field in the macOS menu bar. The changes introduced in #2981 are no longer needed, and are reverted.


sunderme:
thanks

PR #3022, opened 2023-03-18
rework tooltips, insert code with placeholders for math envs

### LINKED ISSUE #3002: Suggestions of improvements for translatable strings
The following translatable strings:

- >The equation environment centers one equation on the page and places the equation number in the margin.

  isn't quite right. Indeed:

  - the equation is centered only if the `fleqn` option — normally selected through the `\documentclass` command — has not been used,
  - more important, the number isn't in the margin but at the end of the line (or at its beginning if the `leqno` option — normally selected through the `\documentclass` command — has been used).

- > Fraction - \dfrac{}{}”

  could be precised:
  > Big Fraction - \dfrac{}{}”

  in order to make it different from the previous line in the menu:
  > Fraction - \frac{}{}”

- >The cases environment provides a strucure that displays as a brace left to an array of lines with two left-aligned columns.

  could be precised:
  >The cases environment provides a structure that is displayed as a left brace extending vertically across the height of an array of lines left-aligned with two left-aligned columns.

octaeder:
This PR closes #3002. Thanks for discussing to @sunderme and @dbitouze.

### DIFF
--- a/uiconfig.xml
+++ b/uiconfig.xml
@@ -259,24 +259,26 @@
   <insert id="array" text="\begin{array}" insert="\begin{array}{%&lt;columns%>}%\%&lt;content%|%>%\\end{array}" info="\begin{array}{col1col2...coln}%ncolumn 1 entry &amp; column 2 entry ... &amp; column n entry \\%n...%n\end{array}"/>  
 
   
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

Issue #3025, opened 2023-03-19
add icons and separators from submenu to toolbar combobox drop-down list


octaeder:
### Describe the feature and the current behavior/state
As an example add the Commands menu from main menu Tools to the Custom Toolbar in the Config dialog (page Toolbars). Then compare the to menus:

![grafik](https://user-images.githubusercontent.com/102688820/226180742-b8821888-6384-473e-9ae1-94265d7a0df0.png)

All icons and separators are missing.

### Who will benefit with this feature?
All who use such a configuration.

### Any Other info
Work in progress

dbitouze:
I would have added such a separator in the Math → Equation sub menu as well, between numbered and unnumbered environments.

octaeder:
just to be clear, I didn't add any separator. I just made sure that the menu added to the combobox looks the same as the source from that it is a copy.

Work on the math menu is still in progress. I'll implement this then.

PR #3026, opened 2023-03-19
copy icons+separators from submenu to toolbar combobox drop-down list

octaeder:
This PR enhances copying data from main (sub)menus when the menu is added to a combobox in a toolbar (s. config dialog->Toolbars), and solves #3025.
Now icons for actions and separators from the menu are copied as well.

#### Example
Add menu main/tools/commands to the custom toolbar in the config dialog. Now compare the main menu (in the middle) with the added menu as in txs version 4.5.1 (to the left) and in txs build with this PR (to the right):

grafik


sunderme:
thanks

PR #3029, opened 2023-03-19
Update advanced.md

### LINKED ISSUE #2989: Macro Editor allows multiple lines for env macros
### Environment
- TeXstudio: 4.5.1

### Expected behavior
As far as I understand an env macro consists solely of one line with the env name.
- in any case this should be clarified in the manual
- if I'm right then the editor should prevent the user from doing anything else as much as possible

### Actual behavior
It seems as one could also add text that prefills the environment like
\begin{info}
   A can tell you that this won't be a good idea.
\end{info}

### How to reproduce
Use the Macro Editor and select env type

octaeder:
regarding env macros. This PR closes #2989.

### DIFF
--- a/utilities/manual/source/advanced.md
+++ b/utilities/manual/source/advanced.md
@@ -221,7 +221,7 @@ inserted as:
 
 ```{note}
 TeXstudio needs that the env-name starts with \"%\", though that
-character is not placed on insertion.
+character is not placed on insertion. Do not provide more data than this.
 ```
 
 ### Script Macros


sunderme:
too imprecise. What does "more data" in this context mean ?
I would not change that text at all. The user will note that multi-line env-names don't make sense.

octaeder:
any data, for ex. one could try to prefill the env with some text by adding some lines. But for this a script macro is needed.

sunderme:
The note describes "environment" macros where the given text is inserted as env-name as argument for `\begin` and `\end`. An example is given, that should suffice.

octaeder:
thanks for clarification. I close this and #2989

PR #3034, opened 2023-03-19
add tooltips + remove labels in math definitions menu (ntheorem)

### LINKED ISSUE #2875: Math definitions needs `ntheorem` loaded with its `standard` option
The new “Math definitions” feature states `ntheorem` package is needed. But most of the users who would use it would be puzzled that it is not enough: without the `standard` option given to  `ntheorem`, the theorems from the standard set provided by `ntheorem` don't work.

I don't know how to warn the users but, if they aren't warned, the feature is misleading IMO.

octaeder:
This PR closes #2875
> My suggestions regarding ntheorem:
    * Add tooltip "Needs \usepackage[standard]{ntheorem}, and
    * remove label statements

tooltip tells how to load the package because special option is needed

### DIFF
--- a/uiconfig.xml
+++ b/uiconfig.xml
@@ -317,14 +317,14 @@
   
   <!-- definitions from \usepackage[standard]{ntheorem} -->
   <menu id="definitions" text="Math &amp;Definitions">
-     <insert id="Corollary" text="env Corollary (ntheorem)" insert="\begin{Corollary}\label{%&lt;key%>}%\%|%\\end{Corollary}"/>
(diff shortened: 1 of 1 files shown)


sunderme:
to make live easier for translator, how about info "Needs \\usepackage[standard]{ntheorem}" instead of "abc needs ..."

octaeder:
> to make live easier for translator, how about info "Needs \usepackage[standard]{ntheorem}" instead of "abc needs ..."

but now PR is merged. So should I change it in a new PR?

sunderme:
if it is fine with you, I can quickly change it here.

PR #3038, opened 2023-03-21
small changes to items in math menu (text, info)

### LINKED ISSUE #3036: continue work on Math menu
@dbitouze @sunderme , there are still things to discuss.

| Item | Info |
|--------|---|
| Inline math mode ``$``...``$`` | The math environment can be used in both paragraph and LR mode |
| less technical tooltip needed | _Write math in the line of text (math mode), same as ``\(``...``\)`` or env math_ |
| LaTeX inline math mode `` \(`` ...`` \)`` | The LaTeX math environment can be used in both paragraph and LR mode_ |
| less technical tooltip needed | _Write math in the line of text (math mode), same as ``$``...``$``_ |
| Display math mode ``\[``...``\]`` | The displaymath environment can be used only in paragraph mode |
| less technical tooltip needed | _Set math apart from text (display math mode), same as env displaymath_ |
| Fraction - ``\frac{}{}`` | inserts ``\frac{%&lt;num%\|%:translatable%>}{%&lt;den%:translatable%>}`` |
|  | In general: do we need translations for placeholders?|
| Fraction - ``\dfrac{}{}`` (1) | \dfrac is an amsmath package shorthand for \displaystyle\frac... |
| suggestion | _\dfrac (amsmath) always uses display math mode style for the fraction_ (2) |
| \begin{array} | \begin{array}{col1col2...coln}%ncolumn 1 entry &amp; column 2 entry _............_ |
| be consistent: _env array_ | do not repeat code: _tabular for math_ |

Footnotes:

(1) @dbitouze suggested in #3002 following change to the menu entry:
>    Fraction - \dfrac{}{}” could be precised: Big Fraction - \dfrac{}{}”

I'm not sure what people expect when reading Big Fraction.
What do you think?

(2) Let me mention that the third entry of the Math menu is "Display math mode ``\[``...``\]``" (s. table above). So the info should make some sense to the user.

octaeder:
This PR closes #3036. Changes:

grafik
(click to enlarge)
green dots mark changed item texts
blue dots mark new tooltips

### DIFF
--- a/uiconfig.xml
+++ b/uiconfig.xml
@@ -246,17 +246,17 @@
 
 <menu id="main/math" text="&amp;Math">
   
-  <insert id="mathmode" text="Inline math mode $...$" insert="$%&lt; %>$" info="The math environment can be used in both paragraph and LR mode" icon="mathmode" shortcut="Ctrl+Shift+M"/>
-  <insert id="latexmathmode" text="LaTeX inline math mode \(...\)" insert="\( %| \)" info="The LaTeX math environment can be used in both paragraph and LR mode"/>
-  <insert id="displaymath" text="Display math mode \[...\]" insert="\[ %| \]" info="The displaymath environment can be used only in paragraph mode" shortcut="Alt+Shift+M"/>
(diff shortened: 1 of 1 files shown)


sunderme:
"Size for opening delimiter" does not really describe the function of `\left`.
"Autosize opening delimiter" ? or similar ?

sunderme:
thanks

PR #3061, opened 2023-04-05
docs: update syntax of subcommands argument modifieres

muzimuzhi:
When using sth like `% !TeX TS-program = txs:///pdflatex{-synctex}`. I saw
image

Related source lines were added no earlier than commit 652bd52f (add some debug possibilities to QSynctex, 2017-10-31)
so I think the corresponding doc is outdated.
https://github.com/texstudio-org/texstudio/blob/22a71e9e2450217073e94773c345fe3c35e0c3a4/src/buildmanager.cpp#L994-L997


sunderme:
thanks

sunderme:
I am surprised that something like `% !TeX TS-program = txs:///pdflatex/{-synctex}` should work ?
It does not work here and it is not described to work that way in the manual either ...

octaeder:
I'm not sure how to use this. I added `% !TeX TS-program = txs:///pdflatex/{-synctex}` as first line and started a compile (lualatex). I  got following message in the message pane:

```
%!TeX program not recognized! (txs:///pdflatex/{-synctex}). Using default.
Process started: "C:\Users\***\AppData\Local\Programs\MiKTeX 2.9\miktex\bin\x64\lualatex.exe"  -synctex=1 -interaction=nonstopmode --shell-escape "texstudio_eUBUiO".tex
```

I would not describe a feature if it is not sure that it works.

muzimuzhi:
@octaeder For me on macOS both
- `% !TeX TS-program = txs:///pdflatex/{-synctex}` and
- `% !TeX program = txs:///pdflatex/{-synctex}`

work fine with "Build & View" but fail with "Compile" with message similar to yours.  Could you check if it's the same on Windows?

octaeder:
I can confirm for win that the error message with "Build" disappears with "Build & View" (but I wouldn't expect this) and I didn't recognize any difference in behavior between TS-program and program. BUT the command is executed without the given option (we see one extra blank).
BTW the option should be -synctex=1.
```
Process started: "C:\Users\***\AppData\Local\Programs\MiKTeX 2.9\miktex\bin\x64\pdflatex.exe"   -interaction=nonstopmode --shell-escape "texstudio_eUBUiO".tex
```

octaeder:
And from the manual it should be `% !TeX TXS-program:bibliography = txs:///biber` as an example so something is missing in `% !TeX TS-program = txs:///pdflatex/{-synctex}`

octaeder:
Now no error message with "Build" when using `% !TeX TXS-program:compile = txs:///pdflatex/{-synctex=1}` (be aware of TXS and :compile). But still no synctex option there.

octaeder:
and if we do as the manual says `% !TeX TXS-program:compile = txs:///pdflatex/{--synctex=1}' (be aware of --) then it works with "Build" and "Build & View":
```
Process started: "C:\Users\***\AppData\Local\Programs\MiKTeX 2.9\miktex\bin\x64\pdflatex.exe"  -synctex=1 -interaction=nonstopmode --shell-escape "texstudio_UFFwqu".tex
```

muzimuzhi:
> BTW the option should be -synctex=1.

@octaeder Manual says ([link](https://texstudio-org.github.io/configuration.html#advanced-configuration-of-the-build-system) to rendered webpage)

https://github.com/texstudio-org/texstudio/blob/503283c68f131d52a4b5faa875b5c699c8fd35d1/utilities/manual/source/configuration.md?plain=1#L491

> Reason for the error message with "Build" is usage of wrong magic comment. It works when using `% !TeX TXS-program:compile = txs:///pdflatex/{-synctex=1}` (be aware of TXS and :compile).

Great! Looks like I used magic comment in a wrong syntax for long time. Do you suggest to add a generic menu for `% !TeX TXS-program:<command> =`?
https://github.com/texstudio-org/texstudio/blob/503283c68f131d52a4b5faa875b5c699c8fd35d1/utilities/manual/source/advanced.md?plain=1#L95-L107

octaeder:
Summary:
If the command `txs:///pdflatex` is defined as `pdflatex.exe"  -synctex=1 -interaction=nonstopmode %.tex` then
* `% !TeX TXS-program:compile = txs:///pdflatex/[-synctex=0]` (be aware of the brackets) changes  option `-synctex=1` present in the definition to `-synctex=0`
* `% !TeX TXS-program:compile = txs:///pdflatex/{-synctex=1}` (be aware of the braces) removes option `-synctex=1` present in the definition, and `% !TeX TXS-program:compile = txs:///pdflatex/{-synctex}` removes any usage of option `-synctex`.
* `% !TeX TXS-program:compile = txs:///pdflatex/{}` removes all options from the command (and it will hang because the file name is missing too).
* '`% !TeX program = pdflatex` changes the default compiler to `txs:///pdflatex`. **This** should be clarified in the manual (ex. _To be precise, it overrides the default compiler (command txs:///compile) with `txs:///pdflatex`. To change options use TXS-program:compile_). Thus "Build" and "Build & View" are affected.

This can be found in the manual.

octaeder:
@sunderme please _make html_, thanks.

octaeder:
@sunderme The LaTeX menu
image
adds `% !TeX TS-program = `. But this is not described in the manual (exepet in `% !BIB TS-program`). Is it missing or should it be `% !TeX program = ` (i.e. without "TS-")? In addition changelog for 2.12.6 says _- fix: remove incorrect warning "Unknown magic comment" for "% !TeX TS-program = "_. TS-program seems to work like program (I tried with `% !TeX TS-program = pdflatex`).

And what do you think about @muzimuzhi's  question _Do you suggest to add a generic menu for % !TeX TXS-program:<command> =?_ The special case for bibliography is there (s. image above).

PR #3066, opened 2023-04-13
add alignedat env to QuickArray wizard

### LINKED ISSUE #2921: Alignat Wizard
Hi,
would it be possible to implement the alignat wizard?
It would help me a lot in writing a system of equations.
Thank you for considering it.

TeXStudioAlignat

octaeder:
This PR resolves #2921. Key information:

- add alignedat environment for QuickArray Wizard (1)
- align table text accordingly to alignment settings given or set by the environment
- in case of array env choose alignment for each column separately
- for array env replace sequences of the same alignment character by *{n}{c} expressions (checkbox)
- show short preview of env statement in window title

(1) since all currently supported envs need math mode I added alignedat, not alignat or alignat*.

### DIFF
--- a/src/arraydialog.cpp
+++ b/src/arraydialog.cpp
@@ -134,7 +134,7 @@ QStringList ArrayDialog::getEnvBeginEndStatements() {
 	else if (env == "alignedat") {
 		begin += QString("{%1}").arg((ncols + 1) / 2);
 	}
-	else if (env.last(1) == "*") {   // mathtools
+	else if (env.endsWith("*")) {   // mathtools
 		int idx = ui.comboAlignment->currentIndex();
 		begin += "[" + QString(ui.comboAlignment->itemData(idx).toChar()) + "]";
 	}
@@ -202,7 +202,7 @@ int ArrayDialog::currentColAlignIndex(int col) {
 			ix = alignList.indexOf(Qt::AlignLeft);
 		}
 	}
-	else if (env.last(1) == "*") {  // mathtools
+	else if (env.endsWith("*")) {  // mathtools
 		ix = ui.comboAlignment->currentIndex();
 	}
 	else {  // amsmath
@@ -219,7 +219,7 @@ void ArrayDialog::slotAlignmentChanged(int ix) {
 		Qt::AlignmentFlag align = alignList[ix];
 		setColAlignment(col, align);
 	}
-	else if (env.last(1) != "*" || env == "alignedat") {}
+	else if (!env.endsWith("*") || env == "alignedat") {}
 	else
 		setColAlignments();
 	setTitle();
--- a/src/arraydialog.cpp
+++ b/src/arraydialog.cpp
@@ -178,7 +178,6 @@ void ArrayDialog::slotPackageChanged() {
 	QString package = ui.comboPackage->currentText();
 	QString currentEnv = ui.comboEnvironment->currentText();
 	int currentEnvIx = ui.comboEnvironment->currentIndex();
-	int currentAlignIx = ui.comboAlignment->currentIndex();
 	setComboEnv(package);
 
 	if (package == "amsmath" || package == "mathtools") {
@@ -291,7 +290,6 @@ void ArrayDialog::newRows(int rows) {
 	int rowsOld = ui.tableWidget->rowCount();
 	int cols = ui.tableWidget->columnCount();
 	ui.tableWidget->setRowCount(rows);
-	int colds = ui.tableWidget->columnCount();
 	if (rows > rowsOld) {
 		for (int col = 0; col < cols; col++) {
 			int ix = currentColAlignIndex(col);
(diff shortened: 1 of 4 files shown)


sunderme:
this will be integrated for 4.5.3 when accepted.

sunderme:
I have some comments:
1) "Squash column definition" is superfluous. Just take a reasonable behavior and stick with it. There is no real benefit in making this adjustable. (kind of adjustable in the first place)
2) I don't like the "Package support", **especially** as the default only contains one entry. I really don't see the necessity to filter here. If need be, remark what package the env belongs to.
3) Last choice of settings is not remembered. That is usually unexpected and here detrimental as one has to go to the same steps again and again. (what env, how many cols, what alignment). A default button may be a quick route to reset these.
4) preview in the title is nice. I would still expect to see alignment information in the column header.

octaeder:
ad 1) Squash option can be disabled in case you want to modify the alignment settings afterwards. Such modifications could be adding vertical lines or using paragraph cells. In such cases unsquashing might be unnecessary work.

ad 2) I see the point, even so some remarks. One tricky thing of the package names is that those of mathtools (the stared ones) look similar to those from amsmath without star. Someone might conclude that this is the same as with envs not having a number. But in this case the star has a completely different meaning. So selecting the package doesn't mix things and keeps the list short. And marking the package for each entry is not my preferred solution. Maybe we can find something better.

ad 3) In this QuickArray dialog most settings can be reproduced easily. Quick Tabular has much more settings but doesn't remember any of them. Overall I'm not convinced that it's likely that same settings are needed in sequence. If I'm wrong someone will create an enhancement request. What could be desirable is a way of setting the alignment for several columns of an array at once (by selecting cells in different columns). But one has to think about how the workflow for this could be (because only one alignment can be selected in the alignment combo box).

ad 4) Why note something in the header that is obvious? QuickTabular is different (s. 1) and shows things (vertical lines) in the header which are not visible in the cells. An example for QuickArray shows this (right, center, left):

image

sunderme:
1) just chose the likely scenario. Squashing (or the reverse) looks like a task for a script.
2) as the number of envs. is small, one list is easier to use and avoids misunderstandings. I assume that users are capable, i.e. they don't randomly select environments ...
3) this is supposed to be quick and easy. Especially the justification if awkward. (I don't say tabular wizard is good/better, that is another story)
4) it is not obvious when the cell is empty ...

octaeder:
First update follows

octaeder:
@sunderme regarding build error, [last()](https://doc.qt.io/qt-6/qstring.html#last) is QT6, is Linux not up to date?

sunderme:
Qt5 (5.12) is still mandatory for txs.

sunderme:
looks okay.
maybe gray out "Column Alignment" when it is not used ? (apparently in matrix)

octaeder:
I suggest not to inhibit changing alignment if it seems that this has no effect. Most important if you add columns then this value is used as default alignment for these columns when you later on choose env array (even if current env is not array). Also you can pre-select the alignment for matrix* and the like (even so this can be done easily after changing the env).

Currently working on storing options. Main problem you may help me with is that I need to store the array of alignment indices (for array columns), i.e. a list of integers. Currently I use these lines (not following immediately each other):

```
QList<int> ArrayDialog::arrayAligns;
configManagerInterface->linkOptionToDialogWidget(&arrayAligns, this);
configManager.registerOption("QuickArray/arrayAlignments", &arrayAligns);
```
I use this because there is no other object like combo boxes of the dialog (so I assume that "this" is the dialog object). Error message:

```
arraydialog.cpp:487:37: error: no matching function for call to
'ConfigManagerInterface::registerOption(const char [27], QList<int>*)'
  487 |   configManager.registerOption("QuickArray/arrayAlignments", &arrayAligns);
```

sunderme:
I don't think that `registerOptions` works here, as then the array dialog needs to exist during start-up/shutdown.
It is probably sufficient just to create the dialog once (with defaults) and reuse it , so that at least it reopens with the last settings.

octaeder:
texstudio.cpp calls a ArrayDialog:: method that does the registerOptions. This is the same as QuickStart does. And I can see that all other options of QuickArray are stored.

```
void Texstudio::readSettings(bool reread)
{
    QuickDocumentDialog::registerOptions(configManager);
    QuickBeamerDialog::registerOptions(configManager);
    ArrayDialog::registerOptions(configManager);
    buildManager.registerOptions(configManager);
```

sunderme:
QList<int> is simply not offered. You can use QList<QVariant> instead.

octaeder:
> QList is simply not offered. You can use QList instead.

??

sunderme:
You saw that I edited my previous comment ?

sunderme:
So, I merge this request as is ?

octaeder:
you may do so. I hope that I can add storing settings in the next five weeks.

PR #3069, opened 2023-04-15
fix issue in changeLog

octaeder:
escape each of the two backslashes

@sunderme Please check [https://texstudio-org.github.io/CHANGELOG.html](https://texstudio-org.github.io/CHANGELOG.html) because there is no backslash present. But current source of two backslashes should show one backslash.
Thanks


muzimuzhi:
> But current source of two backslashes should show one backslash.

It's a [GitHub Flavored Markdown](https://github.github.com/gfm/#backslash-escapes) feature inherited from [CommonMark](https://spec.commonmark.org/0.30/#backslash-escapes):

> Any ASCII punctuation character may be backslash-escaped:
> - _Input_: `` \!\"\#\$\%\&\'\(\)\*\+\,\-\.\/\:\;\<\=\>\?\@\[\\\]\^\_\`\{\|\}\~ ``
> - _Output_: `` <p>!&quot;#$%&amp;'()*+,-./:;&lt;=&gt;?@[\]^_`{|}~</p> ``

PR #3075, opened 2023-04-17
Cwls

mbertucci47:
New cwls for `cleveref-usedon`, `jiazhu`, and `quickreaction`. Several updated cwls.


sunderme:
why is secsty.cwl deleted ?

sunderme:
okay, found the commit 
completion.qrc is incomplete, hence the build fails ...

mbertucci47:
Ah sorry forgot to delete secsty from completion.qrc

PR #3084, opened 2023-04-25
updated cwls

mbertucci47:
New cwls for the `TianQing` Beamer theme, the `hitex` format, and the `se2packages` package. Several updated cwls.


sunderme:
Thanks

PR #3090, opened 2023-04-27
Display error for directory arguments

alichtman:
When running `$ texstudio DIRECTORY_CONTAINING_TEX_FILES`, I get the following error from TeXstudio: 

```
You do not have read permission to the file $DIRECTORY_CONTAINING_TEX_FILES.
```

This PR makes TeXstudio open valid `.tex` files if a directory is passed as an argument, or display an appropriate error if there are none.

image

image

Building + Testing:

```bash
$ cd texstudio
$ mkdir build && cd build
$ cmake .. && cmake --build . --parallel && ./texstudio $ARGUMENTS

```

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -7006,12 +7006,15 @@ void Texstudio::executeCommandLine(const QStringList &args, bool realCmdLine)
 	}
 #endif
 
-	// execute command line
+    	// execute command line
 	foreach (const QString &fileToLoad, filesToLoad) {
 		QFileInfo ftl(fileToLoad);
 		if (fileToLoad != "") {
-			if (ftl.exists()) {
-                if (ftl.suffix() == Session::fileExtension() || ftl.suffix() == "txss") {
+			if (ftl.isDir()) {
+				QMessageBox::warning(this, "", tr("%1 is an invalid TeX project directory.").arg(ftl.absoluteFilePath()));
+				exit(1);
+			} else if (ftl.exists()) {
+				if (ftl.suffix() == Session::fileExtension() || ftl.suffix() == "txss") {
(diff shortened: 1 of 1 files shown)


sunderme:
We can display a warning that folders can't be opened.
In order to open multiple files, you use wildcards.
The suggested change is not acceptable.

alichtman:
Hey @sunderme,

Thanks for your comments.

> We can display a warning that folders can't be opened.

Folders _can_ be opened. Just not all the time. Here's an example to demonstrate:
If I run `$ texstudio /proc`, it makes sense to throw an error since there should be no `.tex` files in `/proc`, and `texstudio` therefore has no idea what to do.

But if I run `$ texstudio $TEX_PROJECT_DIR`, `texstudio` can figure out that I'm trying to work on the `.tex` files inside `$TEX_PROJECT_DIR` and open them for me. 

>  In order to open multiple files, you use wildcards.

Sure, you can open multiple files using a wildcard. Since the wildcard expansion is done in the shell, this is really just passing multiple file arguments to `texstudio`. But this behavior is:

1. not changed by my PR, and
2. irrelevant (since wildcard expansion turns into *file* opening, and my PR is related to *folder* opening).


>  The suggested change is not acceptable.

Can you elaborate on the problem? I don't really understand your argument.

sunderme:
1) it is redundant: folder/*.tex gives exactly the same results
2) it is non-standard: txs does not offer a open folder function, and no other editors open "random" files in that case
3) as it is non-standard, chances are high that the command is a mistake ... and then txs gets flooded with open tex files.

alichtman:
Amended commit to just display an error.

sunderme:
thanks

PR #3096, opened 2023-04-29
Style of XY icon aligned with others

geolta:
Solving the issue with inconsistent styles for the drawing package icons (https://github.com/texstudio-org/texstudio/pull/2645#issuecomment-1346261801).


sunderme:
thanks

PR #3100, opened 2023-05-05
Enlarge icons

groverlynn:
Enlarge icons to 32x32 to get sharp images in high resolution (retina etc.) screens, and also get rid of jagged edges on low resolution screens while main screen is of high resolution.


sunderme:
I am a bit confused how this should make a difference.
SVG is a vector format.

groverlynn:
> I am a bit confused how this should make a difference.
> SVG is a vector format.

SVG is vector, but QT import it as bitmap and 24x24 is too small for retina and also as it's not a power of 2, there are aliasing issues

sunderme:
I can't confirm the claims on windows or linux.
You are talking about OSX ?

sunderme:
please check the latest dev build. (https://github.com/texstudio-org/texstudio/actions/runs/4917970712)
I have deactivated an old OSX work-around concerning icons.

groverlynn:
> please check the latest dev build. (https://github.com/texstudio-org/texstudio/actions/runs/4917970712)
> I have deactivated an old OSX work-around concerning icons.

seems like problem solved.

PR #3102, opened 2023-05-07
add a lorem ipsum generator

derFogel:
I added a lorem ipsum generator to the default text generator to create a simple pseudo text.

### DIFF
--- a/src/randomtextgenerator.cpp
+++ b/src/randomtextgenerator.cpp
@@ -92,9 +92,15 @@ void RandomTextGenerator::generateText()
 		}
 	}
 
+        //----------------------------------generating
+        //---------------------------------------
+        // lorem ipsum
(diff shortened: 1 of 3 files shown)


octaeder:
Packages like blindtext or lipsum seem to be more flexible. Since there commands generate the code at build time it's easy to change parameters as needed. The text they generate can be quite long without repeating the same short text. In case you really need the final text in your tex document you can copy it from an external pdf viewer.

Comparison of the gui change:

image

Notes concerning old version (and unchanged in new version):
1. I think that the space (1) is misleading. The gui looks like there are two groups of radio buttons (one for characters and one for words) and one would conclude that one of the groups (here the second one) is not initialized.
2. Isn't the word (2) Punctuation?

sunderme:
thanks

PR #3120, opened 2023-05-21
remove \address and \signature from scrlttr2 template

mbertucci47:
`\address` and `\signature` are not defined by this class


sunderme:
thanks

PR #3127, opened 2023-05-25
new and updated cwls

mbertucci47:
New cwls for `class-hu-berlin-letter`, `hu-berlin-base`, and `movement-arrows`. Several updated cwls.


sunderme:
thanks

PR #3130, opened 2023-05-28
Update macrobrowserui.cpp

octaeder:
- make "Macro name" translatable
- do not allow edit of table item on double click
- clear fields name & descript. when clicking on folder


sunderme:
thanks

PR #3132, opened 2023-05-31
no horizontal touchpad scroll when fit to text width in pdf viewer

### LINKED ISSUE #1526: Option 'Disable horizontal scrolling for "Fit to Text Width" not working'
### Environment
- TeXstudio: 3.1.1
- Qt: 5.15.2
- OS:MigSur 11.2.2
- TeX distribution: texlive

### Expected behavior
Option 'Disable horizontal scrolling for "Fit to Text Width" should work

### Actual behavior
Option 'Disable horizontal scrolling for "Fit to Text Width" is not working

### How to reproduce
Select the mode "Fit to Text Width" and ascertain verify that horizontal scrolling is allowed, despite the fact that the option  'Disable horizontal scrolling for "Fit to Text Width" is selected

octaeder:
This PR resolves #1526. So when scale option of the pdf viewer is set to _fit to text width_ and option _Disable horizontal scrolling for "Fit to Text Width"_ is set then horizontal scrolling is not possible neither with the touchpad nor with tilting the scroll wheel. The horizontal scrollbar is removed in this case, but you can still scroll horizontally with left/right arrow keys if needed.

image

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -1685,7 +1685,10 @@ void PDFWidget::wheelEvent(QWheelEvent *event)
         double numDegrees = event->angleDelta().x() / 8.0;
         const int degreesPerStep = 15; // for a typical mouse (some may have finer resolution, but that's k with the co
         QScrollBar *scrollBar = getScrollArea()->horizontalScrollBar();
-        if (scrollBar->minimum() < scrollBar->maximum() && !globalConfig->disableHorizontalScrollingForFitToTextWidth) { //if scrollbar visible
+
+        if ( scrollBar->minimum() < scrollBar->maximum() &&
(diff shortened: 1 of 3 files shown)


sunderme:
now horizontal scrolling per mouse wheel seems not to work at all.

octaeder:
"at all" means when fitToTextWidth and the option is set?
Since the option is "Disable horizontal scrolling" I wonder why scrolling should be possible with the mouse wheel.

sunderme:
I zoomed in a little, scrollbar shows up, horizontal scrolling via wheel does not work.

octaeder:
oh, I see.

octaeder:
From my commented code you can see that I tried with a function in PDFDocument that can access actionFit_To_Text_Width, but it didn't work to use it from PDFWidget. So I made this action variable a global one and it works. Hope you can help me how to do better.

Issue #2785, opened 2022-12-19
Show error logs for compile run of previews


yashpalgoyal1304:
### Describe the feature and the current behavior/state

Currently:
* The logs tab only shows the errors from explicit full compiler runs. I suppose that is shows only the readout from the project root directory's logfile
* So, in case there's some error in the currently preview file, the only way other than manual guessing is to do full compile run ...
* which is very inconvenient if the full compile takes huge amount of time due to project size.

Feature:
* So, when there are errors in the compile run of the preview, there should be some way to view its logs.

This will also help when the error is not in the code, but due to the file inclusions etc which the preview compile run misses very often.

### Who will benefit with this feature?
Anyone using following workflow:
* splitting a project in multiple files (for example for each chapter, or for graphics, etc.)
* making sure each file is as intended (both syntactically as well as semantically) by compiling them individually by using "preview" feature

### Any Other info
TXS: 4.3.1
OS: win 10 20h2

octaeder:
as a workaround: search in the folder %tmp% for the name of the log file.

yashpalgoyal1304:
woah, i just did the following mentioned steps to reach the folder, and found that the directory contains following files all mingled in with rest of the contents of temp 😨😱

Files in the 
* `.log`: 1287
* `.tex`: 1282
* `.aux`: 1251
* `.gz`: 923
* `.pdf`: 923
* `.synctex(busy)`: 59


Steps: (other than `%TEMP%`)
* `^ click` the preview to open the file in the editor tab
* go to side panel > structure tab
* right click on the file > Show in Explorer

octaeder:
Select the file name of interest and copy it to the clipboard:

![grafik](https://user-images.githubusercontent.com/102688820/210191142-21d27a15-057a-4130-a51c-ee0fdb397ff9.png)

Press Ctrl+O that opens the file select dialog. Enter %tmp% into the address field and press enter to switch to the temporary files folder:

![grafik](https://user-images.githubusercontent.com/102688820/210191224-327122de-6990-44f8-b2af-0e321f6e548c.png)

Paste the file name from the clipboard into the file name field and select the log file with a mouse click:

![grafik](https://user-images.githubusercontent.com/102688820/210191274-f44902b6-914e-43d2-9de0-1b01dbd140f0.png)

Press Open button and confirme following dialog:

![grafik](https://user-images.githubusercontent.com/102688820/210191316-735e4de3-548b-455f-96dc-d5a9535be005.png)

[...]

PR #3133, opened 2023-05-31
describe loading logs (manual)

octaeder:
follow up of #2785
also reflect change in main menu Idefix


sunderme:
to be honest, I find the text a bit confusing.
It also rather describes a "trick" to load the log instead of core txs functionality (which simply is not there).
From functionality point of view, a clickable link in the message window would probably suffice.
From manual point of view, I would rather move the section into the wiki, if that is okay with you.

octaeder:
can you keep the change in viewing.md? Or should I make a new PR for that...?

sunderme:
I can keep that change.

octaeder:
ok, thanks.

sunderme:
moved to wiki/tricks

PR #3135, opened 2023-06-02
fix editor moving one more line

### LINKED ISSUE #3131: Move Line Up/Down incorrectly moves the line below selected line(s)
### Environment
- TeXstudio: 4.5.2
- Qt: Qt6
- OS: Windows10
- TeX distribution: miktex

### Expected behavior

Select one or multiple lines (from start of line N to start of line N+m), then use Move Line Up/Down to move the selected line(s) up and down. Only those selected lines (line N to line N+m-1 exactly, inclusive at both ends) move, not including line N+m.

### Actual behavior

Line N+m moves along as well. 

If the selection ends at the end of line N+m-1, then line N+m doesn't move along. It seems that the EOL character in line N+m-1 leads to the inclusion of line N+m. This may relate to the way of implementation regarding how many lines are considered.

Popular text editors, such as vscode, behave consistently regardless whether the EOL is included or not.

octaeder:
This PR fixes #3131.

When a cursor is found with a selection then the indices of all lines from selection.startLine to selection.endLine are added to a list of lines being moved. But the last line should only be added if the selection doesn't end in column 0.

### DIFF
--- a/src/latexeditorview.cpp
+++ b/src/latexeditorview.cpp
@@ -928,7 +928,9 @@ QList<QPair<int, int> > LatexEditorView::getSelectedLineBlocks()
 	for (int i=0;i<cursors.length();i++) {
 		if (cursors[i].hasSelection()) {
 			QDocumentSelection sel = cursors[i].selection();
-			for (int l=sel.startLine;l<=sel.endLine;l++)
+			int lastLineOfBlock = sel.endLine;
+			if (cursors[i].selectionEnd().columnNumber()==0) lastLineOfBlock -= 1;
+			for (int l=sel.startLine;l<=lastLineOfBlock;l++)
 				lines << l;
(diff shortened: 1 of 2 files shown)


sunderme:
looks if there is more broken here.
```
1
2
3
4
5
6
7
```
select 3-5 (including 5), move up.
Only the selection is moved, no content.

I have to take a closer look.

octaeder:
works here
image

octaeder:
but doesn't work when you only type the numbers in the editor

sunderme:
the behaviour is a bit inconsistent, sometimes the lines move, sometimes the selection.

jzniu:
This is not fixed. See [](https://github.com/texstudio-org/texstudio/issues/3131#issuecomment-1657240843)

PR #3140, opened 2023-06-03
new and updated cwls

mbertucci47:
New cwls for `postit`, `string-diagrams`, and `tikz2d-fr`. Several updated cwls


sunderme:
thanks

PR #3146, opened 2023-06-07
show current stable version in About Dialog

octaeder:
It's quite common to use About Dialog to figure out whether your version is up to date, or to visit the home page. Starting About Dialog is much faster than starting Config Dialog. Also, moving the link to the home page to the beginning of the text makes it much easier to access.

image


octaeder:
Test "suppress starting Update dialog from About dialog" by simulating current version is 4.5.1. 

Update dialog starts when pressing Check Now:
image

About dialog shows new version but Update dialog isn't started:
image

sunderme:
thanks

PR #3150, opened 2023-06-09
wrong height display in QuickWizard/Geometry

octaeder:
drawing lower margin line is not correct (uses value from top margin), so top and bottom lines are not the same even so height equals 0.

grafik

after fix:

grafik

Hint: The horizontal line is not in the middle of the page because there is a default ratio of 40% for the top margin.


sunderme:
Thanks

PR #3152, opened 2023-06-09
update cwls for new LaTeX

mbertucci47:
Updated cwls with commands from the [new LaTeX format](https://github.com/latex3/latex2e/releases/tag/release-2023-06-01)


sunderme:
thanks

PR #3153, opened 2023-06-09
don't add duplicate packages in QuickStart wizard

octaeder:
With this PR it is not possible to add packages already available in the Packages tab of the Quick Start wizard. Furthermore adding any of "inputenc", "fontenc", "geometry", or "babel" is not accepted. Note: "inputenc" is not used by the dialog and should not be used by the user.

### Test
Package tikz is user added, babel is used on first tab by line edit field language.
grafik

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -556,12 +556,26 @@ void QuickDocumentDialog::addUserOptions()
 
 void QuickDocumentDialog::addUserPackages()
 {
+	QStringList packagesList;
+	QTableWidget *table = ui.tableWidgetPackages;
+	for (int i=0; i < table->rowCount(); ++i) {
+		QTableWidgetItem *itemPkgName = table->item(i,0);
+		packagesList << itemPkgName->text();
+	}
+	packagesList << "inputenc" << "fontenc" << "geometry" << "babel"; // additional packages you shouldn't add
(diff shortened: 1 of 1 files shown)


sunderme:
"'inputenc' is not used by the dialog and should not be used by the user", that is not correct.
It is generally unnecessary to use inputenc, that does not mean that txs prohibits its use, and there are use cases for this.
The wording "package is already in use" is also a bit misleading. "Is already defined" ?

octaeder:
User can add inputenc:

grafik

Package inputenc appears at top:
```latex
\documentclass[ngerman,10pt,a4paper]{article}
\usepackage{inputenc}
\usepackage[T1]{fontenc}
\usepackage[width=15.00cm, height=24.00cm]{geometry}
\usepackage{graphicx}
\usepackage{mathtools}
\usepackage{amssymb}
\usepackage{amsthm}
\usepackage{thmtools}
\usepackage{xcolor}
\usepackage{nameref}
\usepackage{babel}
\usepackage{hyperref}
\usepackage{tikz}
\begin{document}

\end{document}
```

sunderme:
not yet updated ?

sunderme:
from the looks of it `Package %1 is already defined` has been inadvertently redefined again.

sunderme:
thanks

PR #3156, opened 2023-06-10
enhance usage of units in the Quick Start wizard (Geometry tab)

octaeder:
This PR enhances the page geometry configuration that is part of the Quick Start wizard. It addresses following observations.

1. Lengths are given with a number combined with a unit. When you change the unit from cm to mm the number is not adjusted (image). Simply changing the unit means that the length shrinks to one tenth of the original length in this case. Even so the image doesn't change (workaround: change temp. the number with up/down buttons). In fact the image should not change but the number should increase by a factor of 10. This is implemented by the PR.

   grafik

2. You can enter unknown units (like pico meters pm or nano meters nm) or even any text. In this case dialog uses mm internally. Unit dm is not in the list of the combo boxes, but is recognized when you enter it manually. But dm is not known to LaTeX. In fact there is no reason why one should enter units manually (in general necessary unit conversion formulae would miss). The PR therefore sets attribute editable to false for those combo boxes.

### Test
Changing units: 127mm = 12.7cm = 5in (since 25.4*5=125+2=127):

grafik

These values are correctly saved in and loaded from the ini file. The LaTeX code looks fine:
```
Quick\Geometry%20Page%20Width=127
Quick\Geometry%20Page%20Width%20Unit=mm
```
```latex
\usepackage[width=127.00mm, top=5.00cm, bottom=5.00cm]{geometry}
```
```
Quick\Geometry%20Page%20Width=12.7
Quick\Geometry%20Page%20Width%20Unit=cm
```
```latex
\usepackage[width=12.70cm, top=5.00cm, bottom=5.00cm]{geometry}
```
```
Quick\Geometry%20Page%20Width=5
Quick\Geometry%20Page%20Width%20Unit=in
```
```latex
\usepackage[width=5.00in, top=5.00cm, bottom=5.00cm]{geometry}
```

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -62,12 +62,12 @@ QuickDocumentDialog::QuickDocumentDialog(QWidget *parent, const QString &name)
 	connect(ui.tabWidget, SIGNAL(tabBarClicked(int)), SLOT(setPkgTabToolTip(int)));
 
 	//Geometry package
-	connect(ui.comboBoxUnitGeometryPageWidth, SIGNAL(currentTextChanged()), SLOT(geometryUnitsChanged()));
-	connect(ui.comboBoxUnitGeometryPageHeight, SIGNAL(currentTextChanged()), SLOT(geometryUnitsChanged()));
-	connect(ui.comboBoxUnitGeometryMarginLeft, SIGNAL(currentTextChanged()), SLOT(geometryUnitsChanged()));
-	connect(ui.comboBoxUnitGeometryMarginRight, SIGNAL(currentTextChanged()), SLOT(geometryUnitsChanged()));
-	connect(ui.comboBoxUnitGeometryMarginTop, SIGNAL(currentTextChanged()), SLOT(geometryUnitsChanged()));
-	connect(ui.comboBoxUnitGeometryMarginBottom, SIGNAL(currentTextChanged()), SLOT(geometryUnitsChanged()));
+	connect(ui.comboBoxUnitGeometryPageWidth, SIGNAL(currentTextChanged(QString)), SLOT(geometryUnitsChanged(QString)));
+	connect(ui.comboBoxUnitGeometryPageHeight, SIGNAL(currentTextChanged(QString)), SLOT(geometryUnitsChanged(QString)));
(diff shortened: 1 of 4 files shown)


sunderme:
hmmm, what is the use case for this ?
It is not like someone who starts a new document wants to play around with different units.
I would assume a user takes values from previous experiments or from some other info and puts it in.
The change of units then probably represents that the default unit was wrong and another is to be used. 
But changing then the entry would be undesired ...

From a GUI point of view, having the units twice is undesired.
I would rather have a number only to the left and unit to the right, but that is another question.

octaeder:
Your case has a quite simple solution: Choose your unit first. You may have to copy paste your number, but that can be done without thinking: When you notice that changing the unit also changes the value, then undo the change of unit, copy the number to clipboard, change the unit again and past the number. The user will quickly learn that changing a unit will change the number. Since conversion ratios are not close to 1 it's quite unlikely that you used 18cm previously and now you want to use 18in from some source, because 18in is a quite large size. So in this case it's likely that you will need new numbers.

But what do you do when your source states a length in unit "in" or "pt" but you are used to or all your other lengths are already in mm or cm? Then you need a calculator, must know the conversion rule, and must retype the new numbers. This is quite complicated. Another case: You want to do your setup with mouse clicks. Then changing from 100mm to 10cm might be useful so 5 clicks are enough to change value from 10cm to value 15cm. But if you start with 100mm then this would be nearly impossible. Or you want to go on from 15cm to 15.5cm. Here you could change to 150mm, use 5 clicks and you are done, or you change back to 15.5cm in case you prefer cm in your LaTeX document.

In addition: I can remove the suffix from the spin boxes. Please confirm that I should do so.

octaeder:
Note: When units are changed we need the old and the new unit to know how to convert the number. Currently the old value is derived from the suffix of the spin box value. So the suffixes must be replaced by some variables holding the old unit.

sunderme:
From my point of view, use case of quick start/geometry in descending order of probability:
1) enter values with correct unit
2) enter values without changing unit first, change unit then
3) there is no real scenario where one enter inches and expects txs to convert it to something else. Using the mouse to use up/down spin is highly unlikely. 

In any case, the change makes a probable scenario harder to use in favor of a very unlikely one.

octaeder:
new PR #3157

PR #3157, opened 2023-06-11
fix issues with units in the Quick Start wizard (Geometry tab)

octaeder:
This PR enhances the page geometry configuration that is part of the Quick Start wizard.

**Note:** This PR is a cut down of #3156. A detailed discussion can be found there.

The PR addresses following observations:

1. Lengths are given with a number combined with a unit. This unit is also mirrored as a suffix to the number in the spin box. When you alter a unit then the number is not adjusted (s. image below). This means that the length changes. But the picture visualizing the settings doesn't change (workaround: change temp. the number with up/down buttons). This is fixed.

2. The mirrored units are removed by the PR (s. Test below).

   QuickDialogChangingUnitsIssue
   Fig,: Visualization doesn't reflect change of width when altering unit in combo box. Unit is editable (s. cursor mark).

3. You can enter unknown units (like pico meters pm or nano meters nm) or even any text. In this case dialog falls back to mm. Unit dm is not in the list of the combo boxes, but is recognized when you enter it manually. But dm is not known to LaTeX. The PR therefore sets attribute editable to false for those combo boxes. As side effects signal `editTextChanged()` is replaced by `currentTextChanged()` and the names for the unit combo boxes are corrected from `spinBoxUnitGeometryPageWidth` to `comboBoxUnitGeometryPageWidth`, and so on.

4. Code setting checkBoxes without effect is removed. **Update:** Since not totally correct this part is reverted.

### Test
QuickDialogChangingUnits
Fig.: Visualization now reflects change of width when altering unit in combo box (not editable). Suffix units are not present.

These values are correctly saved in and loaded from the ini file. The LaTeX code looks fine:
```
Quick\Geometry%20Page%20Width=15
Quick\Geometry%20Page%20Width%20Unit=mm
```
```latex
\usepackage[width=15mm, height=23cm]{geometry}
```
```
Quick\Geometry%20Page%20Width=15
Quick\Geometry%20Page%20Width%20Unit=cm
```
```latex
\usepackage[width=15cm, height=23cm]{geometry}
```

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -375,6 +375,12 @@ void QuickDocumentDialog::geometryUnitsChanged()
 {
 	//update all units (easier than just the changed one, slower, but need probably less memory)
 	// it was decided not to recalculate number part (s. https://github.com/texstudio-org/texstudio/pull/3156) and to not use suffix in spinBoxGeometry...
+	if (sender() == ui.comboBoxUnitGeometryPageWidth) ui.checkBoxGeometryPageWidth->setChecked(true);
+	else if (sender() == ui.comboBoxUnitGeometryPageHeight) ui.checkBoxGeometryPageHeight->setChecked(true);
+	else if (sender() == ui.comboBoxUnitGeometryMarginRight) ui.checkBoxGeometryMarginRight->setChecked(true);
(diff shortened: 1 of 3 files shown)


sunderme:
thanks

PR #3162, opened 2023-06-13
new cwls

mbertucci47:
New cwls for `javascripthttp`, `luahttp`, `quizztex`, `zref-abspos`, `zref-marks`, and `zref-runs`. Several updated cwls.


sunderme:
Thanks

Issue #3166, opened 2023-06-17
Show inline preview using original preamble only


octaeder:
### Describe the feature and the current behavior/state
Compiling the document with the complete include tree has many advantages: Layout is exactly as expected, you have full access to the contents and the index, no problems with missing references. An obvious disadvantage is that this method can be quite time consuming especially when multiple runs are necessary. Cutting down the include three may speed up things but all the advantages have gone.

But txs offers a great preview functionality. Unfortunately the result is often not similar to what you get from a full compile. This happens because the preview compile adds some extra code at the end of the original preamble (s. texstudio.cpp):
```latex
\usepackage[active,tightpage]{preview}
\usepackage{varwidth}
\AtBeginDocument{\begin{preview}\begin{varwidth}{\linewidth}}
\AtEndDocument{\end{varwidth}\end{preview}}
```
This optimizes the output, which is quite nice particularly for tooltips. But the inline preview will not be what one will get from the final compile, since \parindent is changed (anything else?). But I often need to check if the text layout is fine in each line: Spacing between words, text fits in the line so nothing is moved into the right margin, and I want to avoid line breaks within inline formulae. Fine tuning the document in this way is very time consuming, thus a preview option that gives you a preview similar to the final document would be of great help. In fact this is available when you set Preview Display Mode to Show in embedded viewer. But in this way you can't compare the new result to the old one.

### Who will benefit with this feature?
Almost anyone want's to avoid the time consuming work of adjusting text. Therefore a preview with original preamble should be of great help.

### Any Other info
Current behavior;
![grafik](https://github.com/texstudio-org/texstudio/assets/102688820/035540a9-4784-47ac-9d4e-84f8f056a637)

Expected behavior:
![grafik](https://github.com/texstudio-org/texstudio/assets/102688820/da5b9755-56de-48af-8a43-dad3c550db9f)

Derived from (build with lualatex):
```latex
\documentclass[ngerman,10pt,a5paper]{article}
\usepackage[T1]{fontenc}
\usepackage[width=11cm, height=17cm]{geometry}
\usepackage{mathtools}
\usepackage{babel}
\begin{document}
Lorem ipsumdolor sitamet, consetetur sadipscing elitrsed diamnonumy eirmod
tempor invidunt utore etadolor magna ali $\forall (a,b,c,d,e,f,g,\ldots,z)$
erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea
rebum. Stet clita kasd gubergren

[...]

PR #3175, opened 2023-06-24
show preview with document's preamble (option)

octaeder:
This PR gives the user the option to decide whether preview uses the documents preamble as is or with some additions (the current way). For details s. #3166.

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -4050,6 +4050,19 @@ them here.</string>
                    </widget>
                   </item>
                   <item row="7" column="0" colspan="2">
+                   <widget class="QCheckBox" name="checkBoxShrinkPreviewPage">
+                    <property name="text">
+                     <string>Shrink Preview Page</string>
+                    </property>
+                    <property name="toolTip">
+                     <string>With this option, the preview uses the smallest possible page to display the preview data within a minipage.</string>
+                    </property>
+                    <property name="checked">
+                     <bool>true</bool>
+                    </property>
+                   </widget>
+                  </item>
+                  <item row="8" column="0" colspan="2">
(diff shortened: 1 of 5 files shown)


sunderme:
I am not sure how ignoring all I say should be helpful here.
Especially as showing the preview in the embedded viewer basically already fulfill all your needs.

octaeder:
I appreciate your opinion very much, and I read your words carefully. But I would say the imbedded viewer fulfills basic needs. I showed in detail in #3166 what the weaknesses of the current state are and what slows down the workflow. But I find no answer to my last comment. From my point of view I'm still waiting for significant counter-arguments. With this pull request I want to emphasize how useful I think this is. Anyone now has the chance to try out the changes to get a better feel for them.

Counter-arguments I found in your comments: Preview is not _designed_ for this. What does this mean? Embedded viewer solves my problem. No, it doesn't. User may often need to change the option. Why? With defaults nothing changes, and switching by keys is very fast.

sunderme:
1) switching mode manually is simply bad/lazy gui design
2) there seems to be no willingness to reach a compromise or consensus.

So, you personally have a solution which works for you, why should anyone else compromise (and even adding an option is already compromising)

octaeder:
Well, I can remove the option. Instead I would add a new item _Page Preview_ to the menu.

sunderme:
the question remains: it is a function which no one else will use.
Why put it in the central repository ? You can work with your build just fine.

octaeder:
Since it comes to statics: Do you have any numbers of how many downloads a release has, or how many people in which countries do use txs?
Ok, but going with a proprietary build isn't something you want to make permanent.

sunderme:
txs has about 90000 downloads per month.

The alternative to proprietary build is implementing the desired functionality with macros.

PR #3178, opened 2023-06-27
setCheckedPreviewModeAction is not a slot

octaeder:
- moved to private
- some cosmetics


sunderme:
why ?

octaeder:
because it doesn't appear in a connect

PR #3186, opened 2023-06-28
make xparse.cwl and xfp.cwl stubs

mbertucci47:
The commands from the `xparse` and `xfp` packages have been fully integrated into the LaTeX kernel and are listed in `latex-document.cwl` or `latex-dev.cwl`. This PR makes `xparse.cwl` and `xfp.cwl` into stubs and removes all `include:`'s for these packages, as well as for `expl3` (this has also been integrated into the kernel). Beyond cosmetics this is helpful because multiple conflicting cwl listings can lead to unexpected highlighting/completion. A few updated cwls as well.


sunderme:
thanks

PR #3206, opened 2023-07-13
scale tooltip preview to keep the image width and height (physical length) constant

### LINKED ISSUE #3197: Scale image tooltip preview with respect to resolution, and optionally load larger image files
When Adv. Editor option _Show image tooltip on image files_ is checked then txs shows a preview of image files in a tooltip when the mouse hovers over the file name. To avoid large tooltips, the preview width is limited to 400pix by scaling down images wider than this. This process has two limitations: 

1. By default Qt doesn't load large image files (limit approx. 30MiPix) and the user can't change this value.
2. Print size of the image (say in cm) in the tooltip depends on the screen's dpi value because the image's dpi info is ignored (and the 400pix limit).

I would like to add an option so the user can change the 30MiPix limit, and would scale preview images with respect to dpi ratio. What do you think?

### Example:
A typical way to include images is the following:
```latex
\documentclass[10pt,a5paper]{article}
\usepackage{graphicx}
\begin{document}
\includegraphics{image}
\end{document}
```
Assume image is of type png, has 8'504pix width and 4'783pix height (38.79MiPix, 16:9), and has 2'400dpi (a typical value for book printing). Thus it would be printed with 9cm x 5.1cm (calculation of width: 8'504pix/2'400dpi*2.54cm/in = 9cm). Currently one would see no image:

grafik
(zoomed 2x)

The tooltip shows part of the frame that surrounds the image and a shadow for the frame, but the image is not loaded (by 1st limitation).

To circumvent 1st limitation, we reduce the number of image pixels by a factor of 1/16 to 2.54MiPix. For this scale width, height and dpi by a factor of 1/4. Hence the print size doesn't change. The image now has a width of 2'126pix, is loaded,  and is then scaled to a width of 400pix (limited tooltip size):

grafik

But a short calculation shows that the image displays larger on a screen with 94dpi than it would be printed: 400pix/(94pix/in)*2.54cm/in = 10.8cm. The correct number of pixels would be 9cm/(2.54cm/in)*94pix/in = 333pix. So the image is displayed about 20% larger. In fact it can happen that images are displayed with the same size even so they don't print in the same size, or vice versa.

Anyone who uses image preview.

### Any Other info
void QImageReader::setAllocationLimit is available since Qt 6.0.

octaeder:
This PR resolves #3197. Tooltips for images show when you hover (with the mouse pointer) over a file name or complete a file name with the completer. With this PR the image is scaled proportional to the quotient of both the pixel densities (resolutions) of the screen and the value given in the image data. By this the quotient of width in pixels divided by resolution, i.e. the physical width (and also height) keeps constant.

From version 6 on, Qt offers a [function](https://doc.qt.io/qt-6/qimagereader.html#setAllocationLimit) that can change the amount of storage a loaded image may use. Larger images will not be loaded as a precaution. This is addressed by a new option in the ini file (not in the config dialog), so the user can change the limit. The options name is `Editor\ImageToolTipLoadMaxPixels` with a default value of $30$. This means that the storage calculated for the image, with 4 bytes per pixel, is $30 \cdot 4\text{MiBytes}=30 \cdot 1024^2 \cdot 4\text{Bytes}$. It follows that the number of pixels the image can have, could be a little bit larger than $30\text{Mpixels}$ (where M means $1000^2$). A value of $0$ means no limit.

Let me note that there is another limit in txs that affects image tooltips: The width can't be larger than 400pix, otherwise the image is scaled to this limit, that is stored in option `Editor\MaxImageTooltipWidth`.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -546,6 +546,7 @@ ConfigManager::ConfigManager(QObject *parent): QObject (parent),
 	registerOption("Editor/ToolTip Help", &editorConfig->toolTipHelp, true, &pseudoDialog->checkBoxToolTipHelp2);
 	registerOption("Editor/ToolTip Preview", &editorConfig->toolTipPreview, true, &pseudoDialog->checkBoxToolTipPreview);
 	registerOption("Editor/ImageToolTip", &editorConfig->imageToolTip, true, &pseudoDialog->checkBoxImageToolTip);
+	registerOption("Editor/ImageToolTipLoadMaxPixels", &editorConfig->imageToolTipLoadMaxPixels, 30);
 	registerOption("Editor/MaxImageTooltipWidth", &editorConfig->maxImageTooltipWidth, 400);
 	registerOption("Editor/ContextMenuKeyboardModifiers", &editorConfig->contextMenuKeyboardModifiers, Qt::ShiftModifier);
 	registerOption("Editor/ContextMenuSpellcheckingEntryLocation", &editorConfig->contextMenuSpellcheckingEntryLocation, 0, &pseudoDialog->comboBoxContextMenuSpellcheckingEntryLocation);
--- a/src/latexeditorview_config.h
+++ b/src/latexeditorview_config.h
@@ -43,6 +43,7 @@ public:
 	bool toolTipPreview;
 	bool toolTipHelp;
 	bool imageToolTip;
+	int imageToolTipLoadMaxPixels;
 	int maxImageTooltipWidth;
 	bool texdocHelpInInternalViewer;
 	bool monitorFilesForExternalChanges;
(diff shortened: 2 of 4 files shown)


sunderme:
I am not convinced on this PR.
1) not sure why i would need dpi dependent width. It just shows me the image as a reminder, size is anyway determined by includegpraphics scaling.
2) if imagereader is already involved, why not use it to read scaled images, avoiding any issue with memory limitations
3) if you run into the imagereader allocation limits, your images are oversized (unless we are talking posters), but that limit can also be removed via env variables, so there is not even a need for a txs option.

sunderme:
I am not merging this PR.
1) if you run into images size limit, the images are too large. In any case, an env variable can be set to disable that limit, see [here](https://doc.qt.io/qt-6/qimagereader.html#setAllocationLimit)
2) latex usually sets the width, so preview width is of no relevance

PR #3217, opened 2023-07-21
new cwls

mbertucci47:
New cwls for `hep-feynman`, `hep-graphic`, `hep-plot`, and `rit-fonts`. Several updated cwls, including removing `#include:textcomp` from all package cwls since textcomp is now part of the format and listed in `latex-document.cwl`


sunderme:
thanks

PR #3240, opened 2023-08-02
real cancel for insertRef dialog

### LINKED ISSUE #3230: Unexpected behavior of the Cancel button of the Ref window.
### Environment
- TeXstudio: 4.6.2
- Qt: Qt 6.5.1
- OS: Windows 10
- TeX distribution: TeX Live 2023

### Actual behavior
Do the following steps:

1. `File / New`
2.  `LaTeX / Cross References / ref`
3. `Cancel`

TXS inserts the following at the cursor:
`\ref{}`

### Expected behavior
Nothing happens.

octaeder:
This PR resolves #3230. Cancel button now quits the dialog without any effect.
First entry in the list of labels is a surrogate that allows inserting `\ref{}` (empty command argument) into the document.

Note: When the dialog is presented you can immediately (at least on windows)

- scroll through the list with the up/down arrow keys
- open the list with Alt+down arrow key (image to the right)
- insert `\ref{}` by pressing Enter or clicking OK button

image

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -5739,7 +5739,7 @@ void Texstudio::insertRef(const QString &refCmd)
 		foreach (const LatexDocument *doc, docs)
 			labels << doc->labelItems();
 	} else return;
-	QString listItemUseNoLabel = "<" + tr("use no label") + ">";	// of course we can't insert in the document a ref to this label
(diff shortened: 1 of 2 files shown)


sunderme:
Could we change it to `<empty>` ?

sunderme:
Thanks

PR #3248, opened 2023-08-09
new cwls

mbertucci47:
New cwls for `addtoluatexpath`, `class-apa6`, `setspaceenhanced`, and `wordcloud`. Several updated cwls


sunderme:
thanks

Issue #2848, opened 2023-01-14
Loading Package information from CTAN


sunderme:
CTAN is a huge latex packaage database.
Apart from providing the actual packages, it also contains a caption and a short description on every package.

The idea is to download that data to provide info for all packages within txs.
### Describe the feature and the current behavior/state
Load caption/description from CTAN to provide package information for quickstart wizard, package insertion via menu or with the completer.
### Who will benefit with this feature?
Beginners who need some more introductory information on some packages.

sunderme:
CTAN offers a rest json api.
e.g.: ```https://www.ctan.org/json/2.0/pkg/mathtools```

see also https://www.ctan.org/help/json/2.0/pkg

sunderme:
they also offer a categorized overview which may be interesting for a kind of package wizard ?

muzimuzhi:
> CTAN offers a rest json api. e.g.: `https://www.ctan.org/json/2.0/pkg/mathtools`
> 
> see also https://www.ctan.org/help/json/2.0/pkg

Corner cases: The CTAN package name may not be the same as the name in `.sty` or `.cls` files.
- There're CTAN bundle packages that provide several `.sty` and/or `.cls`. For example, CTAN package `amsmath` provides at least `amsmath.sty` and `amscd.sty`, which are designed to load separately. And of cause the `koma-script`, `tikz`, and `zref` CTAN packages.
- There're CTAN packages that each of them does only provide a single `.sty` or `.cls` file, but with a file name different from the corresponding CTAN package name. I can't name one of them right now, but if my remember is alright...

To sum up, a mapping from `.sty/.cls` (base) file names to CTAN package names is needed. Also note that CTAN's current searching API doesn't support file search, see https://www.ctan.org/help/json/2.0/searching .

mbertucci47:
> There're CTAN packages that each of them does only provide a single .sty or .cls file, but with a file name different from the corresponding CTAN package name. I can't name one of them right now, but if my remember is alright...

For example [erewhon-math](https://www.ctan.org/pkg/erewhon-math) provides `fourier-otf.sty`.

octaeder:
What about the Package Help in the Help menu? A quite interesting feature.

muzimuzhi:
> > There're CTAN packages that each of them does only provide a single .sty or .cls file, but with a file name different from the corresponding CTAN package name. I can't name one of them right now, but if my remember is alright...
> 
[...]

PR #3259, opened 2023-08-16
Enhance TeXdoc dialog

octaeder:
- use package list from CTAN
- show package list as a table
- regExp search filter for package names
- add a button which opens CTAN package doc
- fix enabled open button while testing whether texdoc is available

### DIFF
--- a/src/latexrepository.cpp
+++ b/src/latexrepository.cpp
@@ -1,4 +1,5 @@
 #include "latexrepository.h"
+#include <QXmlStreamReader>
 #include <QMutex>
 
 LatexRepository *LatexRepository::m_Instance = nullptr;
@@ -6,7 +7,7 @@ LatexRepository *LatexRepository::m_Instance = nullptr;
 LatexRepository::LatexRepository() :
     QObject(nullptr), m_dataSource(None)
 {
-	loadStaticPackageList(":/utilities/packageList");
+	loadStaticPackageList("packages.xml");
 }
 
 LatexRepository *LatexRepository::instance()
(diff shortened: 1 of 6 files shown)


dbitouze:
> * regExp search filter for package names

I didn't follow the discussion and have no much time right now but not regExp search on packages descriptions as well?

Anyway, many thanks for such a great improvement!

octaeder:
@dbitouze Thanks for the compliment. Yes, currently the regExp is applied to the name column only. I already thought about searching the caption (this is the ctan notation they use for the headline, the package description is more text). But this needs some more design decisions. When you apply the expr. to both columns you may need to type more until you find your package. Adding a new search field for the caption might look a bit too much for the ui, and you may need options telling how to combine the results (and/or). Or use only one option telling which column you want to search (this could include a third case for "both columns". And this is still not all I need to know.
You may have a look at the lower end of #2848 for an image.

sunderme:
thanks

octaeder:
> > Or use only one option telling which column you want to search (this could include a third case for "both columns".
> 
> @octaeder Would be nice, IMO.

I can create a PR for this:

grafik

sunderme:
the text should search title and description,  no options

sunderme:
you can play around with the result order

sunderme:
Yes

octaeder:
I will not guess what you want

sunderme:
alright

octaeder:
so it's OR

octaeder:
You need to update the tooltip.
Since you always search the description case insensitive, it makes no sense to enter upper case letters in the search field, because the name field searched in fact is the key, which is always in lower case. So a distinction of foundLevel 2 and 3 is useless (s. https://github.com/texstudio-org/texstudio/commit/26d0ede547ff6f9094eddbd110baa6c9be6f7b81).

octaeder:
changing the checkbox doesn't give the appropriate result for the search term (inconsistent)

tmelorc:
I'm using TeXstudio 4.8.2 (git 4.7.1-420-g6872e90d1).
I'd suggest to use smaller window's height.
When I open the Packages Help on my laptop, the window is too tall so that buttons are almost out of screen.

PR #3269, opened 2023-08-27
improve TeXdoc dialog

octaeder:
This PR mainly addresses a matching issue for Captions: For example, when you enter `longtables` no match can be found in the first column (Package) but in the second one (Caption). But none of the rows shown is selected (i.e. not set as current row).

Also enabling/disabling of buttons/fields not worked correctly in all cases.


sunderme:
thanks

PR #3272, opened 2023-08-28
enhance retrieving packages and captions for tooltips

octaeder:
This PR addresses following: Editor shows tool tips on hovering the mouse over arguments of commands like \documentclass and \usepackage. Currently not all arguments are found at all or the package description can't be retrieved:

\documentclass:

grafik

The Koma class is not found. With the fix a package with the same name is found and the caption can be shown from the ctan list introduced recently:

grafik

\usepackage:

grafik

After fix package and caption are found:

grafik

This affects many packages in my test.

### DIFF
--- a/src/latexeditorview.cpp
+++ b/src/latexeditorview.cpp
@@ -2696,15 +2696,15 @@ void LatexEditorView::mouseHovered(QPoint pos)
 				type = tr("Beamer Theme");
 				type.replace(' ', "&nbsp;");
 			}
-            QString text = QString("%1:&nbsp;<b>%2</b>").arg(type,value);
-            if (latexPackageList->find(preambel + value) != latexPackageList->end()) {
+			QString text = QString("%1:&nbsp;<b>%2</b>").arg(type,value);
+			if (LatexRepository::instance()->packageExists(value)
+				|| latexPackageList->find(preambel + value) != latexPackageList->end()) {
 				QString description = LatexRepository::instance()->shortDescription(value);
 				if (!description.isEmpty()) text += "<br>" + description;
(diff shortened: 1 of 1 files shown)


sunderme:
Basically you change it, so that not-installed packages get tooltips.
In that case, the tooltip may give a false impression that the package is actually available and it contradicts the green wave below the package name. I can't say that I see that as an advantage.

sunderme:
a short test seems to show tooltips also with the old code, so not sure what the change actually is ?

octaeder:
yes, this was not intended. I've overseen that `Texstudio::texdocHelp` (line 6766ff) is the one that creates the (partial) package list. Maybe we should have a look at the code there, because what you describe isn't true in all cases:

grafik

You can see the green wave (tooltip says not found), even so the package notespages is installed. So something is wrong. Same with class article.

sunderme:
what latex distribution do you use ?
Here Miktex on windows seems to work fine, package detection wise.

octaeder:
win10 miktex

sunderme:
I have just rechecked on windows (11), package detection works with txs.
Do you see package name cache (appdata) ?
Screenshot 2023-09-01 182404

octaeder:
yes
image

sunderme:
if those files have meaningful content, you would need to run a debugger to see what is going on...

octaeder:
files look good

octaeder:
but files are not complete. I debuged LatexEditorView::updatePackageFormats and MiktexPackageScanner::run. This showed that the missing packages were not listed. After updating/repairing miktex the scanner now works ok.

PR #3281, opened 2023-09-02
new cwls

mbertucci47:
New cwls for `cahierprof`, `highlightx`, `iwonamath`, and `metsymb`. Several updated cwls


sunderme:
thanks

PR #3296, opened 2023-09-14
Invert border colour of magnified area for inverted colours

samcarter:
If one uses the inverted colour option of the pdf viewer, the black border/shadow of the magnified area are hard to see. They would be much easier to see if one would invert the colours in such a case.


sunderme:
thanks

PR #3304, opened 2023-09-20
cwl: use arg name `basename` and disable syntax check for most `basename` args

muzimuzhi:
`\glsxtrresourcefile` in `glossaries-extra` takes a basename without file extension as its mandatory argument. For example, `\glsxtrresourcefile{\jobname}` specifies `\jobname.glstex` as the resource file.

See `glossaries-extra` package manual (v1.52, 2023-06-28), sec. 11. `bib2gls`: Managing Reference Databases
- image
- image

### DIFF
--- a/completion/glossaries-extra.cwl
+++ b/completion/glossaries-extra.cwl
@@ -2556,8 +2556,8 @@ topicmcols
 \glsifcategoryattributehasitem{category}{attribute}{item}{true}{false}#*
 
 ### 11 bib2gls: Managing Reference Databases ###
-\glsxtrresourcefile{filename%file}
-\glsxtrresourcefile[options%keyvals]{filename%file}
+\glsxtrresourcefile{basename}
+\glsxtrresourcefile[options%keyvals]{basename}
 \GlsXtrLoadResources
 \GlsXtrLoadResources[options%keyvals]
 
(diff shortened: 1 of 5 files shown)


mbertucci47:
Perhaps better to have `{basename%definition}` to prevent `file_name` from being highlighted as an error?

muzimuzhi:
It seems `{basename}` is not treated as `{basename%text}`, thus using `_` in basename like `\glsxtrresourcefile{file_name}` won't trigger highlighting error.

image

completion/glossaries-extra.cwl
```diff
@@ -2556,8 +2556,8 @@ topicmcols
 \glsifcategoryattributehasitem{category}{attribute}{item}{true}{false}#*
 
 ### 11 bib2gls: Managing Reference Databases ###
-\glsxtrresourcefile{filename%file}
```
sunderme, line 2559: `%file`actually means that the txs completer tries to complete filenames on disk.

muzimuzhi: But the completer provides a list of full filenames. Neither is perfect.

Feel free to close this PR if you prefer filename completer.

sunderme:
then I would rather have a real basename completion

PR #3355, opened 2023-10-31
make `grayscale` options available for embedded viewer users

samcarter:
For users of the embedded pdf viewer, options like "Grayscale", "Scrolling follows cursor", "Cursor follows scrolling" or "Synchronize multiple views" were not easily accessible (and even harder to discover). 

If they were listed in the right-click context menu, such users could use them more easily.

(inspired by this question https://tex.stackexchange.com/q/700072/36296)

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -3382,6 +3382,7 @@ void PDFDocument::init(bool embedded)
     //connect(actionInvertColors, SIGNAL(triggered()), pdfWidget, SLOT(update()));
 	conf->registerOption("Preview/Grayscale", &globalConfig->grayscale, false);
 	conf->linkOptionToObject(&globalConfig->grayscale, actionGrayscale);
+    pdfWidget->addAction(actionGrayscale); // add grayscale to widget context menu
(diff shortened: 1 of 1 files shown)


samcarter:
The option "Ignore for syncronization" might also be nice to have for embedded viewer users, but it was not immediately obvious to me how to add it.

sunderme:
the gray scale option I can follow, but the follow cursor actions are button at the left hand side at the bottom of the embedded viewer, so no need to replicate it.

sunderme:
and I am not sure what "sync multiple views" should do in context of the embedded viewer ?

samcarter:
> the gray scale option I can follow, but the follow cursor actions are button at the left hand side at the bottom of the embedded viewer, so no need to replicate it.

Oh, indeed! I never noticed these buttons at the bottom.  Sorry for the noise!

(I'll trim the PR down to just the grayscale option)

samcarter:
Update the PR to only add the grayscale option

sunderme:
thanks

samcarter:
@sunderme Thanks for your patience with this botched PR!

sunderme:
no problem

PR #3357, opened 2023-11-03
updated cwls

mbertucci47:
New cwl for `autotype`. Several updated cwls, including for the 2023/11/01 LaTeX release


sunderme:
thanks

PR #3360, opened 2023-11-10
new cwls

mbertucci47:
New cwls for `affilauthor`, `argumentation`, `ipsum`, `pmdraw`, and `polyhedra`. Several updated cwls


sunderme:
thanks

PR #3380, opened 2023-11-20
Configdialog: Move option 'Show preview on formulas' to Adv. Editor

### LINKED ISSUE #3379: Move option "Show Preview ... on Formulas" to Adv. Editor
This option (s. image) is rather an option that configures behavior of editor than behavior of preview. So I suggest moving it to the place shown:

grafik

If it's ok, I would provide a PR.
A better text would be "Show preview on formulas", especially because the preview might appear as tooltip, in preview panel or in pdf viewer (but not inline) depending on the Preview Display Mode settings. Is this acceptable with respect to efforts for translations?

Preview then looks like this:

grafik

octaeder:
This PR resolves #3379.

### DIFF
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -2231,6 +2231,16 @@ Then you can select a new shortcut by one of the following ways:
                    </widget>
                   </item>
                   <item row="3" column="2" colspan="2">
+                   <widget class="QCheckBox" name="checkBoxToolTipPreview">
+                    <property name="text">
+                     <string>Show preview as tooltip on formulas in editor</string>
+                    </property>
+                    <property name="toolTip">
+                     <string>Note: Even if the preview display mode is set to Inline, a tooltip is used.</string>
+                    </property>
+                   </widget>
+                  </item>
+                  <item row="4" column="2" colspan="2">
                    <widget class="QCheckBox" name="checkBoxImageToolTip">
                     <property name="text">
                      <string>Show image tooltip on image files</string>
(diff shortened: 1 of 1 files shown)


sunderme:
it needs to say that it affect tooltip review on formulas. Other previews work independently ...

octaeder:
So this?
Note: Even if the preview display mode is set to Inline, a tooltip is used to preview formulas.

octaeder:
Maybe you mean the option text. In fact the option affects preview by hovering the mouse. Another preview exists for selections/parentheses. It seems a bit cumbersome to point this out in the option text.

octaeder:
What I mean is, that by hovering the mouse the preview could be in the preview pane or in the pdf viewer for ex. So there is no tooltip at all.

sunderme:
the main text `<string>Show preview on formulas</string>` was changed from `<string>Show preview as tooltip on formulas in editor</string>` and I don't see a reason why ?

octaeder:
If you set this option and set Preview Display Mode to say pdf viewer there will be no tooltip. If you unset the option there will be no tooltip and no other preview. So why should there be "tooltip"?

sunderme:
we can continue once 4.7.0 is released.

sunderme:
When 4.7.0 is released, it will be published on ppa.

sunderme:
yes, keeping the original text is the idea. The main point is that it works for hover and by default tooltips are shown.

octaeder:
Show preview as tooltip on formulas in editor: keep _in editor_? we are in _Adv. Editor_ section.

octaeder:
image

sunderme:
looks okay

sunderme:
I will pull after the release of 4.7.0 any way.

sunderme:
by the way, Note: Even if the preview display mode is set to Inline, a tooltip is used. does not make much sense in the context where the option is placed now.

octaeder:
change to?
"When the Preview Display Mode is changed, the tooltip text is displayed accordingly (but for Inline it's still a tooltip)."

sunderme:
I think there is no need to have a note here.
It is rather for where the mode is actually changed.

octaeder:
image
Like this, text ok?

sunderme:
this make sense

PR #3386, opened 2023-11-26
add menu action Idefix/Show Preview on Formulas

octaeder:
This PR enables changing the same config option from the Idefix Menu:

image

Reasons for this change are similar to those for integrating sub menu Preview Display Mode (s. image).
By this change the _set of important preview config settings_ is quickly available in the Idefix menu with just a few key strokes. Less advanced users don't have to search through dozens of options in the config dialog which also has a clear start delay. While Show Preview on Formulas is often needed, it is difficult to prevent preview creation when the mouse pointer hovers over a text cluttered with formulas and math envs. Even more when Preview Display Mode is set to Preview Panel or Pdf Viewer (with unlimited viewing time) an already existing preview can be replaced with a new one by accident. With both options, Preview Display Mode (already there) and Show Preview on Formulas, one can organize work much easier and faster. Viewing previews of formulas and tikzpictures at the same time is often useful and you don't need do read the complex latex code. For doing so you need quick access to both options mentioned.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -1090,6 +1090,9 @@ void Texstudio::setupMenus()
 	menu->addSeparator();
     newManagedAction(menu, "previewLatex", tr("Pre&view Selection/Parentheses"), SLOT(previewLatex()), Qt::ALT | Qt::Key_P);
 	newManagedAction(menu, "removePreviewLatex", tr("C&lear Inline Preview"), SLOT(clearPreview()));
+	act = newManagedAction(menu, "toggleToolTipPreview", tr("Show Preview on &Formulas"), SLOT(setToolTipPreview()));
+	act->setCheckable(true);
+	setCheckedToolTipPreviewAction();
 
 	submenu = newManagedMenu(menu, "previewMode", tr("Preview Dis&play Mode"));
 	QActionGroup *previewModeGroup = new QActionGroup(this);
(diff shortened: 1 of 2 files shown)


octaeder:
I assume that it's clear why I want to disable the option. But having no temporary preview with the ease of moving the mouse pointer is clearly a drawback. Advantages of tooltips are: There is no need to remove them, there is no change in the way the text is colored (current line color is invisible, cluttered coloring at soft line breaks), no text is moved around, no duplicate previews that sometimes appear inline for the same area and no wrong symbols in the line number panel. So for small inline formulas I prefer tooltips as preview especially when there are already previews inline and in the pdf viewer. In summary I can quickly choose the option best suited for my next steps.

sunderme:
this is one of the options for a user base of <1.
It is way too inconvenient to go to a menu and switch the operating mode.
That is why: tooltip via hover, other preview with alt-p.
I really don't see any reason to deviate from that concept.

sunderme:
by the way, `Note: Even if the preview display mode is set to Inline, a tooltip is used.` does not make much sense in the context where the option is placed now.

octaeder:
PDM option now shows _When the mode is changed, the preview on formulas is displayed accordingly (but for Inline it's still a tooltip)_ as a tooltip. So it's clear that SPF interferes with preview selection/parentheses.

I already repeated my point of view and how I work. Since you say that this option isn't necessary, I conclude that I'm doing wrong. Please let me know how to proceed so that changing this option is rarely if ever necessary.

PR #3394, opened 2023-12-02
cwl: mark \DeclareExpandableDocumentCommand as xparse/ltcmd command

muzimuzhi:
All other commands were already marked in commit 051fc9b3 (support xparse command defintion, 2022-01-22).


sunderme:
thanks

Issue #3297, opened 2023-09-15
"Arrow pointing downwards then curving rightwards" SemiBold glyph is not available


idamir:
In Windows and MacOS the Unicode character for wrapping arrow is got from special fonts (Gothic Regular, Wingdings).
Other *nix system cannot have these fonts.

### Environment
- TeXstudio: 4.5.2-4.6.3
- Qt: 6.4.2
- OS: Debian
- TeX distribution: does not matter

### Expected behavior
A normal or bold wrapping arrow is shown for the current line.

### Actual behavior
The wrapping arrow 0x2937 has no symbol for the current line because it has no Bold glyph. As a result, a box is drawn instead of the arrow, as highlighted by red in the screenshot.
![image](https://github.com/texstudio-org/texstudio/assets/4737692/89def876-b06e-4159-b0ce-ee4e90f49d04)

### How to fix
I see two possible solutions.
1. if n == cursorLine and drawText == wrappingArrow, then do not setWeight=QFont::Bold (pseudo code)
2. ship in distrib a font that include bold glyph of 0x2937 character, for example cambria-math [1,2], and use is for wrappingArrow 0x2937 as the special font, may be for all OSes.

Other fonts with required glyphs: https://wutils.com/unicode/10551/fonts-all.htm

[1] https://wutils.com/fonts/cambria-math-regular/charmap
[2] https://wutils.com/fonts/cambria-math-bold/charmap

PR #3412, opened 2023-12-08
Do not set setWeight=QFont::Bold for wrappingArrow …

idamir:
…in the line-number panel for non-win and non-mac OSes.

This is solution 1 to issue #3297


idamir:
Now the curve arrow is printed.
image

PR #3414, opened 2023-12-10
Don't include libX11 on Haiku

Begasus:


Begasus:
Came accross this when checking out latest 4.7.1 version on Haiku.

Begasus:
Thanks! 👍

Issue #3124, opened 2023-05-25
Using GNUInstallDirs


Begasus:
Using install directions by using `CMAKE_INSTALL_BINDIR` et all instead of using paths like `bin` ... `share/local` ...

### Describe the feature and the current behavior/state

On Haiku when building the source it will try to install the data into `/usr/share/local` which doesn't exist (Haiku uses `/boot/system/data` for that), so the installation will fail because of this.

### Who will benefit with this feature?

Any OS out there has support for this (can't think of any that doesn't), and can use the default one or use the strings provided by GNUInstallDirs.

### Any Other info

Just created a package for the latest version (previous versions were build with qmake) with cmake, had to patch the paths to the correct sentence for the installation part.
That work can be found at: https://github.com/Begasus/haikuports/blob/222a2acfda50fee4d568a0de36aa49e5b4453e9b/app-office/texstudio/patches/texstudio-4.5.2.patchset

Begasus:
Fixed! :)

PR #3415, opened 2023-12-10
Switch to more common installation variables

Begasus:


Begasus:
Should be checked out, follow up for: https://github.com/texstudio-org/texstudio/issues/3124

sunderme:
the recommendation is to use `include(GNUInstallDirs)` to predefine the paths.
Any reason not to use this here ?

Begasus:
> the recommendation is to use `include(GNUInstallDirs)` to predefine the paths. Any reason not to use this here ?

Not directly, just haven't used it so far in any patches, I'll take a deeper look into this. Thanks for replying.

Begasus:
Think this should do it, if not please comment :) (checked localy)

sunderme:
thanks

Begasus:
Thanks! Saves patching it for the next release :)

PR #3419, opened 2023-12-14
fix issues with windowed pdf-viewer

octaeder:
This PR solves following issues found in the UI of the internal windowed pdf-viewer:

- When running fullscreen window mode and directly switching to presentation window mode (with F5) results in restoring the settings of the fullscreen window mode on returning to normal window mode. Thus the toolbar is missing and page mode is set to fit to window (what  might not be the original value from normal mode).
- When changing toolbar visibility or continuous page mode while in fullscreen window mode then these values are used on returning to normal window mode, not the ones used previously.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -2809,6 +2809,18 @@ void PDFWidget::restoreState()
 	emit changedScaleOption(scaleOption);
 }
 
+void PDFWidget::magnifierClicked()
+{
+	setTool(kMagnifier);
+	updateCursor();
+}
+
+void PDFWidget::scrollClicked()
+{
+	setTool(kScroll);
+	updateCursor();
+}
+
 PDFScrollArea *PDFWidget::getScrollArea() const
 {
 	QWidget *parent = parentWidget();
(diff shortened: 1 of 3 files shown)


src/pdfviewer/PDFDocument.cpp
```diff
@@ -3898,6 +3913,7 @@ bool PDFDocument::closeElement()
 	else if (dwOverview && dwOverview->isVisible()) dwOverview->hide();
 	else if (configManager->useEscForClosingEmbeddedViewer && isVisible()) {
 		// Note: avoid crash on osx where esc key is passed to hidden window
+		toggleFullScreen(false);
```
sunderme, line 3916: what scenario is fixed here ?

src/pdfviewer/PDFDocument.cpp
```diff
@@ -3000,8 +3012,10 @@ void PDFDocument::setupMenus(bool embedded)
     actionGrayscale=configManager->newManagedAction(menuroot,menuEdit, "grayscale", tr("Grayscale"), pdfWidget, SLOT(update()), QList<QKeySequence>());
     actionGrayscale->setCheckable(true);
 
-    actionMagnify=configManager->newManagedAction(menuroot,menuView, "magnify", tr("&Magnify"), this, "", QList<QKeySequence>(),"magnifier-button");
-    actionScroll=configManager->newManagedAction(menuroot,menuView, "scroll", tr("&Scroll"), this, "", QList<QKeySequence>(),"hand");
+    actionMagnify=configManager->newManagedAction(menuroot,menuView, "magnify", tr("&Magnify"), pdfWidget, SLOT(magnifierClicked()), QList<QKeySequence>(),"magnifier-button");
+    actionMagnify->setCheckable(true);
```
sunderme, line 3016: now magnify can also be unchecked in the menu ...

sunderme:
I find it hard to decipher what changes are supposed to do what, so I would prefer smaller PR.

src/pdfviewer/PDFDocument.cpp
octaeder: before closing the window we need to undo the changes done for fullscreen window mode.

src/pdfviewer/PDFDocument.cpp
octaeder: ok, needs to be fixed

src/pdfviewer/PDFDocument.cpp
octaeder: see 3rd list item of PR

octaeder:
This PR is implemented via #3426, #3427, #3428,  #3429, #3430, and  #3431

PR #3426, opened 2023-12-17
pdf-viewer: make actions scroll and magnify checkable

octaeder:
Make actions for scroll and magnify tools (s. menu and toolbar icons) checkable for a consistent UI. This fixes the bug (due to a negative toolid) that returning to normal window mode from fullscreen or presentation mode resulted in a wrong arrow mouse cursor.


sunderme:
thanks

PR #3427, opened 2023-12-17
pdf-viewer: Closing fullscreen window with ESC keeps toolbar removed in case ESC doesn't end fullscreen first

octaeder:
This PR resolves following issue:
Uncheck option "ESC closes fullscreen mode" in Shortcut config.
Start windowed pdf-viewer and ensure that toolbar is visible.
Start fullscreen window mode (Ctrl+Shift+F).
Press ESC, thus the pdf-viewer window is closed immediately (but internally fullscreen mode is not ended!).
Start again windowed pdf-viewer and observe that toolbar keeps invisible as it was in fullscreen mode. But it should be visible.


sunderme:
thanks

PR #3428, opened 2023-12-17
pdf-viewer: fix wrong scaling on return from fullscreen when page mode is fit to window

octaeder:
This PR fixes following issue:
Start windowed pdf-viewer.
Set page mode to fit to window.
Switch to fullscreen window mode (Ctrl+Shift+F).
End fullscreen mode with ESC.
Now scaling is wrong because when starting fullscreen mode the settings to be restored are saved after some internal changes on the UI.


sunderme:
thanks

PR #3429, opened 2023-12-17
pdf-viewer: fix continuous mode is set even so it wasn't set before

octaeder:
This PR fixes following issue:

Start windowed pdf-viewer.
Set continuous page mode (menu) in normal window mode.
Start presentation window mode (F5), then return to normal window mode (ESC).
Deactivate continuous page mode, then start fullscreen window mode (Ctrl+Shift+F).
Return to normal window mode (ESC).
Now continuous page mode is set even so it had been unset previously.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -4455,7 +4455,7 @@ void PDFDocument::toggleFullScreen(bool fullscreen)
 	bool presentation = false;
 	if (fullscreen) {
 		// entering full-screen mode
-		wasContinuous = actionContinuous->isChecked() == true;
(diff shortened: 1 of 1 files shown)


src/pdfviewer/PDFDocument.cpp
```diff
@@ -4455,6 +4455,7 @@ void PDFDocument::toggleFullScreen(bool fullscreen)
 	bool presentation = false;
 	if (fullscreen) {
 		// entering full-screen mode
+		wasContinuous = actionContinuous->isChecked() == true;
```
sunderme, line 4458: it is enough to assign the isChecked state.
If you insist on `== true` , put it into bracket s to improve readability.

sunderme:
thanks

PR #3434, opened 2023-12-18
ci: upload to uniquely named artifacts

muzimuzhi:
`actions/download-artifact@v4` disables uploading to same named artifact multiple times.

See
- https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/
- https://github.com/actions/toolkit/tree/68f22927e727a60caff909aaaec1ab7267b39f75/packages/artifact

Unfortunately this change is only tested when a tag is pushed.

Currently `actions/download-artifact@v4` only supports [downloading all artifacts](https://github.com/actions/download-artifact/tree/v4/?tab=readme-ov-file#download-all-artifacts), but not a list of selected ones (see https://github.com/actions/download-artifact/issues/216), hence three `actions/download-artifact@v4` steps are used.

See also
- https://github.com/actions/upload-artifact/issues/472

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -80,7 +80,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: startsWith(github.ref, 'refs/tags/')
         with:
-          name: release
+          name: release-win
           path: |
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #3444, opened 2023-12-23
execute any type of macro from Macro Dialog

octaeder:
This PR enhances User Menu Dialog (Macro Dialog). Currently you can execute macros by pressing button `Run Script` only for type `Script`. For any other type (Snippet or Environment) the button isn't enabled, so the only thing you can do is editing the macro.


sunderme:
why would you want to "execute"/insert text/environments via the edit macro dialog ?

octaeder:
This may not be an everyday use case. I can think of adding several snippets for example. This would be slightly simpler than always reopening the menu. If you are not sure what you want to use, you can read the description. Or you want to make a change before inserting the snippet. In fact the main benefit is a simplification and reduction of code that offers more options for the user. Imposing restrictions that are totally unnecessary and need more lines or more complex code is not reasonable and the somewhat strange behavior irritates the user. When going through the macro tree the button may often change its appearance without any good reason. So this is clearly an improvement.

PR #3445, opened 2023-12-23
Quick Start -> Packages > Allow to select environments for algorithms

firefly-cpp:
It would also be nice to have the option to select algorithm environments that are must-use packages among computer science researchers.

### DIFF
--- a/src/quickdocumentdialog.cpp
+++ b/src/quickdocumentdialog.cpp
@@ -92,7 +92,7 @@ QuickDocumentDialog::~QuickDocumentDialog()
 
 QString QuickDocumentDialog::getNewDocumentText()
 {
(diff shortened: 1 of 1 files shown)


octaeder:
There was a long ongoing discussion which packages should be integrated and what the needs are for a typical user of the wizard. Other solutions: You can add your packages to the list on your own, or you can use templates (s. menu and manual).

sunderme:
the mentioned discussion s here: https://github.com/texstudio-org/texstudio/issues/2831

That said, the current code structure is not really suited for extension.

Finally the offered packages should give a good start for a document, not a complete list of available packages. Hence 3 packages which seem to do the same thing are not really needed here.

firefly-cpp:
Thanks for your explanation.

PR #3448, opened 2023-12-24
fix No such signal QNetworkReply::error in Macro (Repository) Browser

octaeder:
This PR fixes following in txs 4.7.2 current dev:

**Disconnect from network**
Open Macro Editor
Press `Browse` button to connect to repository. No data can be retrieved and expected message isn't shown:

grafik

Debug shows following message indicating why expected error message is missing:

```
qt.core.qobject.connect: QObject::connect: No such signal QNetworkReplyHttpImpl::error(QNetworkReply::NetworkError)
```

Looking at Qt’s online documentation for versions [5](https://doc.qt.io/qt-6/qnetworkreply.html#signals) and [6](https://doc.qt.io/qt-5/qnetworkreply.html#signals) we find the signal’s right name:

``void errorOccurred(QNetworkReply::NetworkError code)``

The connect is coded in macrobrowserui.cpp. A code search finds following in updatechecker.cpp:
```c++
#if QT_VERSION_MAJOR<6
        connect(reply, SIGNAL(error(QNetworkReply::NetworkError)), this, SLOT(onRequestError(QNetworkReply::NetworkError)));
#else
        connect(reply, &QNetworkReply::errorOccurred,this, &UpdateChecker::onRequestError);
#endif
```
From Qt’s information I would conclude that the else path would suffice, but I keep it as is, since I can’t test Qt<6. So let’s see what update checker (from Config) shows:

grafik

Start changing macrobrowserui.cpp in a similar way:
```c++
#if QT_VERSION_MAJOR<6
    connect(reply, SIGNAL(error(QNetworkReply::NetworkError)), this, SLOT(onRequestError()));
#else
    connect(reply, SIGNAL(errorOccurred(QNetworkReply::NetworkError)), SLOT(onRequestError()));
#endif
```
The slots are used without arguments because it is declared in that way in `macrobrowserui.h`. On running this fix the expected message appears:

grafik

Be aware of the missing space after the colon. After adding a line break the text looks the same as for the update checker (s. above, no change of translations):

grafik

Press `OK` to close the message, `Cancel` to close Macro Browser and again `Cancel` to close Macro Editor. Now debug shows following message:

`` QCoreApplication::postEvent: Unexpected null receiver``

This message appears only when the repository couldn’t be found. This is fixed in the destructor of the Macro Browser.

### DIFF
--- a/src/macrobrowserui.cpp
+++ b/src/macrobrowserui.cpp
@@ -47,8 +47,10 @@ MacroBrowserUI::MacroBrowserUI(QWidget *parent):QDialog (parent)
 
 MacroBrowserUI::~MacroBrowserUI()
 {
-    networkManager->deleteLater();
-    networkManager=nullptr;
+    if (networkManager) {
+        networkManager->deleteLater();
+        networkManager=nullptr;
+    }
     foreach(QList<QTableWidgetItem *>lst,itemCache){
         foreach(auto *item,lst){
             delete item;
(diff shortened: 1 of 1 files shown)


sunderme:
Thanks

PR #3449, opened 2023-12-26
check rc after loading macro from Json

octaeder:
This PR tries to prevent situations where a macro can't be imported/loaded and thus will be replaced in the macro editor by an empty macro without warning. This happens when an old macro is loaded where the JSon has no formatVersion tag. The empty macro is also listed in the tree of macros (without name, ..., so barely visible). When the macro editor or txs is closed the file will overwritten with the empty macro. This could be very harmful to the user. The change will not create an empty member but presents a warning for the user in this case. As a consequence the file will not be overwritten as described before, but deleted.

To reproduce remove the tag from the Json and import the file (manually or by starting txs).

Note:
Current code may detect that Json parsing may not work. This case should probably handled in the same way.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -1066,8 +1066,9 @@ QSettings *ConfigManager::readSettings(bool reread)
                 QString fileName=joinPath(configBaseDir,"macro",QString("Macro_%1.txsMacro").arg(i));
                 if(QFile(fileName).exists()){
                     Macro macro;
-                    macro.load(fileName);
-                    completerConfig->userMacros.append(macro);
+                    if (macro.load(fileName)) {
+                        completerConfig->userMacros.append(macro);
+                    }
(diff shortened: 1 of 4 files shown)


sunderme:
that change is from 2019. So what real life scenario is affected here ?

octaeder:
The same case appears with json parsing errors as mentioned above. When you think that there also nothing should be done, then you can close the pr. Otherwise I would change it.

sunderme:
the change in loadFromText breaks its general logic, so no pop-up there and especially no "filename" as argument.

Also pop-ups need to be used carefully. I just had a cascade of pop-up on windows when something went wrong.

So, I don't see a real need to change something here as it is a rather theoretical question.
But if, it needs to be passive/non-blocking.

octaeder:
I understand, thanks.

PR #3451, opened 2023-12-27
export (all) macro(s) by popup menu

octaeder:
This PR adds a popup menu to the Export button of the Edit Macros dialog. This allows the user to choose whether to export current or all macros.

grafik

So there is no need to know where txs stores the macros internally.

### DIFF
--- a/src/usermenudialog.cpp
+++ b/src/usermenudialog.cpp
@@ -40,26 +40,32 @@ UserMenuDialog::UserMenuDialog(QWidget *parent,  QString name, QLanguageFactory 
 	connect(ui.runScriptButton, SIGNAL(clicked()), SLOT(slotExecMacro()));
 
 	connect(ui.pushButtonAdd, SIGNAL(clicked()), SLOT(slotAdd()));
-    connect(ui.pushButtonAddFolder, SIGNAL(clicked()), SLOT(slotAddFolder()));
+	connect(ui.pushButtonAddFolder, SIGNAL(clicked()), SLOT(slotAddFolder()));
 	connect(ui.pushButtonRemove, SIGNAL(clicked()), SLOT(slotRemove()));
 	connect(ui.pushButtonUp, SIGNAL(clicked()), SLOT(slotMoveUp()));
 	connect(ui.pushButtonDown, SIGNAL(clicked()), SLOT(slotMoveDown()));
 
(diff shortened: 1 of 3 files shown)


sunderme:
the usual gui element for this kind of things is a combobox (see sections/labeles etc. in the default toolbar).
Main advantage: single click for normal export, can be changed to export all.

octaeder:
I face following issues after spending hours of effort to resolve. I also compared to `createComboToolButton` in `utilsUI.cpp`. Current state is (almost same functionality as in this PR, but worse UI):

grafik

* You can click on the button showing Export by default and the dialog asking for the target path+name opens. This is what you wanted.
* Clicking on the arrow also allows to choose Export all and the dialog asking for the target directory opens.
But when all macros have been exported the button still shows Export (for a single macro) and behaves as described above. Why? The known toolbars from txs show the last action triggered from the menu.
* How can be guaranteed that the toolbutton has always the same size as the surrounding buttons and the same frame?

From my side I would suggest merging to get new functionality. After that one could try to fix the very small issue with the UI element - if possible. Here is my code I tried with:

```html
<layout class="QHBoxLayout" name="horizontalLayout_7">
 ...
 <item>
  <widget class="QToolBar" name="toolBarExport">
  </widget>
 </item>
 ...
</layout>
```

 ```c++
QToolButton *toolButton = new QToolButton(ui.toolBarExport);
toolButton->setToolButtonStyle(Qt::ToolButtonTextOnly);
toolButton->setPopupMode(QToolButton::MenuButtonPopup);

QMenu *menuExport = new QMenu(toolButton);
menuExport->setToolTipsVisible(true);

QAction *act = menuExport->addAction(tr("Export"),this,exportMacro);
act->setToolTip("Export macro to file");
act->setData(QVariant(QString("current")));
toolButton->setDefaultAction(act);

act = menuExport->addAction(tr("Export all"),this,exportMacro);
act->setToolTip("Export all macros to files");
act->setData(QVariant(QString("all")));

toolButton->setMenu(menuExport);
ui.toolBarExport->addWidget(toolButton);
```

I also tried to surround the toolbar item with a QVBoxLayout as decribed [here](https://doc.qt.io/qt-6/qtoolbar.html#details). But this didn't change anything.

sunderme:
i have taken your code and implemented it with an apropriate toolbutton on dev

PR #3454, opened 2023-12-28
Update/improve Swedish translations

oscargus:
Seems like the editor (QT Linguist) is adding some changes as well...


sunderme:
translations need to be made via https://app.transifex.com/texstudio/texstudio

PR #3456, opened 2023-12-29
reorder buttons in Edit Macros dialog

octaeder:
As mentioned in my PR #3451 reordering the buttons from Browse, Export, Import to Export, Import, Browse gives better alignment of the buttons and brings Import and Browse together, which are basically the same function.

grafik


sunderme:
Thanks

PR #3458, opened 2023-12-29
Edit Macros dialog: Don't add the macro type to the first line of the editor field

octaeder:
This PR eliminates the confusing interaction between the radio buttons for selecting the type of the macro and the first line of the macro, at the beginning of which the type is encoded. What makes things even more difficult is that Normal type macros are coded without a type at all. When entering the first line, unexpected effects occur if you want to start with a comment. The comment character % is recognized as code for the Environment type and changes the radio buttons to Environment. It is not clear a priori that the character loses this meaning through duplication, but ultimately the type changes back to normal. If the type is changed using the radio buttons, the coding in the first line changes. Multiple changes of type may be irreversible. From the user's point of view as well as that of the technical implementation, the duplicate type in the editor is unnecessary.

Solution: Use radio buttons only to select the type. The type must be saved in the macro object and when exporting to the Json file. Old files (format version 1) will be migrated on import.

grafik <-- old vs. new --> grafik

Json Example:
```json
{
    "abbrev": "",
    "description": [
        "Transposes the two characters to the left and right of the cursor."
    ],
    "formatVersion": 2,
    "menu": "",
    "name": "Transpose Characters",
    "shortcut": "Alt+X",
    "tag": [
        "/* V1.1 2022-04-18 by octaeder */",
        "cursor.beginEditBlock();",
        "cursor.movePosition(1,cursorEnums.Left,cursorEnums.KeepAnchor);",
        "leftchar = cursor.selectedText();",
        "cursor.removeSelectedText();",
        "cursor.movePosition(1,cursorEnums.Right,cursorEnums.KeepAnchor);",
        "rightchar = cursor.selectedText();",
        "cursor.removeSelectedText();",
        "editor.write(rightchar + leftchar);",
        "cursor.movePosition(1,cursorEnums.Left);",
        "cursor.endEditBlock();"
    ],
    "trigger": "",
    "type": "Script"
}
```

### DIFF
--- a/src/usermacro.cpp
+++ b/src/usermacro.cpp
@@ -200,6 +200,21 @@ void Macro::setTypedTag(const QString &m_tag)
     tag=parseTypedTag(m_tag,type);
 }
 
+QString Macro::getTag()
+{
+    return tag;
+}
+
+void Macro::setTag(const QString &ntag)
+{
+    tag=ntag;
+}
+
+void Macro::setType(const Macro::Type &ntype)
+{
+    type=ntype;
+}
+
 QString Macro::parseTypedTag(QString typedTag, Macro::Type &retType)
 {
 	if (typedTag.startsWith("%SCRIPT\n")) {
@@ -250,9 +265,9 @@ bool Macro::save(const QString &fileName) const {
         return false;
 
     QJsonObject dd;
-    dd.insert("formatVersion",1);
+    dd.insert("formatVersion",2);
     dd.insert("name",name);
-    QString tag= typedTag();
+    dd.insert("type", ( type==Script ? "Script" : ( type==Environment ? "Environment" : "Snippet" ) ) );
     dd.insert("tag",QJsonArray::fromStringList(tag.split("\n")));
     dd.insert("description",QJsonArray::fromStringList(description.split("\n")));
     dd.insert("abbrev",abbrev);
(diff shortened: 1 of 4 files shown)


octaeder:
Note: I looked into texstudio.cpp, latexcompleter.cpp, latexdocument.cpp, and scriptengine.cpp but couldn't see dependencies to this change. I tested inserting snippets from menu/toolbar like math envs. I selected some lines in the editor and tried scripts enumerate/itemize from menu.

sunderme:
there are two independent changes here:
1) the macro file format is changed
2) the macro editor does not display the internal tags in the first line.
The first is not necessary and affects only internal cosmetics.
The second change may be helpful. Personally I find the clear distinction between the modes which is clearly visible in the editor helpful. So, maybe we need other visual clues ?

octaeder:
The first line is no longer visible anyway if you scroll down a few lines. I suggest a combobox:

macroType (3 images)

I could make a separate PR for a switch to the combobox first.

octaeder:
typed tag data (see method typedTag) is the result of merging two attributes into one. In general one shouldn't do so. In fact the result becomes ambiguous in some cases. To avoid this you need further workarounds then. This becomes obvious when you save a normal macro which begins with a % in the first line followed by letters. All this can be avoided, including special coding and exception rules for the user, by simply storing the type in the Json. Of course one would like to avoid a version change. But with the ability to migrate old macros automatically this should not be a big problem.  And you don't need to do it immediately for the Json in the macro repository. Even so it will still be possible to create and parse typed tags with the code. Do you see any specific problems with the version change?

sunderme:
> 
> I could make a separate PR for a switch to the combobox first.

That is not better than the radio boxes.

sunderme:
the visual clue what is inserted is weak.

PR #3463, opened 2023-12-31
basic update of advanced.md with respect to macros

octaeder:


utilities/manual/source/advanced.md
```diff
@@ -167,8 +167,7 @@ in the macro-menu. It is rearranged with the
 \"up\"/\"down\"/\"add\"/\"remove\" buttons or with drag and drop.
 Folders can be added to sort a larger number of macros sensibly. To move
 macros into/from folders, only drag and drop works.
-The \"run script\" button directly executes a script in the editor for
-testing.
+The \"Exec Macro\" button directly places the text, environment, or the result from executing your script in the editor.
```
sunderme, line 170: "or the result from executing your script in the editor" is factually wrong. It just executes the script.
There is no need to elaborate here. "Exec macro" does just that.

utilities/manual/source/advanced.md
```diff
@@ -225,18 +224,15 @@ inserted as:
 \end{environment}
 ```
 
-```{note}
-TeXstudio needs that the env-name starts with \"%\", though that
-character is not placed on insertion.
-```
+Don't forget to set the macro type to \"Environment\" by selecting that radio button.
```
sunderme, line 227: redundant

sunderme:
thanks

PR #3464, opened 2024-01-01
fix #3190

### LINKED ISSUE #3190: toggling between embedded and windowed pdf viewer
### Environment
- TeXstudio: 4.5.2 as well as current build
- Qt: 6.5.0/6.5.1
- OS:  Win 10
- TeX distribution: miktex

### Expected behavior
switching from embedded to windowed pdf viewer should show the window in the same state as it has been before embedding the viewer, i.e. position, size and maximize state of the window,  and position and size when changing the window back from maximized to a normal window.

### Actual behavior
1. After embedding a maximized window it will not be restored to maximized state when switching back to windowed viewer.
2. Switching from maximized to normal window should restore previous window position and size.

### How to reproduce
For simplicity remove profile information for txs. Open an already compiled tex file and start the pdf viewer, then switch to windowed viewer (s. animation below from current master). Maximize window, then embed the viewer, then switch back to windowed viewer. Look at the upper and lower right corners of the window to convince yourself that the window is not in maximized state. Maximize window, then switch to normal window. Position and size of the window are not the same as at the beginning. To repeat just resize or move the pdf viewer, then maximize, embed, ...

pdfMaxWindow-issue
(click animation to enlarge, 6 images, 4s per image)

### Note
Moving two lines (s. image) ensures that the window is restored maximized when it was maximized before. But this has some unwanted side effects. I tried to change toggleEmbedded, but I couldn't fix the issue.

octaeder:
This PR fixes #3190.  Pdf-viewer window now can be restored to maximized window from embedded state.
Main reasons for the issue are: Resize and move after setting to maximized. Bad coordinates appear because they were retrieved and stored while in maximized state. But maximizing a window is solely achieved by setting appropriate attributes not by window coordinates.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -2857,7 +2857,7 @@ PDFDocument::PDFDocument(PDFDocumentConfig *const pdfConfig, bool embedded)
         int &h = globalConfig->windowHeight;
         QRect screen = UtilsUi::getAvailableGeometryAt(QPoint(x, y));
         // add some tolerance, as fullscreen seems to have negative coordinate (KDE, Win7 ...)
-        screen.adjust(-8, -8, +8, +8);
+//        screen.adjust(-8, -8, +8, +8);
         if (!screen.contains(x, y)) {
             // top left is not on screen
             x = screen.x() + screen.width() * 2 / 3;
@@ -2865,13 +2865,13 @@ PDFDocument::PDFDocument(PDFDocumentConfig *const pdfConfig, bool embedded)
             if (x + w > screen.right()) w = screen.width() / 3 - 26;
             if (y + h > screen.height()) h = screen.height() - 100;
         }
+        resize(w, h); //important to first resize then move then maximize
+        move(x, y);
         if (globalConfig->windowMaximized)
             showMaximized();
         else
             setWindowState(Qt::WindowNoState);
 
-        resize(w, h); //important to first resize then move
-        move(x, y);
         if (!globalConfig->windowState.isEmpty()) restoreState(globalConfig->windowState);
         toolBar->setVisible(globalConfig->toolbarVisible);
         statusbar->setVisible(true);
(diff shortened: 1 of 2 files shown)


sunderme:
unfortunately I am busy and I don't have time to review this. May take a while.

src/pdfviewer/PDFDocument.cpp
```diff
@@ -4298,11 +4298,15 @@ PDFDocument *PDFDocument::findDocument(const QString &fileName)
 
 void PDFDocument::saveGeometryToConfig()
 {
-	globalConfig->windowLeft = x();
```
sunderme, line 4301: if the window position is only saved when not maximised, then the move(x,y) is possibly undefined or wrong.
This probably comes into play for multi-monitor set-ups ...

octaeder: What I've seen with debug under Win10 lead me to the conclusion that retrieving window coordinates while in maximized (and similar) modes returns invalid values. Especially the coordinate of the upper left window corner has negative values. It seems that line 2860 was introduced to fix this. So I assume that maximized/fullscreen needs to be handled by flags not by coordinates.

Be aware that the code retrieves coordinates from different things: The window corner as mentioned before and the size of the pdf widget. This can introduce problems because we and Qt have no access to the geometry of the window decoration (how thick are the border lines of the window surrounding the widget, height of the window title, and so on, which are under control of the window manager). Typically we don't need to know these details when all changes to the coordinates (and flags) are retrieved. Because with these values we should be able to restore a correct window at any time. So the move should be no problem here. Can you reproduce some case?

But this is currently not the case. After my PR I observed following: Maximize and restore pdf window. Thus window coordinates are retrieved and stored. Now move or resize the window, then maximize it, then end txs. When starting txs again the maximized window mode is restored with my PR, but when the maximized flag is cleared, then the normal window has the last saved coordinates instead of the last ones used. I tried to improve this with following code where I need help:
```
connect(window(), SIGNAL(PDFDocument::windowStateChanged(Qt::WindowState)), this, SLOT(slotWindowStateChanged(Qt::WindowState)));
```
But I couldn't figure out how to find the window as the sender of the signal. If this would work then the current coordinates of the window could be retrieved by txs on maximizing the window (and the like). This should be sufficient for single monitor setups.

In multi monitor setups this may lead to cases where the window is placed on the wrong screens. And what about the case that one drags the maximized window to another screen such that it appears maximized (or normal) there?

octaeder:
Maybe someone else will find a good solution

PR #3478, opened 2024-01-16
Package Help dialog updated

CSMathematics:
Updated CTAN package database with more information for each package.
TexDialog


sunderme:
looks good

* packageList.xml can be removed from qrc and git if we take this solution.
* where is the json database coming from ?
* the dialog is a bit small on my 4k monitor, maybe that can be optimized (especially as it does not remember resized dialogs)

src/texdocdialog.h
```diff
@@ -38,6 +39,7 @@ private slots:
 	QString lastDocRequest;
     Help *help;
     QStringList m_packages;
+    QButtonGroup ButtonGroup;
```
sunderme, line 42: to follow (implicit) naming strategy, the variable should start with a lower case letter (buttonGroup or even m_buttonGroup)

src/texdocdialog.ui
```diff
@@ -6,32 +6,72 @@
...
       </property>
-     </column>
+      <widget class="QTextBrowser" name="PackageInfoBrowser"/>
+      <widget class="QWidget" name="">
+       <layout class="QVBoxLayout" name="verticalLayout">
+        <property name="spacing">
+         <number>0</number>
+        </property>
+        <item>
+         <layout class="QHBoxLayout" name="LanguagesLayout">
```
sunderme, line 43: the ususal naming strategy is something like lyLanguages or similar.
lower case first letter helps readability.

src/texdocdialog.ui
```diff
@@ -6,32 +6,72 @@
...
+              <width>40</width>
+              <height>20</height>
+             </size>
+            </property>
+           </spacer>
+          </item>
+         </layout>
+        </item>
+        <item>
+         <widget class="QTextBrowser" name="PackageDescriptions"/>
```
sunderme, line 67: see comment above

src/texdocdialog.h
CSMathematics: Good point. I will review the code again to make these changes.

CSMathematics:
The json database is drawn from [CTAN](https://www.ctan.org/help/json/2.0/packages) but it has been edittted.
I merged 4 different json files into 1 database file by building a custom function that gets the data and creating the file. It is not included in this project. Ithink I will be able to upload it in my repository in a few days.

CSMathematics:
The only thing I couldn't solve is to find a way to open the package documentation when the link is clicked inside the QTextBrowser.

CSMathematics:
All changes are made.

sunderme:
does not compile on OSX (LLVM)

src/texdocdialog.cpp
```diff
@@ -140,7 +143,28 @@ void TexdocDialog::itemChanged(QTableWidgetItem* item)
...
+    buttonGroup.setExclusive(true);
+
+    while(buttonGroup.buttons().count()>0){
+        QAbstractButton * bt = buttonGroup.buttons().first();
+        ui->LanguagesLayout->removeWidget(bt);
+        buttonGroup.removeButton(bt);
+        delete bt;
+    }
+    for(CTANDescription description : package.descriptions){
+        QPushButton * LangButton = new QPushButton(description.language,this);
```
sunderme, line 159: *L*angbutton

src/texdocdialog.cpp
```diff
@@ -140,7 +143,28 @@ void TexdocDialog::itemChanged(QTableWidgetItem* item)
...
-	delayedCheckDocAvailable(text);
+    LatexPackageInfo package = item->data(Qt::UserRole).value<LatexPackageInfo>();
+    QString Info = LatexRepository::packageInfo(package);
+    ui->packageInfoBrowser->setMarkdown(Info);
+    delayedCheckDocAvailable(text);
+    buttonGroup.setExclusive(true);
+
+    while(buttonGroup.buttons().count()>0){
+        QAbstractButton * bt = buttonGroup.buttons().first();
+        ui->LanguagesLayout->removeWidget(bt);
```
sunderme, line 154: LanguageLayout should start with lower case

CSMathematics: Missed that. I will fix it right away

CSMathematics:
> does not compile on OSX (LLVM)

CSMathematics:
> > does not compile on OSX (LLVM)

I changed the variable names. But I can't see the reason that compilation fails on OSX. I looked at the compile log and if I'm not mistaken the errors are on other files than the ones I modified.

sunderme:
thanks

tmelorc:
There are blank spaces before the colon. 
For example, `Name : a0poster`

octaeder:
As discussed with latex help and txs manual, it is preferred to hold sources of information locally. So the question is what happens here when you have no internet connection? Can txs handle this? Is the list of packages including captions still available? Has anyone tested this case?

tmelorc:
In addition, I tried to compile it now:
```
-- The current git hash is a8c0199
-- The current git describe is 4.4.0-1053-ga8c01995f
```

But I got the error:
Seleção_101

octaeder:
I was able to build with msys2 ucrt64

octaeder:
TeXstudio 4.7.3 (git 4.7.2-38-ga8c01995f)
Using Qt Version 6.6.1, compiled with Qt 6.6.1 R

tmelorc:
> TeXstudio 4.7.3 (git 4.7.2-38-ga8c01995f) Using Qt Version 6.6.1, compiled with Qt 6.6.1 R

It seems that `setMarkdown` is available only for Qt6.6.

octaeder:
Is this the same?
https://doc.qt.io/qt-6.2/qtextdocument.html#setMarkdown
https://doc.qt.io/qt-5/qtextdocument.html#setMarkdown

tmelorc:
> Is this the same? https://doc.qt.io/qt-6.2/qtextdocument.html#setMarkdown https://doc.qt.io/qt-5/qtextdocument.html#setMarkdown

Looks like. But the doc says _since 5.14_. Let me check my Qt version. Maybe I have older one.

dbitouze:
Nice feature!

But I don't understand the point of by default disabling the search text field: as long as the “all the packages” option is disabled (the corresponding checkbox is unchecked), nothing can be done. What I would expect would be that, by default:

- the “all the packages” option is disabled
- the search text field is enabled, allowing the user to search for a particular package (without cluttering the interface with the list of all packages).

octaeder:
@dbitouze this is impossible for you?

image

and I would say that the option is always disabled when the dialog is started

tmelorc:
I'm using the version below and the text field is locked, as screenshot below.
```
TeXstudio 4.7.3 (git 4.7.1-81-gf0f4089a9)
Utilizando Qt Versão 5.15.3, compilado com Qt 5.15.3 R

```
Documentação dos Pacotes (Texdoc)_002

octaeder:
Including 5 extra developments and missing the latest 2 (reagarding cwls) I use this (msys2 build) on Win10:

TeXstudio 4.7.3 (git 4.7.2-51-gbd8a0bd91)
Using Qt Version 6.6.1, compiled with Qt 6.6.1 R

I wonder why you see no packages at all. In any case the packages list should not be empty at start. How do you get the packages into place?

tmelorc:
> I wonder why you see no packages at all. In any case the packages list should not be empty at start. How do you get the packages into place?

Sorry, I didn't follow the question.

tmelorc:
The packages list only appears after the check button is on. Then the text field is unlocked.

Documentação dos Pacotes (Texdoc)_003

octaeder:
there's something wrong.

tmelorc:
> there's something wrong.

Maybe this line: `ui->lineEditSearch->setEnabled(n>0);`

octaeder:
since search field is not enabled I would conclude that n equals 0. But then m_packages is empty & state<=0. Since I can't reproduce, can can you debug the the values of m_packages.count() and state immediately after line 48 (int n=0;)?

octaeder:
btw: do you use packages in your latex document?

tmelorc:
> btw: do you use packages in your latex document?

You got it! I was opening the texdoc window with no opened document.
Now it works.

Anyway, it would be good to have the feature available also for no documents.

octaeder:
in this case the option should be checked

PR #3483, opened 2024-01-18
Package help dialog- Update #2

CSMathematics:
1.  CTAN links are now clickable and open external link

2. Flag icon is added when the package has multiple descriptions in different languages.


octaeder:
Info for those who want to see the flags. The cases I know of with two languages are:
biblatex-dw
catalanbib
dehyph-exptl
dhua
facture
grafcet
GuIT-corso
latex-pro-pragmatiky
systeme
tui

octaeder:
Maybe the flags could have the language code as a tooltip (for color blinds, automatic text readers...)
The flags are so nice that I would wish a single flag would also be visible

CSMathematics:
> Maybe the flags could have the language code as a tooltip (for color blinds, automatic text readers...)
> The flags are so nice that I would wish a single flag would also be visible

That's a good idea. I will also add the tooltip.

sunderme:
the flags are overkill for such a limited functionality
language code as is, is already sufficient


did I not already activate the links ?

CSMathematics:
> the flags are overkill for such a limited functionality language code as is, is already sufficient
> 
> did I not already activate the links ?

The links where active and clickable inside the QTextBrowser but the `packageInfoBrowser` had `openExternalLinks` set to false. To be honest, I didn't check if the links opened after the 1st pull request.
I will replace the flag with the language code, remove the tooltip and show only the langButton if the package has 2 descriptions, as it was before.

CSMathematics:
I removed all the recent changes and I will check only if the links open externaly.

PR #3485, opened 2024-01-19
Improve gitignore rule ignoring `build` directories

muzimuzhi:
Old rule `build*` also ignores files like `./src/buildmanager.cpp`, which will influence other tools respecting gitignore, such as ripgrep.


sunderme:
i do have more than one build folder,e.g. build-qt5 ...

sunderme:
thanks

PR #3487, opened 2024-01-21
Messages Pane: Add action to menu to copy any data of all issues to clipboard

octaeder:
This PR adds new action `Copy All Rows And Columns` retrieving all data needed for multi document projects.
Name could be changed to `Copy Table` or action could replace action `Copy All`, s. discussion below.

Discuss with me before merging (nothing of this is currently done):

- Mnemonics are all the same (&Copy) and are not visible. This could be improved.
- `Copy` and `Copy All` actions: Why should one simply copy one or all messages without further data? Since all columns of a row are selected (model behavior) the names suggest that
    - `Copy` copies complete row -> change code to do so
    - `Copy All` copies all rows (i.e. all table data) ->  remove or change to behavior of new action (to keep action name)
- `Copy All With Line Numbers` -> unchanged

### DIFF
--- a/src/latexlogwidget.cpp
+++ b/src/latexlogwidget.cpp
@@ -55,13 +55,13 @@ LatexLogWidget::LatexLogWidget(QWidget *parent) :
 	errorTable->setModel(proxyModel);
 
 	QAction *act = new QAction(tr("&Copy"), errorTable);
-	connect(act, SIGNAL(triggered()), SLOT(copyMessage()));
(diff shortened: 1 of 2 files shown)


sunderme:
I don't have any preferences here.

octaeder:
Then I'll change `Copy` and `Copy all` such that they deliver all columns.

octaeder:
finished work

PR #3488, opened 2024-01-25
cwl: add starred variants of `\VerbatimInput` and friends

muzimuzhi:
These starred variants have been supported for a long time, but not explicitly documented until `fancyvrb` v4.5c (2024-01-20), see changes of `fancyvrb` package manual.


sunderme:
Thanks

PR #3494, opened 2024-01-28
cwl: \standardstatesymbol should be \standardstate

muzimuzhi:
See source code in `chemmacros` package
https://github.com/cgnieder/chemmacros/blob/05e767ac4a9d50a134a228f4aabdd952f6296765/code/chemmacros.symbols.code.tex#L36-L40


sunderme:
Thanks

PR #3497, opened 2024-01-30
check checkbox Show all packages when package list would be empty on startup

octaeder:
When the package list of Package Help dialog would be empty then start the dialog with checkbox _Show all packages_ checked.
This resolves https://github.com/texstudio-org/texstudio/pull/3478#issuecomment-1915874683

### DIFF
--- a/src/texdocdialog.cpp
+++ b/src/texdocdialog.cpp
@@ -182,7 +182,10 @@ void TexdocDialog::itemChanged(QTableWidgetItem* item)
 void TexdocDialog::setPackageNames(const QStringList &packages)
 {
     m_packages=packages;
-    regenerateTable();
+    bool checked = false;
+    if (m_packages.count()==0) checked = true;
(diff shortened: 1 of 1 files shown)


src/texdocdialog.cpp
```diff
@@ -39,20 +39,20 @@ TexdocDialog::~TexdocDialog()
  * \brief generate all rows for table
  * \param all filter to used packages or all ctan packages
  */
-void TexdocDialog::regenerateTable(int state)
+void TexdocDialog::regenerateTable(bool cbAllPackagesChecked)
```
sunderme, line 42: this kind of works because c++ tends to interpret integer>0 as bool, but the connected signal uses integer 0,1 or 2. (pratically 0 and 2)

So, I would suggest to limit the changes to what is actually reported to be changed which is lines 185..188

octaeder: done

PR #3498, opened 2024-01-30
new cwls

mbertucci47:
New cwls for `beamerthememoloch`, `class-amnestyreport`, `genealogy-profiles`, and `odesandpdes`. Several updated cwls.


sunderme:
Thanks

PR #3507, opened 2024-02-08
insertGraphicsDlg: avoid qt.svg: Cannot open file ':/images/up-arrow-circle-silver.svg', because: No such file or directory

octaeder:
When debugging, above message shows twice when starting InsertGraphics dialog.
The image available in the given path has file extension `.png`.


sunderme:
thanks

PR #3509, opened 2024-02-10
insertGraphicsDlg: prevent elements from moving

octaeder:
Current behavior may prevent that you can click push button to expand and collapse position/placement details:

grafik
1: started dialog 2: expand 3: collapse, after this repeat 2 and 3 (click to enlarge)

Fixed behavior:

grafik
1: started dialog 2: expand, after this repeat 1 and 2 (click to enlarge)


sunderme:
thanks

PR #3510, opened 2024-02-10
packageHelp: improved printing of package.version

octaeder:
small visual improvement, because often not both of version.number and version.date are available:

current:

grafik

grafik

fixed:

grafik

grafik

grafik


sunderme:
Thanks

PR #3512, opened 2024-02-11
mathassistant: No such signal QProcess::error(QProcess::ProcessError)

octaeder:
`qt.core.qobject.connect: QObject::connect: No such signal QProcess::error(QProcess::ProcessError)`

correct name is `errorOccurred`, s. [Qt6 online](https://doc.qt.io/qt-6/qprocess.html#signals), same for Qt5

To reproduce: Start txs with debug. Select `Wizards/Math Assistant...` when it is not installed.

Note: buildmanager l. 2339 uses correct name `errorOccurred`


sunderme:
thanks

PR #3514, opened 2024-02-11
Template Selector: Add a description for top item Online Repository

octaeder:
similar to
grafik
grafik
Since Online Repo doesn't set the data you will see old data, which is misleading.
we now have
grafik


sunderme:
thanks

PR #3517, opened 2024-02-12
Template selector: Don't show hidden directories

Alexander-Wilms:


Alexander-Wilms:
Maybe the testing dir should be hidden as well. Is it used for unit testing?

sunderme:
thanks

sunderme:
we can remove that dir. It was intended for initial testing.

PR #3519, opened 2024-02-13
TemplateSelector: Handle host not available

octaeder:
- show message when host is not available
- when network connection is available reopening a folder will load files

### DIFF
--- a/src/templateselector.cpp
+++ b/src/templateselector.cpp
@@ -104,6 +104,10 @@ TemplateSelector::TemplateSelector(QString name, QWidget *parent)
 
 TemplateSelector::~TemplateSelector()
 {
+    if (networkManager) {
+        networkManager->deleteLater();
+        networkManager=nullptr;
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/templateselector.cpp
```diff
@@ -614,3 +625,15 @@ void TemplateSelector::openTemplateLocation()
 	}
 }
 
+void TemplateSelector::onRequestError()
+{
+    QNetworkReply *reply = qobject_cast<QNetworkReply *>(sender());
+    if (!reply) return;
+
+    QMessageBox::warning(this, tr("Browse template repository"),
```
sunderme, line 633: blocking message boxes again
I would rather prefer to have a message in the tree view.

octaeder: Thanks for pointing out this (again). I resolve this, see below

octaeder:
I follow already available solution (item <loading...>. For example expand item Online Repository while network is not available. The you get:

image

When connection is available collapse and expand item Online Repository and you get:

image

sunderme:
thanks

PR #3520, opened 2024-02-13
TemplateSelector: clear preview when folder Online Repository is selected

octaeder:
same behavior for all top level folders


sunderme:
thanks

PR #3521, opened 2024-02-13
add some icons to template selection dialog

octaeder:
Add icons for builtin and internet categories.
I find it quite hard to follow the change of radio buttons at the lower end of the window while selecting templates. I created an icon that can tell that this is a multifile template. While those automatically choose the directory option, now for non multifile templates it is the editor option (I find this quite natural).

image

Guess where the multifiles are.


src/templateselector.cpp
```diff
@@ -498,6 +502,7 @@ void TemplateSelector::showInfo(QTreeWidgetItem *currentItem, QTreeWidgetItem *p
 			ui.rbCreateInFolder->setChecked(true);
 			ui.rbCreateInEditor->setEnabled(false);
 		} else {
+			ui.rbCreateInEditor->setChecked(true);
```
sunderme, line 505: This is probably not a good choice, as the user may want to set it in a folder. It should not be changed if the user switched to folder.

Alexander-Wilms:
Maybe Colibre already contains the required icons?

src/templateselector.cpp
octaeder: ok, with the multifile icon it should be clear that a switch happened.

octaeder:
> Maybe [Colibre](https://github.com/rizmut/libreoffice-style-colibre) already contains the required icons?

I think you use icon set Colibre, but I use Classic. So I should set icon names depending on the selected icon set, right?

sunderme:
> > Maybe [Colibre](https://github.com/rizmut/libreoffice-style-colibre) already contains the required icons?
> 
> I think you use icon set Colibre, but I use Classic. So I should set icon names depending on the selected icon set, right?

no, you should use a general name which is used in the different icon sets (on linux from the system theme).
getIcon searches for that icon and falls back to built-in if not present

sunderme:
folder-cloud or folder-html  may be your "internet"

Alexander-Wilms:
I've created a PR against your fork: https://github.com/octaeder/texstudio/pull/1

Colibre:

image

Oxygen Modern:

image

octaeder:
thanks @Alexander-Wilms 

Now we have:
image

sunderme:
thanks

PR #3534, opened 2024-02-24
fix open online template location

octaeder:
before change: Action gives an error:

grafik

grafik

after change: Following link is opened in the browser:
[https://github.com/texstudio-org/texstudio-template/tree/main/Letters](https://github.com/texstudio-org/texstudio-template/tree/main/Letters)


sunderme:
thanks

PR #3536, opened 2024-02-25
fix del closing brace

### LINKED ISSUE #3535: alt+del removes command and the opening brace but keeps the closing brace
### Environment
- TeXstudio:  4.7.3 beta3 
- Qt: 
- OS: 
- TeX distribution: 

### How to reproduce
For 4.7.3 beta3, Alt+Del removes opening brace but keeps right brace in case of \cmd{words}. 

The difference from [#3503](https://github.com/texstudio-org/texstudio/issues/3503) is that the argument does not contains linebreak.

For example:

before (uses | for current text cursor, so command \textbf looks like \tex|tbf)
`\tex|tbf{abc}`

after
`|abc}`

octaeder:
This fixes #3535.
orig and to are to large by 1 when deleting opening brace after determining orig and to when in same line. thus delete from right to left.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -4014,10 +4014,10 @@ void Texstudio::editEraseWordCmdEnv()
                 if (cursor.nextChar() == QChar('{')) {
                     QDocumentCursor orig, to;
                     cursor.getMatchingPair(orig, to, false);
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #3537, opened 2024-02-26
macro browser dialog with treeWidget

octaeder:
This PR replaces the table with a tree in the "Browse macros from repository" dialog (similar to Template selector). Also changed:
- same bottom line on left and right side
- handle network error without message box
- after reestablished network connection collapse and expand folder to load items
- use QMap instead of QHash for deterministic order of selected macros

#### Start of dialog
grafik
Note better alignment
#### Select/Expand first directory
grafik
With the tree you always see where you are
#### Check items in last two directories
grafik
Quite easy to see which items you have selected
#### Network not available
grafik
#### After establishing network connection again
grafik
#### Add selected macros to macro list
grafik
Random order to the left (old) vs. alphabetic order of file path (new)


sunderme:
After 4.7.3 release

muzimuzhi:
You can force push to update the head branch (`octaeder:macroTree`).

PR #3539, opened 2024-02-28
macro browser dialog with treeWidget

octaeder:
This PR (a copy of PR #3537) replaces the table with a tree in the "Browse macros from repository" dialog (similar to Template selector). Also changed:
- same bottom line on left and right side
- handle network error without message box
- after reestablished network connection collapse and expand folder to load items
- use QMap instead of QHash for deterministic order of selected macros

#### Start of dialog
grafik
Note better alignment
#### Select/Expand first directory
grafik
With the tree you always see where you are
#### Check items in last two directories
grafik
Quite easy to see which items you have selected
#### Network not available
grafik
#### After establishing network connection again
grafik
#### Add selected macros to macro list
grafik
Random order to the left (old) vs. alphabetic order of file path (new)

### DIFF
--- a/src/macrobrowserui.cpp
+++ b/src/macrobrowserui.cpp
@@ -4,14 +4,16 @@
 #include <QJsonDocument>
 #include <QJsonArray>
 
+const int MacroBrowserUI::FileRole = Qt::UserRole + 0;
+const int MacroBrowserUI::UrlRole = Qt::UserRole + 1;
+const int MacroBrowserUI::PathRole = Qt::UserRole + 2;
+const int MacroBrowserUI::PopulatedRole = Qt::UserRole + 3;
+Q_DECLARE_METATYPE(QTreeWidgetItem *)
 
 MacroBrowserUI::MacroBrowserUI(QWidget *parent):QDialog (parent)
 {
-    tableWidget=new QTableWidget(4,1);
-    tableWidget->setHorizontalHeaderLabels(QStringList()<<tr("Macro name"));
-    tableWidget->setEditTriggers(QAbstractItemView::NoEditTriggers);
-    tableWidget->horizontalHeader()->setStretchLastSection(true);
-    connect(tableWidget,SIGNAL(itemClicked(QTableWidgetItem*)),SLOT(itemClicked(QTableWidgetItem*)));
+    treeWidget=new QTreeWidget(parent);
+    treeWidget->setHeaderHidden(true);
     auto *lblName=new QLabel(tr("Name"));
     lblName->setAlignment(Qt::AlignRight);
     auto *lblDescription=new QLabel(tr("Description"));
@@ -24,16 +26,14 @@ MacroBrowserUI::MacroBrowserUI(QWidget *parent):QDialog (parent)
     gridLay->setColumnStretch(0,1);
     gridLay->setColumnStretch(1,0);
     gridLay->setColumnStretch(2,1);
-    gridLay->addWidget(tableWidget,1,0,5,1);
+    gridLay->addWidget(treeWidget,1,0,5,1);
     gridLay->addWidget(lblName,1,1);
     gridLay->addWidget(lblDescription,2,1);
     gridLay->addWidget(leName,1,2);
-    gridLay->addWidget(teDescription,2,2);
+    gridLay->addWidget(teDescription,2,2,4,1);
 
     buttonBox=new QDialogButtonBox();
     buttonBox->setStandardButtons(QDialogButtonBox::Ok|QDialogButtonBox::Cancel);
-    connect(buttonBox, &QDialogButtonBox::accepted, this, &QDialog::accept);
-    connect(buttonBox, &QDialogButtonBox::rejected, this, &QDialog::reject);
     auto *layout=new QVBoxLayout();
     layout->addLayout(gridLay);
     layout->addWidget(buttonBox);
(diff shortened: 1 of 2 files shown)


src/macrobrowserui.cpp
```diff
@@ -51,20 +57,16 @@ MacroBrowserUI::~MacroBrowserUI()
         networkManager->deleteLater();
         networkManager=nullptr;
     }
-    foreach(QList<QTableWidgetItem *>lst,itemCache){
-        foreach(auto *item,lst){
-            delete item;
-        }
-    }
+    treeWidget->~QTreeWidget();
```
sunderme, line 60: this looks odd. Why call the destructor directly ?

src/macrobrowserui.cpp
```diff
@@ -4,14 +4,16 @@
 #include <QJsonDocument>
 #include <QJsonArray>
 
+const int MacroBrowserUI::FileRole = Qt::UserRole + 0;
+const int MacroBrowserUI::UrlRole = Qt::UserRole + 1;
+const int MacroBrowserUI::PathRole = Qt::UserRole + 2;
+const int MacroBrowserUI::PopulatedRole = Qt::UserRole + 3;
+Q_DECLARE_METATYPE(QTreeWidgetItem *)
```
sunderme, line 11: why is the Declare metatype needed ?

src/macrobrowserui.cpp
octaeder: Thanks for reviewing. [From manual: Destroys the tree widget and all its items.](https://doc.qt.io/qt-6/qtreewidget.html#dtor.QTreeWidget)
So I assume, that this way the loop is not needed. When I assume that closing the dialog destroys all elements, we should not even need deleting the treeWidget, right? Then why has there been the loop? What should I do?

src/macrobrowserui.cpp
octaeder: First I didn't use it, but the the linux build failed after posing the PR. Message said that the metatype macro should be used. I tried some workarounds but without success (sorry, that I already deleted all the CI/CD runs). Then I searched the manual and found an [enumeration](https://doc.qt.io/qt-6/qmetatype.html#Type-enum). The QtreeWidgetItem isn't in the list. So I used the macro.

Today I found similar code in bookmark.cpp (l. 115, 131) that might help:
```
QDocumentLineHandle *dlh = qvariant_cast<QDocumentLineHandle *>(item->data(DocLineHandle));
```
I just tried the following:
```
QTreeWidgetItem *currentItem =
             qvariant_cast<QTreeWidgetItem *>(reply->request().attribute(mbAttributeItem));
```
Same [error](https://github.com/texstudio-org/texstudio/actions/runs/8088743327/job/22103390150/#step:6:1):
```
In file included from /home/runner/work/texstudio/texstudio/src/modifiedQObject.h:5,
                 from /home/runner/work/texstudio/texstudio/src/mostQtHeaders.h:12,
                 from /home/runner/work/texstudio/texstudio/CMakeFiles/texstudio.dir/cmake_pch.hxx:5,
                 from <command-line>:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qmetatype.h: In instantiation of ‘constexpr int qMetaTypeId() [with T = QTreeWidgetItem*]’:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qvariant.h:371:37:   required from ‘static QVariant QVariant::fromValue(const T&) [with T = QTreeWidgetItem*]’
/home/runner/work/texstudio/texstudio/src/macrobrowserui.cpp:109:61:   required from here
/usr/include/x86_64-linux-gnu/qt5/QtCore/qmetatype.h:1916:5: error: static assertion failed: Type is not registered, please use the Q_DECLARE_METATYPE macro to make it known to Qt's meta-object system
 1916 |     Q_STATIC_ASSERT_X(QMetaTypeId2<T>::Defined, "Type is not registered, please use the Q_DECLARE_METATYPE macro to make it known to Qt's meta-object system");
      |     ^~~~~~~~~~~~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qmetatype.h:1916:5: note: ‘false’ evaluates to false
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qobject.h:54,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QObject:1,
                 from /home/runner/work/texstudio/texstudio/src/modifiedQObject.h:35,
                 from /home/runner/work/texstudio/texstudio/src/mostQtHeaders.h:12,
                 from /home/runner/work/texstudio/texstudio/CMakeFiles/texstudio.dir/cmake_pch.hxx:5,
                 from <command-line>:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qmetatype.h: In instantiation of ‘static constexpr int QMetaTypeId2<T>::qt_metatype_id() [with T = QTreeWidgetItem*]’:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qmetatype.h:1917:43:   required from ‘constexpr int qMetaTypeId() [with T = QTreeWidgetItem*]’
/usr/include/x86_64-linux-gnu/qt5/QtCore/qvariant.h:371:37:   required from ‘static QVariant QVariant::fromValue(const T&) [with T = QTreeWidgetItem*]’
/home/runner/work/texstudio/texstudio/src/macrobrowserui.cpp:109:61:   required from here
/usr/include/x86_64-linux-gnu/qt5/QtCore/qmetatype.h:1776:96: error: ‘qt_metatype_id’ is not a member of ‘QMetaTypeId<QTreeWidgetItem*>’
 1776 |     static inline Q_DECL_CONSTEXPR int qt_metatype_id() { return QMetaTypeId<T>::qt_metatype_id(); }
      |                                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
make[2]: *** [CMakeFiles/texstudio.dir/build.make:9572: CMakeFiles/texstudio.dir/src/macrobrowserui.cpp.o] Error 1
```
So I would return to my solution when you have no other suggestion.

muzimuzhi:
Thank you! I like this improvement.

src/macrobrowserui.cpp
sunderme: the loop is not needed as it was only necessary for the cache.
to destruct an object, call "delete treewidget"

src/macrobrowserui.h
```diff
@@ -23,17 +23,24 @@ class MacroBrowserUI : public QDialog
...
 protected:
-    QTableWidget *tableWidget;
+    QTreeWidget *treeWidget;
     QDialogButtonBox *buttonBox;
     QLineEdit *leName;
     QPlainTextEdit *teDescription;
-    QString currentPath;
     QHash<QString,QString> cache;
-    QHash<QString,QList<QTableWidgetItem *> > itemCache;
+    QMap<QString,QList<QTreeWidgetItem *> > itemCache;
```
sunderme, line 43: the itemcache is not necessary because you place the item in the tree widget where they stay until destroyed.
For the table, a cache was necessary as items become invisible (and deleted from the table) when leaving a folder.

src/macrobrowserui.cpp
```diff
@@ -198,53 +170,88 @@ void MacroBrowserUI::onRequestCompleted()
...
-        if(!currentPath.isEmpty()){
-            auto *item=new QTableWidgetItem(QIcon::fromTheme("file"),"..");
-            tableWidget->setRowCount(i+1);
-            tableWidget->setItem(i++,0,item);
+        if (currentItem) {     // not needed when inserting top level items, s. (1)
+            currentItem->takeChildren();
+            currentItem->setData(0,PopulatedRole,true);
         }
-        QList<QTableWidgetItem*> listOfItems;
+        QList<QTreeWidgetItem*> listOfItems;
```
sunderme, line 182: unused var (only items added,never read)

src/macrobrowserui.h
sunderme: still in header

sunderme:
thanks

PR #3554, opened 2024-03-07
Add grid actions in case of embedded PDF-Viewer

### LINKED ISSUE #222: Enhancement: Access to grid view, while viewer is embedded.
The builtin pdf viewer has rather powerful viewing features, such as a grid view, that isn't even provided to such a degree by fullfledged standalone viewers. This is particularly useful for Beamer, where having an overview over tens of slides becomes cumbersome in other viewers, that at best allow a two-paged view.

Sadly, this functionality is bound to using the "windowed" mode of the internal viewer, which *also* changes the interaction between viewer and editor significantly.

  - In windowed mode, focus switches from editor to viewer, once "build&view" is finished.
  - Fullscreen mode doesn't work with multiple windows.

### Desired behaviour

Access to grid view in the embedded viewer.

### Environment
- TeXstudio: 2.12.8
- Qt: Using Qt Version 5.8.0, compiled with Qt 5.8.0 R
- OS: Windows 10 (Microsoft Windows [Version 10.0.16299.492])
- TeX distribution: MikTeX

octaeder:
This PR resolves #222. For this the context menu of the embedded (not the windowed) PDF-viewer integrates some of the actions known from the View menu of the windowed viewer.

#### 2x1 grid, non continuous
image

#### context menu of embedded viewer
image

This looks similar to the View menu:

image

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -3346,27 +3346,27 @@ void PDFDocument::init(bool embedded)
 		pdfWidget->setGridSize(globalConfig->gridx, globalConfig->gridy, true);
 		conf->registerOption("Preview/PageOffset", &globalConfig->pageOffset, 0);
 		pdfWidget->setPageOffset(globalConfig->pageOffset, true);
-        // set grid menu entry checked
-        QString gs=QString("%1x%2").arg(globalConfig->gridx).arg(globalConfig->gridy);
-        bool found=false;
-        for(QAction *a:actionGroupGrid->actions()){
-            if(a->property("grid").toString()==gs){
-                a->setChecked(true);
-                found=true;
-                break;
-            }
-        }
-        if(!found){
-            // if no other grid action fits, use custom
-            actionCustom->setChecked(true);
-        }
+		// set grid menu entry checked
+		QString gs=QString("%1x%2").arg(globalConfig->gridx).arg(globalConfig->gridy);
+		bool found=false;
+		for(QAction *a:actionGroupGrid->actions()){
+			if(a->property("grid").toString()==gs){
+				a->setChecked(true);
+				found=true;
+				break;
+			}
+		}
+		if(!found){
+			// if no other grid action fits, use custom
+			actionCustom->setChecked(true);
+		}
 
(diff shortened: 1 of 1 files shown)


sunderme:
I would prefer to keep at least the config for the two viewers separate.
You can well use 2x1 continous for the windowed viewer , but 1x1 for the embedded ...

octaeder:
This sound a little tricky. I started with
```
QString previewOptionBase = QString("Preview");
if(embedded) previewOptionBase = "PreviewEmbedded";
conf->registerOption(previewOptionBase+"/GridX", &globalConfig->gridx, 1);
```
and so on. So variable gridx holds a value that is one time for the embedded and then for the windowed case. So `toggleEmdedded()` needs to store gridx in the appropriate option. But how can this be done? I think the value is written when we close txs but there we can't distinguish cases, just the last value is stored.

Currently I see options of both cases in the ini file but having the same value respectively:

```
Preview\GridX=1
PreviewEmbedded\GridX=1
```
But the first value has been 2 until I switched to the embedded viewer (getting defaults) and then closed txt.

sunderme:
alternatively start always with 1x1 in embedded mode ?

sunderme:
are you still working on this ?

octaeder:
I tried to split the affected options (continuous,...) into two options (say continuous1 and continuous2) and copy the one actually needed to the target variable (continuous) and back as necessary. But I gave up on that try.
From my point of view using same options for both cases makes it easy switching between embedded and windowed viewer. The relevance for using different setups may even decrease because the embedded viewer works almost the same as the windowed one. So one may prefer that there is no pending window.

sunderme:
> From my point of view using same options for both cases makes it easy ...

No.

The side effect is that if someone wants to have an overview on the windowed viewer, the same view is shown then on the embedded one. This is clearly not desirable and is a functional break with current implementation/expectation.

octaeder:
How could this be resolved? I would not expect that setting a 1x1 grid on start of embedded viewer will solve the problem without splitting options.
What I couldn't handle are some cases where those options are used directly:
`conf->linkOptionToObject(&globalConfig->continuous, actionContinuous, LO_NONE);`
`QString gs=QString("%1x%2").arg(globalConfig->gridx).arg(globalConfig->gridy);`

sunderme:
are you working on this ?

octaeder:
I won't work on this for a long time. Feel free to close the PR.

PR #3557, opened 2024-03-09
Cwls

mbertucci47:
New cwls for `chemformula-ru`, `didec`, `fontscale`, `lastbib`, and `useclass`. Several updated cwls


sunderme:
thanks

PR #3566, opened 2024-03-15
Replace user tags in side panel

### LINKED ISSUE #3561: Add custom beamer commands to side panel (beamer section)?
How can I add items to the "beamer commands" section in the side panel? I know how to create macros ([example](https://tex.stackexchange.com/a/393725/77895)), but would like to add them to the beamer section.

octaeder:
The tag lists in the side panel for brackets, pstricks, metapost, tikz, asymptote, beamer, and xymatrix are fixed so the user can't change or add things needed.
This PR resolves #3561 by enhancing txs such that each of those lists can be replaced by new tag lists provided by the user in the user profile (say %appdata% for win).

Note: Description changed because adding new groups has been changed to replacing tag list of existing groups like beamer (s. comment below).

#### Example
The author of #3561 describes that some items should be added to the Beamer Commands. To do so, the user may [download here](https://github.com/texstudio-org/texstudio/tree/master/tags) the xml and change it. In this case four more tags are added:

grafik

#### Description
To the left of the image above you can see the standard version of the Beamer Commands tags. To the right there is an extended version shown that txs loaded from user profile. For this you have to add a folder named `tags` to the user profile. There you have to place a file with the xml containing the description of the tags. For the example given I downloaded and changed the file `beamer_tags.xml`. The name must not be changed.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -589,12 +589,18 @@ void Texstudio::addTagList(const QString &id, const QString &iconName, const QSt
 {
 	XmlTagsListWidget *list = qobject_cast<XmlTagsListWidget *>(leftPanel->widget(id));
 	if (!list) {
-		list = new XmlTagsListWidget(this, ":/tags/" + tagFile);
+		// check for user tags
+		QString configBaseDir = configManager.configBaseDir;
+		QString pathPrefix=joinPath(configBaseDir,"tags/");
+		QFileInfo userTagFile(pathPrefix+tagFile);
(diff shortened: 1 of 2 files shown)


sunderme:
I would refer like to see this resolved like it is done with completion files.
Look if there is a user tags file of the same name, take that if yes, otherwise use txs default.
There is no large demand for this and the user specifically asked to extend the beamer tags.

sunderme:
thanks

PR #3570, opened 2024-03-18
ci: bump GeekyEggo/delete-artifact to v5

muzimuzhi:
See https://github.com/GeekyEggo/delete-artifact/releases/tag/v5.0.0.

Also remove unneeded and deprecated `token` input.


sunderme:
Thanks

PR #3571, opened 2024-03-19
Cwl: minor improvements for `\numberwithin` and `\counter(within|without)`

muzimuzhi:
- The two mandatory arguments that `\numberwithin` from `amsmath` takes are both counter names, not one env and one counter.
  See `texdoc amsmath`, sec. 3.11.1 "Numbering hierarchy"
  image
- Both `\counterwithin` and `\counterwithout` in the LaTeX2e kernel take an optional star and an optional argument since LaTeX2e 2021-11-15.
  See `texdoc ltnews34`, sec. "New argument for `\counterwithin`/`without`"
  image
  The forms where two optional arguments are present, like `\counterwithin*[format]{counter}{within-counter}`, have no real use case so I didn't add them.

### DIFF
--- a/completion/amsmath.cwl
+++ b/completion/amsmath.cwl
@@ -135,8 +135,8 @@ alignedleftspaceyesifneg
 \multlinetaggap#*
 \nobreakdash
 \notag#m
-\numberwithin{env}{counter}
-\numberwithin[format]{env}{counter}
+\numberwithin{counter}{within-counter}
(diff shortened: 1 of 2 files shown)


muzimuzhi:
> * The forms where two optional arguments are present, like `\counterwithin*[format]{counter}{within-counter}`, have no real use case so I didn't add them.

Wait a minute. I may be wrong.

sunderme:
Thanks

PR #3604, opened 2024-04-15
remove duplicate word

octaeder:
in new text


octaeder:
@sunderme I wanna note that advanced,md needs processing to generate the html file. Thanks in advance.

PR #3613, opened 2024-04-18
activate expl3 syntax highlighting for `\ProvidesExplPackage` packages

samcarter:
The new l3 syntax highlighting from your recent commit https://github.com/texstudio-org/texstudio/commit/f14060826c6fa2847761fd3e3ff9e7cfa81f9841 seems to work great for code within `\ExplSyntaxOn ... \ExplSyntaxOff`.

It would be nice to have the same syntax highlighting for l3 packages (they can be recognised by the `\ProvidesExplPackage` instead of `\ProvidesPackage`). If you need an example for testing, `siunitx.sty` is one such package.

My naive attempt is to just use

```
 if(word=="\\ExplSyntaxOn" || word=="\\ProvidesExplPackage"){
```

in `syntaxcheck.cpp`. This seems to work, but I don't know if there is a better solution...

Screenshot 2024-04-18 at 11 33 57

### DIFF
--- a/src/syntaxcheck.cpp
+++ b/src/syntaxcheck.cpp
@@ -1053,7 +1053,8 @@ void SyntaxCheck::checkLine(const QString &line, Ranges &newRanges, StackEnviron
 					word = word + line.mid(tkEnvName.start, tkEnvName.length);
 				}
 			}
-            // special treatment for \ExplSyntaxOn and \ExplSyntaxOff
+            // special treatment for \ExplSyntaxOn, \ExplSyntaxOff
+            // \ProvidesExplPackage and \ProvidesExplClass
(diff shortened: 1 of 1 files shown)


muzimuzhi:
For completeness, maybe also detect `\ProvidesExplClass` and `\ProvidesExplFile`.

samcarter:
@muzimuzhi Good idea! I added it as well.

muzimuzhi:
`\ProvidesExplFile` is also in use, for example the `l3debug.def` loaded by first use of expl3 function `\debug_on:n {...}`.

samcarter:
@muzimuzhi  Added (I don't have syntax colour for .def files, but it works if I test with a different file type)

sunderme:
Thanks

PR #3618, opened 2024-04-24
set editor with all line numbers as default

### LINKED ISSUE #3617: Make line numbers default
Reading #804 I thought to myself "Why isn't this default behavior?".

Make "Show Line Numbers: All" default behavior instead of "No Line Numbers". Even without having "Advanced Options" it is still not that easy to find because of the overall high number of options. Line numbers make lines referencable but more importantly they make it easier to see where a new line starts which I find especially important working with git-versioned documents where each sentence is a new line.

The only reason I can imagine why people wouldn't want line numbers in texstudio is because of small screen sizes. And then I think it would be ok to configure.

New users, collaborating users, lazy user :wink:

octaeder:
this PR resolves #3617
this changes the default of no line numbers to all line numbers (s. Editor options)


sunderme:
No intention to pull this, see feature request

PR #3623, opened 2024-04-28
preview with xelatex

octaeder:
This PR adds Preview with Xelatex (s. Config, Preview option _Command:_). Resolves #3621.

Citation regarding precompile of preamble:
years later, most large packages end up loading something related to fontspec (Koma script classes do, tikz does, unicode-math and fontspec and polyglossia do of course, etc.). As a result dumping isn't very effective at all for LuaLaTeX and XeTeX. – 
by Clément [Dec 15, 2021 at 1:10](https://tex.stackexchange.com/questions/49295/precompile-header-with-xelatex#comment1563009_65357)

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1428,7 +1428,7 @@ void BuildManager::readSettings(QSettings &settings)
 
 	int md = dvi2pngMode;
 #ifdef NO_POPPLER_PREVIEW
-	if (md == DPM_EMBEDDED_PDF || md == DPM_LUA_EMBEDDED_PDF)
+	if (md == DPM_EMBEDDED_PDF || md == DPM_LUA_EMBEDDED_PDF || md == DPM_XE_EMBEDDED_PDF)
 		md = -1;
 #endif
 	if (md < 0) {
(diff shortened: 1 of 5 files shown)


sunderme:
thanks

muzimuzhi:
> Citation regarding precompile of preamble:
> years later, most large packages end up loading something related to fontspec (Koma script classes do, tikz does, unicode-math and fontspec and polyglossia do of course, etc.). As a result dumping isn't very effective at all for LuaLaTeX and XeTeX. –
> by Clément [Dec 15, 2021 at 1:10](https://tex.stackexchange.com/questions/49295/precompile-header-with-xelatex#comment1563009_65357)

`Tikz` doesn't load fonts, but an optimal-sized OpenType (`.otf`) Latin Modern Roman font would be loaded by `\normalsize` used in `size<n>.clo` font size configuration, which is loaded by standard document classes `article.cls`, `report.cls`, or `book.cls`.

For example by default `article.cls` executes `10pt` option which then loads `size10.clo`, and finally requests the `lmroman10-regular.otf` font.

Issue #3361, opened 2023-11-12
text of all but one editor is positioned at line one after starting txs


octaeder:
### Environment
- TeXstudio: current dev
- Qt: 6.5
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior
After restarting txs, the editor text for not active edit tabs should be positioned at the same top line (when I activate them) as before ending txs. (I didn't check what happens in split screen)

### Actual behavior
Active tab from last session is active and text is almost[^1] positioned to the first line from previous session. But text from all other tabs is positioned at the first line.

### How to reproduce
To reproduce I use a large document and open the root document and several documents from the include tree, scroll text some pages and position cursor line somewhere. To ensure that cursor lines are visible I sometimes move the cursor mark to the right one character to the right and back. I also position one cursor line such that it is partly hidden at the lower end. Then close and restart txs. Activate different edit tabs to check.

[^1]: When the cursor line is at the lower end of the editor and is partly not visible, it might be that on starting txs the previous line is shown at the lower end (so cursor line is not visible). Seems that different calculations are used at different places in the code.

octaeder:
Result for fix: when top line in editor was 181 at closing time of txs then it is 180 at next start. This holds for all editors with more lines than can be presented after current line.

PR #3626, opened 2024-05-03
Fix scrolling to top line shouldn’t depend on cursor position

octaeder:
Txs remembers the top line visible in the editor and cursor position when txs is closed (while tab isn’t closed). Thus, txs should be able to reopen the editor with the same line at top and the cursor at the same position. But sometimes this is not the case (s. footnote of #3361). We start our analysis with an editor where line number 300 is the first line visible while line number 1 holds the cursor (s. status bar):

image

After closing txs a look into the session file reveals:
```
"FirstLine": 299,
"Line": 0	
```
Obviously `FirstLine` is the line index counting from zero. Surprisingly, after starting txs again we see a shift in the line numbers:

image

After closing txs we find:
```
"FirstLine": 298,
"Line": 0
```
Scrolling to `FirstLine` is done by `edView->editor->scrollToFirstLine(f.firstLine)` (s. texstudio.cpp). But `scrollToFirstLine` subtracts 1 from the line parameter (s. below), e.g. a line number is expected, not a line index. Thus using `edView->editor->scrollToFirstLine(f.firstLine+1)` should fix the issue. But why not in all cases?

Before scrolling to `FirstLine` the cursor is set using `edView->editor->setCursorPosition(line, col)`. There is an optional 3rd parameter that is true by default, which means it should be scrolled such that the cursor line gets visible. This is useless since we scroll to `FirstLine` anyway. Therefor we will let the 3rd parameter be false.
We have a closer look at:
```
void QEditor::scrollToFirstLine(int l){
    const qreal ls = document()->getLineSpacing();
    const qreal ypos = m_doc->y(l-1);
    const qreal yval = verticalOffset();
    const int ylen = viewport()->height();
    const qreal yend = ypos + ylen;

    if ( ypos < yval )
        verticalScrollBar()->setValue(qFloor(ypos / ls));
    else if ( yend > (yval + ylen) )
        verticalScrollBar()->setValue(qFloor(1. + (yend - ylen) / ls));
}
```
We do some calculations using the definitions: From `yend = ypos + ylen` follows `yend – ylen = ypos`. From `yend > yval + ylen` follows `yend – ylen > yval`. Hence, `ypos > yval`. So, the code could be written as:
```
    if ( ypos < yval )
        verticalScrollBar()->setValue(qFloor(ypos / ls));
    else if ( ypos > yval )
        verticalScrollBar()->setValue(qFloor(1. + ypos / ls));
```
Since `yvar` is 0 in our first example the scrollbar is set to `1 + ypos/ls`. Without the first fix this gives the expected index of 299. To take the fix into account we must omit the addition of 1. By this the code gets symmetric. Would target position `ypos` be less than indicated by the scrollbar value `yval` then the value would be set to `ypos/ls` which equals 299, due to the first fix. Note that the quotient `ypos/ls` is an integer since `ypos` is an integer multiple of `ls`. Instead of `qFloor` we use `qRound` to be bullet proof and get:
```
    verticalScrollBar()->setValue(qRound(ypos / ls));
```

### DIFF
--- a/src/qcodeedit/lib/qeditor.cpp
+++ b/src/qcodeedit/lib/qeditor.cpp
@@ -5815,15 +5815,7 @@ void QEditor::scrollToFirstLine(int l){
 
     const qreal ls = document()->getLineSpacing();
     const qreal ypos = m_doc->y(l-1);
-    const qreal yval = verticalOffset();
-    const int ylen = viewport()->height();
-    const qreal yend = ypos + ylen;
-
-	if ( ypos < yval )
-        verticalScrollBar()->setValue(qFloor(ypos / ls));
-	else if ( yend > (yval + ylen) )
-        verticalScrollBar()->setValue(qFloor(1. + (yend - ylen) / ls));
-
+    verticalScrollBar()->setValue(qRound(ypos / ls));
 }
 
 void QEditor::setCursorSurroundingLines(int s){
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -3598,8 +3598,8 @@ void Texstudio::restoreSession(const Session &s, bool showProgress, bool warnMis
                     col = 0;
                 }
             }
-            edView->editor->setCursorPosition(line, col);
-            edView->editor->scrollToFirstLine(f.firstLine);
+            edView->editor->setCursorPosition(line, col, false);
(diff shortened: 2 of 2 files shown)


sunderme:
thanks

PR #3627, opened 2024-05-04
txs restores wrong top line with folded lines

octaeder:
Fold 13 lines (i.e. they are hidden) as shown here:

image

Then scroll down (down is the important information here) to line 253 such that this line is at top of the editor:

image

Close txs (but not the editor) and find their indices in the session file:
```
"FirstLine": 252,
"FoldedLines": "227",
```
Starting txs the editor shows at top line:

image

But we expect to see line 253 at top.

Txs calculates the position of `FirstLine` before lines are folded. Since 13 lines are hidden by folding, line 253+13=263 moves to top of the editor. S. `QDocument::y(…)` used in `scrollToFirstLine`, which respects folded and wrapped lines by calling `visualLine`.
To fix this the order of position calculation and folding need to be reversed.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -3599,8 +3599,8 @@ void Texstudio::restoreSession(const Session &s, bool showProgress, bool warnMis
                 }
             }
             edView->editor->setCursorPosition(line, col, false);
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #3631, opened 2024-05-08
Fix incorrect file opening from structure with master doc present

joneuhauser:
This PR fixes an issue which occurs when a master document is set.

If there are two open documents and both documents have an include of the same name (say Version1/a.tex -> Version1/b.tex and Version2/a.tex -> Version2/b.tex, with Version1/a.tex as master), I cannot open Version2/b.tex from the structure because it always interprets the path relative to the root. 

In this case, this makes no sense, as the user explicitly said relative to which document the path should be interpreted. 

Checked that the fix also works when opening a non-existing document. Previously, it was created relative to the root document, now it's created at the correct location.

However, I did not get the red color to disappear. The previous code was wrong (it reinterpreted only the current cursor position, but the parent document doesn't even need to be the active tab, but I haven't figured out a replacement.

Also added full paths to the tooltips of sub-items in the structure.

### DIFF
--- a/src/latexdocument.cpp
+++ b/src/latexdocument.cpp
@@ -943,6 +943,7 @@ void LatexDocument::interpretCommandArguments(QDocumentLineHandle *dlh, const in
             newInclude->level = parent && !parent->indentIncludesInStructure ? 0 : lp->structureDepth() - 1;
             fn = removeQuote(fn);
             newInclude->title = fn;
+            newInclude->tooltip = getAbsoluteFilePath(fn,".tex", QStringList(), true);
             QString name=fn;
             name.replace("\\string~",QDir::homePath());
             QString fname = findFileName(name);
@@ -2263,10 +2264,10 @@ QString LatexDocuments::getLogFileName() const
 	}
 }
 
-QString LatexDocuments::getAbsoluteFilePath(const QString &relName, const QString &extension, const QStringList &additionalSearchPaths) const
+QString LatexDocuments::getAbsoluteFilePath(const QString &relName, const QString &extension, const QStringList &additionalSearchPaths, bool ignore_root) const
 {
 	if (!currentDocument) return relName;
-	return currentDocument->getAbsoluteFilePath(relName, extension, additionalSearchPaths);
+	return currentDocument->getAbsoluteFilePath(relName, extension, additionalSearchPaths, ignore_root);
 }
 
 LatexDocument *LatexDocuments::findDocumentFromName(const QString &fileName) const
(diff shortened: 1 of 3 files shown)


sunderme:
this has to be reviewed in detail **after** 4.8 release

sunderme:
any reason, why you set a "explicit root doc" in the first place ? That is usually not necessary.

sunderme:
unfortunately I can't accept this PR:
1) it breaks basic functionality. LaTeX uses relative paths in includes with respect to the root document. Documents with the form "root" , "folder1/inc1" , "folder1/inc2". The path in "inc1" for "inc2" needs to be "folder1/inc2" .
2) Automatic root detect works fine and should work for user's scenario just fine without any changes.
3) explicit root does exactly that. It forces txs to use the "explicit root" for all aspects, here the calculation of absolute path from relative paths.
4) the code does not check "explicit root" mode, hence it changes/breaks functionality for that mode as well
5) there are uncommented changes at the end of "openExternalFile" which likely break general functionality.

So, general recommendation is: just use the normal "automatically detect root"

joneuhauser:
> any reason, why you set a "explicit root doc" in the first place ? That is usually not necessary.

Automatic root detection very rarely works for me (OS or texstudio version don't matter). I've given up on it. At some point (even after a file has already worked for some time), clicking "compile" on a child file gives the error that there is no `\documentclass` (ie. it can't find its root). I could then switch to the root file and compile that, but that moves the PDF viewer back to the beginning. Setting an explicit root is quicker, but then exposes the bug fixed herein.

Maybe that's the bug that should have been fixed instead. But then there's also cases where I can't reasonably expect TexStudio to find the root - e.g. when the include path only follows after a macro expansion, or the same child file is included in two different open root files.

> it breaks basic functionality. LaTeX uses relative paths in includes with respect to the root document. Documents with the form "root" , "folder1/inc1" , "folder1/inc2". The path in "inc1" for "inc2" needs to be "folder1/inc2" .

Okay, I see your rationale here. It makes sense in a weird way? I've used Texstudio since ~10 years and I've never figured out that this might be inteded behavior! 

I see two solutions to this UX bug:

 - actually making the structure hierarchical, so communicating back to the user which root structure was actually discovered and is present for Texstudio right now (so if explicit root is set, everything becomes a child of that node)
 - prevent TexStudio from displaying the structure of multiple root documents in the same structure pane. That is, either, show different structures depending on which file is currently active, or straight up disallow multiple root documents in the same window (multiple windows being open is of course an option then).

sunderme:
I would propose that you find out what impedes the automatic detection from working.
1) option /editor/automatically load included documents needs to be active
2) the root document needs to be open in txs

The tooltip on the document tab bar shows what txs is using as root.
"help/check active completion files" also list all documents which belong to one root.

PR #3633, opened 2024-05-12
Add missing files in sources.pri

Biswa96:
This fixes the following linking error while building with qmake.
    
    ld.lld: error: undefined symbol: AIChatAssistant::AIChatAssistant(QWidget*)
     referenced by texstudio.cpp
     .obj/texstudio.o:(Texstudio::aiChat())


sunderme:
Thanks.
Qmake is not officially supported any more and may be removed.

PR #3645, opened 2024-05-13
Cwls

mbertucci47:
New cwls for `class-ximera`, `class-xourse`, `nm`, `synthslant`, and `typog`. Several updated cwls


sunderme:
thanks

PR #3662, opened 2024-05-20
Typos

mbertucci47:
Fixes a few typos, mostly from the manual and changelog. Found with [codespell](https://pypi.org/project/codespell/).


sunderme:
the html files for the manual are generated automatically.
I pull this as is, but those do not need changing.

PR #3665, opened 2024-05-21
fix invisible zoom buttons in pdf viewer

samcarter:
With qt 6.7, the zoom in and zoom out buttons of pdf viewer were invisible:

Screenshot 2024-05-21 at 16 27 49


samcarter:
> I have fixed it here with my methodology. Classic icon was also faulty.

Thanks for the quick fix!

PR #3668, opened 2024-05-22
only offer Windows 11 style under Win 11 or higher

octaeder:
Running txs config under Win10 with debug shows: `QWindows11Style: Style is only supported on Windows11 and above`. It's not clear why `QStyleFactory::keys()` returns this value under such conditions. This PR fixes this.

Before:
grafik

After:
grafik


sunderme:
thanks

PR #3674, opened 2024-05-24
hide first toolbar separator for embedded pdf viewer

octaeder:
The embedded pdf viewer has no build action in the toolbar, while the windowed pdf viewer's toolbar shows this action followed by a separator followed by the external viewer action. Thus the toolbars begins respectively as follows:

grafik embedded
grafik windowed

This leads to a different behavior when you move the mouse pointer further to the left. In the second case the size all cursor
cursor-sizeall appears immediately indicating that the toolbar can be moved. In the first case the arrow pointer doesn't change even so the cursor leaves the box of the external viewer's button, because the cursor moves over the separator:

grafik

So, in the embedded case the build button and the separator must be hidden:

grafik


sunderme:
Thanks

PR #3675, opened 2024-05-24
Cwls

mbertucci47:
New cwls for `catppuccinpalette`, `colorblind`, `rigidnotation`, `semesterplannerlua`, and `skillicons`. A few updated cwls


sunderme:
thanks

Issue #3222, opened 2023-07-27
PDF page number text box is very thin


TimeTravelPenguin:
### Environment
- TeXstudio: 4.5.2
- Qt: 6.5.0
- OS: Windows 11
- TeX distribution: miktex

### Problem
The PDF viewer's page number input box is very thin as shown in the screenshot

![image](https://github.com/texstudio-org/texstudio/assets/35803538/914029a3-bcb9-42d2-a008-c7059b039bf4)

GUI scaling doesn't seem to fix the issue, and I cannot find any options that may resolve this issue.

muzimuzhi:
Is this behavior specific to the dark theme in use? Then maybe related to #2186.

TimeTravelPenguin:
I checked and it only occurs with the styles "Adwaita (txs)" and "Adwaita Dark (txs)". Neither the colour scheme nor the icon theme changes this. So, yes, that issue appears to be the same one that I am dealing with.

octaeder:
@TimeTravelPenguin Should be fixed, s. #2186

Issue #3048, opened 2023-03-30
Search box text and PDF page number only partially visible


foerster-rwth:
### Environment

    TeXstudio: TeXstudio 4.5.1 (git 4.5.1)
    Qt: Using Qt Version 6.4.2, compiled with Qt 6.4.2 R
    OS: Windows 10 Education
    TeX distribution: MiKTeX

### Expected behavior

The text entered in the search box should be completely visible. Page numbers of the PDF should be completely visible.

### Actual behavior

They're not. See screenshots.

![Untitled-2](https://user-images.githubusercontent.com/72136104/228926528-3c7608d0-d095-49cc-a666-2abec16069c0.png)

![Untitled-1](https://user-images.githubusercontent.com/72136104/228926570-c77f5eeb-949f-451c-a9ab-87ab090b27a0.png)

### How to reproduce

Bug is permanent. Just search for something, or have a longer PDF open.

octaeder:
Issues regarding find and zoom should be fixed (by Qt version?):

![grafik](https://github.com/texstudio-org/texstudio/assets/102688820/5b38d9d1-17d7-4a0a-a240-fd85ee8de71c)

![grafik](https://github.com/texstudio-org/texstudio/assets/102688820/86ba1af4-d7c8-44c5-8763-513b22d3a03f)

Issue #3046, opened 2023-03-30
Incomplete display of page numbers under Orion Dark Style


earlSagrada:
### Environment
- TeXstudio:  4.5.1 (git 4.5.1)
- Qt:  Using Qt Version 6.4.2, compiled with Qt 6.4.2 R
- OS:  Microsoft Windows 11 Home Edition
- TeX distribution:  MiKTeX

### Expected behaviour
Display of full and exact page number, in the built-in PDF viewer, under Orion Dark Style.

### Actual behaviour

Only the last digit of the page number is displayed. Increasing font size can help a bit.

_Font size: 7, the page number is not visible._
![image](https://user-images.githubusercontent.com/39267845/228847916-eeb4d555-911a-4766-874b-f8e72dbc5153.png)

_Font size: 12, the first digit of page number starts to show._
![image](https://user-images.githubusercontent.com/39267845/228848052-82223ad9-20e0-49f5-ad0c-6dc0feca50b7.png)

### How to reproduce
Simply change the 'General - Appearance - Style' to 'Orion Dark'. Then view PDFs with more than 10 pages.

sunderme:
duplicate #2816

PR #3677, opened 2024-05-27
fix line number visibility in Orion dark

### LINKED ISSUE #2186: Misaligned & Invisible Page Number in Dark Theme
### Environment
- TeXstudio: 4.2.2
- Qt: 6.2.3
- OS: WIndows 10
- TeX distribution: miktex

### Issue

The page numbering in the "go to" part of the internal PDF viewer has the numbers misaligned.

image

This is hardly a big deal in the picture above. It also successfully rescales with more page.

image

However, there is a serious issue when swapping to dark themes. The first two pictures below are for Orion Dark and the other two for Adwatia Dark (txs). The page numbers should be 15 and 298.

image

image

image

image

Orion Dark has a similar issue with the scaling number at the bottom, but AD doesn't.

image

image

I do want to use a Dark Mode, but these issues are a bit annoying...

octaeder:
and alignment QLineEdit vs. QTextEdit. This PR resolves #2186, #3046, #3048, and #3222.

#### Details:
Image shows misalignment (ex. from macro browser):

grafik
old (left) vs. new

Reducing size of padding helps solving the line number issue. Here is a line up of state before fix for several styles (per row), each with a document of 25 pages (at left) and a document of 100.001 pages (at right):

grafik
Be aware of the line number issue with Orion.

With the fix it looks like this:

grafik

Note: The extra amount of space needed for the width of leCurrentPage should be constant (width of borders, paddings) and not depend on the width of the characters of the font. Therefore I changed the formula.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -3714,12 +3714,12 @@ retryNow:
 			pdfWidget->setFocus();
 
 		// set page viewer only once
-        int maxDigits = 1 + qFloor(log10(pdfWidget->realNumPages()));
-		//if (maxDigits < 2) maxDigits = 2;
+		int maxDigits = 1 + qFloor(log10(pdfWidget->realNumPages()));
 		leCurrentPage->setMaxLength(maxDigits);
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

PR #3704, opened 2024-06-21
Cwls

mbertucci47:
New cwls for `linearregression`, `spelatex`, and `standardsectioning`. Several updated cwls


sunderme:
thanks

PR #3747, opened 2024-07-29
Fixing single-instance not working with Qt5 ?

### LINKED ISSUE #3743: TXS always opens file in new instances
### Environment
- TeXstudio: 4.8.1
- Qt: 5.15.3
- OS: KUbuntu 22.04
- TeX distribution: texlive

### Expected behavior
When opening ``.tex`` from the File manager or by invoking ``texstudio bla.tex`` from the command line, an already running instance of TXS should open the file in a new tab.

### Actual behavior
Each invocation starts a new TXS instance. This seems to happen since a recent update.

### How to reproduce
For example, run ``texstudio somefile.tex`` twice.

Note that opening a file using TXS GUI still correctly opens the file within that instance, and running ``texstudio file1.tex file2.tex`` still creates one instance that opens both files.

ysalmon:
For some reason, any of these two lines confuses the single-instance mechanism and makes TXS open each file in a new instance when invoked from the desktop environment on a file or via CLI ``texstudio somefile.tex``.

I tried to say ``new QFileSystemModel()`` instead of ``new QFileSystemModel(this)`` on line 733, but to no avail.

(insert picture of dog in front a a keyboard with caption "I have no idea what I am doing")

Would fix #3743.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -731,12 +731,15 @@ void Texstudio::setupDockWidgets()
     if(!dock){
         fileView=new QTreeView();
         fileExplorerModel = new QFileSystemModel(this);
-        //fileExplorerModel->setRootPath(QDir::currentPath());
+		QString rootDir = QDir::currentPath();
+		if (rootDir == "/tmp")
+			rootDir = "/";
+        fileExplorerModel->setRootPath(rootDir);
         fileView->setModel(fileExplorerModel);
         fileView->setColumnHidden(1,true);
         fileView->setColumnHidden(2,true);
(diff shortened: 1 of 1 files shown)


sunderme:
okay, your changes seem to work and look somewhat reasonable for a work-around

ysalmon:
I should check if this is just ``/tmp`` or more generally files on different mountpoints. My current machine is impractical to test this, but in the next days I should be able to test on a machine where ``/opt`` is a mountpoint.

The underlying cause seems to be a bug in Qt. Maybe it is corrected in Qt6 ?

sunderme:
thanks

PR #3748, opened 2024-07-29
Cwls

mbertucci47:
New cwls for `bonum-otf`, `doibanner`, `ltcaption`, `pgfplotsthemebeamer`, and `syntax`. Several updated cwls


sunderme:
thanks

PR #3752, opened 2024-08-05
Skip codesign if required secret is not set

muzimuzhi:
Currently in workflow `cd.yml` job "win10 build (msys2)", to sign Win binary a SignPath API token stored as secret is used. This makes any (if not all) pushes to forks fail the job, because usually such secret is not set in forks (and it makes tiny sense to set one).
https://github.com/texstudio-org/texstudio/blob/a1181170e32a3f147bccdd6fdadce2049610b7c1/.github/workflows/cd.yml#L80

This PR skips code sign in forks, thus makes the CD job "win10 build (msys2)" pass.

See a passed [job run #10254739600](https://github.com/texstudio-org/texstudio/actions/runs/10254739600) with this change in my fork.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -72,9 +72,20 @@ jobs:
           echo "TXS_VERSION=${TXS_VERSION}">> $GITHUB_OUTPUT
           echo "GIT_VERSION=${GIT_VERSION}">> $GITHUB_OUTPUT
       
+      - name: Detect if codesign
+        run: |
+          # Nonexistent context property evaluates to an empty string.
+          # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#about-contexts
+          if [ -z "${{ secrets.SIGNPATH_API_TOKEN }}" ]; then
+            echo "IF_CODESIGN=false" >> "$GITHUB_ENV"
+            echo "::notice file=.github/workflows/cd.yml,title=Codesign skipped::Secret "SIGNPATH_API_TOKEN" not set."
+          else
+            echo "IF_CODESIGN=true" >> "$GITHUB_ENV"
+          fi
+
       - name: Codesign binary
         id: sign_binary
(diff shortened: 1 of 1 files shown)


octaeder:
@muzimuzhi I have been receiving emails about failed jobs for six weeks now. Thanks for fixing this 👍

muzimuzhi:
It's more flexible now: if the required secret is set, do codesign; otherwise skip codesign and raise a notice.

sunderme:
thanks

.github/workflows/cd.yml
```diff
@@ -72,9 +72,20 @@ jobs:
           echo "TXS_VERSION=${TXS_VERSION}">> $GITHUB_OUTPUT
           echo "GIT_VERSION=${GIT_VERSION}">> $GITHUB_OUTPUT
       
+      - name: Detect if codesign
+        run: |
+          # Nonexistent context property evaluates to an empty string.
+          # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#about-contexts
+          if [ -z "${{ secrets.SIGNPATH_API_TOKEN }}" ]; then
+            echo "IF_CODESIGN=false" >> "$GITHUB_ENV"
+            echo "::notice file=.github/workflows/cd.yml,title=Codesign skipped::Secret "SIGNPATH_API_TOKEN" not set."
```
muzimuzhi, line 81: A little styling improvement: the double quotes enclosing `SIGNPATH_API_TOKEN` should be escaped. Not something important.

`s/"SIGNPATH_API_TOKEN"/\"SIGNPATH_API_TOKEN\"/`

```diff
-echo "::notice file=.github/workflows/cd.yml,title=Codesign skipped::Secret "SIGNPATH_API_TOKEN" not set."
+echo "::notice file=.github/workflows/cd.yml,title=Codesign skipped::Secret \"SIGNPATH_API_TOKEN\" not set."
```

PR #3753, opened 2024-08-06
Suppress Homebrew "already installed" warnings

muzimuzhi:
To make the Annotations list less distracted, this PR suppresses 17 Homebrew warnings which all look like
```
cairo 1.18.0 is already installed and up-to-date. To reinstall 1.18.0, run: brew reinstall cairo
```
raised by `cd.yml` workflow runs.

Example runs
- Before, 17 warnings and possibly 1 notice added by #3752
  image
  https://github.com/texstudio-org/texstudio/actions/runs/10260913774
- After, 0 warnings and possibly 1 notice
  image
  https://github.com/muzimuzhi/texstudio/actions/runs/10261136172


sunderme:
Thanks

Issue #1752, opened 2021-09-10
Right click - show file in explorer


dr-tb:
Please add "Show .tex file in explorer" somwhere. I think the best place would be on drop down menu when you rightclick the tab with the tex document you are working on.

This would save users so much time when searching for names of pictures and aditional files, bibliographies. We all know there are 2 types of people, one that has 1243234 windows open and the other that has 1. Those with 1 will be very gratefull. 

### Describe the feature and the current behavior/state
There is no feature

### Who will benefit with this feature?
Everyone that has supplementary files in root folder of the document (or relative to those).

PR #3770, opened 2024-08-19
Add show the currently opened .tex document in file explorer (#1752)

frostybee:
This PR includes an addition of two menu items  to the tab's context menu of the current document being edited (see below).
Now it's possible to copy the document's file path, and show the document in its containing folder from the tab editor's context menu.

image

This PR resolves: #1752


sunderme:
Thanks

PR #3784, opened 2024-09-02
Cwls

mbertucci47:
New cwls for `domaincoloring`, `enverb`, `interlinear`, `luamml-demo`, and `suanpan-l3`. Several updated cwls


completion/collcell.cwl
```diff
@@ -16,7 +16,6 @@ norobustcr
 #endif
 
 \collectcell
-\collectcell{cmd}#Sd
```
muzimuzhi, line 19: Seems to me `\collectcell{<cmd>}` is the form that should be kept.

The usage of `\collectcell` (in tabular preamble) is always followed by a command
```tex
>{\collectcell\usermacro}c<{\endcollectcell}
```

mbertucci47: Thanks. Meant to remove the `#d` definition part but you're right, it has a mandatory argument.

sunderme:
Thanks

PR #3800, opened 2024-09-20
menu LaTeX/AMS packages

### LINKED ISSUE #3798: Add amsthm to the AMS packages menu item
When using the AMS Packages from LaTeX menu, the following two lines of code are inserted:

```
\usepackage{amsmath}
\usepackage{amssymb}

```
But IMHO (and based on this [TEX.SE](https://tex.stackexchange.com/a/32103/14757) answer) it would be good to add also the `amsthm` package. 

Regards.

octaeder:
This PR resolves #3798. The changed menu action will add package mathtools instead of amsmath, and will also add amsthm. Package `amssymb` will still be present (no change).


sunderme:
thanks

PR #3864, opened 2024-10-31
update cwls for latex release

mbertucci47:
Updates latex-dev.cwl for new latex release. Also moves textcomp commands out of textcomp.cwl and into latex-document.cwl since these are now defined in the format (this actually happened a couple years ago). Now textcomp.cwl is just a stub.


sunderme:
Thanks

PR #3871, opened 2024-11-09
cwl: add `\theH<counter>` to latex-dev.cwl

muzimuzhi:
For now `\theHchapter` is only added to the standard `report` and `book` classes, and kept in `hyperref.cwl`.

Not sure if it's wanted to add all `\theH<counter>` for all new counters defined by `\newcounter` in packages and classes. If so that would take some effort.


sunderme:
Thanks

PR #3873, opened 2024-11-11
always highlight log lines with Process started

octaeder:
This PR aims to improve highlighting of the message log. When you start a LaTeX compile the log shows a green line with the build command. This is because the line contains `.tex` from the file name given (including extension):

grafik

But not in all cases a command uses the file name including extension. For example, when you run Biber all log lines are black:

grafik

So this PR will change the color to green for all lines, which would be black under the same conditions, but contain `Process started:`:

grafik

This helps finding quickly the start of the next process.


sunderme:
Thanks

PR #3911, opened 2024-12-14
add maximize button to packages help dialog

octaeder:
@tmelorc mentioned this as a wish [here](https://github.com/texstudio-org/texstudio/pull/3478#issuecomment-1897383484)

grafik

I didn't made the Minimize Button available, since you can't switch to the main txs window. So this could prevent some confusion.


sunderme:
thanks

Issue #3896, opened 2024-12-03
autocompletion of expl3 commands


mbertucci47:
### Describe the feature and the current behavior/state

Correct highlighting for expl3 code was somewhat recently added to TXS, but there is currently no autocompletion for expl3 commands. Generating an appropriate cwl would be easy since LaTeX Workshop has a script for making a json file of expl3 commands: https://github.com/James-Yu/LaTeX-Workshop/blob/master/data/packages/expl3.json

A quick test shows that commands in cwls can contain underscores and colons, so that part is fine. The only thing that would need extra code is making the autocompletion active only inside `\ExplSyntaxOn...\ExplSyntaxOff` or in a file that declares `\ProvidesExplPackage`, etc.

### Who will benefit with this feature?
Anyone who write expl3 code in TXS.

muzimuzhi:
> Generating an appropriate cwl would be easy since LaTeX Workshop has a script for making a json file of expl3 commands: https://github.com/James-Yu/LaTeX-Workshop/blob/master/data/packages/expl3.json

It even uses `.cwl` (generated by [`latex3command.py`](https://github.com/James-Yu/LaTeX-Workshop/blob/master/dev/latex3command.py)) as an intermediate representation, though currently only the function or variable name is extracted (or, can be reliably extracted?).

sunderme:
it is not strictly necessary that txs restricts the completion wintin `\ExplSyntaxOn...\ExplSyntaxOff`
So, we can just add the commands first, the user can activate them generally and one can add the extra restriction into txs later.

muzimuzhi:
@sunderme Ah yes! Why I didn't realize it earlier?

Tried with cwl entries
```
\tl_new:N %<tl var%>
\tl_const:Nn %<tl const%> {%<content%>}
```
and they do work (currently not necessarily in between `\ExplSyntaxOn ... \ExplSyntaxOff`).
![image](https://github.com/user-attachments/assets/a6ef3b2b-6324-4bd1-bd19-151d76172b6f)

sunderme:
so, we can close this for now ?

mbertucci47:
I think having automatic activation inside expl3 environments is important to this issue as outside such environments, they just pollute the autocompleter. So I’d vote not to close

mbertucci47:
[...]

PR #3915, opened 2024-12-16
Add expl3-commands.cwl with arg specs

mbertucci47:
This adds a cwl for expl3 commands as mentioned in https://github.com/texstudio-org/texstudio/issues/3896. Several things don't work when commands have `_` and `:` in the name. For example, I couldn't get keyval completion to work, and something like
```
\my_command:Nn{cmd}{def}#d
```
does not cause `\my_command:Nn \foo {bar}` to add `\foo` to the completion list.

Suggestions welcome (@muzimuzhi). Could probably be mentioned in the documentation. Ideally this cwl would be automatically activated inside `\ExplSyntaxOn...\ExplSyntaxOff` and not active elsewhere, but for right now the only way to activate it is manually in the TXS configuration. Note we do not want it automatically activated by `\usepackage{expl3}` since that is a legacy package and expl3 is now always loaded by the format.


completion/expl3-commands.cwl
```diff
@@ -0,0 +1,4322 @@
+# expl3 commands, l3kernel 2024-11-02
+# command list generated by https://github.com/James-Yu/LaTeX-Workshop/blob/master/dev/latex3command.py
+# argument specifications added by Matthew Bertucci 2024/12/16
+
+\bitset_addto_named_index:Nn %<⟨bitset var⟩%> {%<⟨name1=index1,name2=index2,...⟩%>}
+\bitset_clear:N %<⟨bitset var⟩%>
+\bitset_clear:c {%<⟨bitset var name⟩%>}
```
muzimuzhi, line 7: Maybe we want a consistent use (or dropping) of `⟨...⟩` brackets. Not a blocker.

<details><summary>Stats of uses of <code>⟨...⟩</code></summary>
<p>

```
$ rg -F '⟨' --count
metre.cwl:17
exerquiz.cwl:3
physics2.cwl:12
pdfextra.cwl:1
eqexam.cwl:9
textopo.cwl:9
semantex.cwl:1
ajmacros.cwl:4
tikzlibrarybraids.cwl:2
plaintex.cwl:133
hiero.cwl:2
wargame.cwl:10
plext.cwl:1
expkv-cs.cwl:10
china2e.cwl:1
luatexja-ajmacros.cwl:4
qcircuit.cwl:1
babel.cwl:10
tikzlibraryext.shapes.circlearrow.cwl:1
texmate.cwl:4
ptex.cwl:45
nath.cwl:6
class-elteikthesis.cwl:5
longnamefilelist.cwl:2
tikzlibraryceltic.cwl:1
tikzlibraryext.transformations.mirror.cwl:2
tex.cwl:12
abspos.cwl:2
xskak.cwl:6
namedef.cwl:2
yathesis.cwl:109
mdwtab.cwl:7
class-iopart.cwl:2
prettytok.cwl:3
algxpar.cwl:4
doafter.cwl:1
easybase.cwl:8
optex.cwl:86
```

</p>
</details>

mbertucci47: I'm a fan of using the angle brackets with "undelimited" arguments. To me `⟨bitset var⟩` signifies "insert something here" that a merely greyed out `bitset var` does not. Since most document commands have delimited arguments, they are not necessary, but I agree consistency would be better. Not sure how easy a mass change would be since I think the brackets need to be inside `%<...%>` which would break a lot of cwl argument types.

muzimuzhi: I agree the unbraced arguments need angle brackets more than braced ones. So, I'm OK to just keep them as is.

sunderme:
thanks

PR #3916, opened 2024-12-18
add LaTeX interfaces to expl3-commands.cwl

mbertucci47:
This adds the kernel expl3 interfaces to LaTeX modules to expl3-commands.cwl.


sunderme:
thanks

PR #3917, opened 2024-12-19
TexDoc dialog: check all packages option without packages present

octaeder:
This PR brings back following functionality (discussion https://github.com/texstudio-org/texstudio/pull/3478#issuecomment-1915874683, lost by 277e8c87850b01deab0c7ebb8062a1ae8aacf434): When txs starts without any documents the Package help (TexDoc) dialog should automatically check option `all packages`. Otherwise you have to do so manually before you can search anything. As a side effect this fixes that the packages list is emtpy although the documentation of a0poster is presented.

### DIFF
--- a/src/help.cpp
+++ b/src/help.cpp
@@ -18,17 +18,25 @@ Help::Help(QObject *parent): QObject(parent),texDocSystem(0)
  */
 void Help::execTexdocDialog(const QStringList &packages, const QString &defaultPackage)
 {
-    TexdocDialog dialog(nullptr,this);
+	bool lastState;
+	lastState = mShowAllPackages;
+	TexdocDialog dialog(nullptr,this);
 	dialog.setPackageNames(packages);
+	if (packages.count()==0) {
+		mShowAllPackages = true;
+	}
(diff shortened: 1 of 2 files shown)


sunderme:
if you open texsdoc once without a document, the "show all" will stick, even if it was not changed by the user.
I would like to see that the mShowAll is only stored if it was a user choice, i.e. that it was not overriden when there is no package list.

src/help.cpp
```diff
@@ -18,17 +18,23 @@ Help::Help(QObject *parent): QObject(parent),texDocSystem(0)
...
-    }else{
-        dialog.setShowAllPackages(mShowAllPackages);
-    }
+	}
 	if (dialog.exec()) {
 		viewTexdoc(dialog.selectedPackage());
 	}
-    mShowAllPackages=dialog.showAllPackages();
+	if (packages.count()==0) {
+		mShowAllPackages = false;
```
sunderme, line 34: It should not force mShowAllPackages to false either. It just should not change the previous state which maybe true or false.

octaeder: @sunderme build is running

sunderme:
thanks

PR #3918, opened 2024-12-21
use filter for searching preferred package

octaeder:
When a package name is selected in the text then starting the Packages Help dialog searches and selects the given package. However, this can be done using the recently introduced package filter.


sunderme:
thanks

PR #3921, opened 2024-12-22
add expl3 package interfaces

mbertucci47:
New and updated cwls after recent support for expl3 completion.

### DIFF
--- a/completion/GS1.cwl
+++ b/completion/GS1.cwl
@@ -1,5 +1,5 @@
 # GS1 package
-# Matthew Bertucci 11/18/2021 for v23
+# Matthew Bertucci 2024/12/21 for v23
 
 #include:rule-D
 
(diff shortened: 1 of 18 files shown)


sunderme:
thanks

completion/GS1.cwl
```diff
@@ -18,4 +18,11 @@ code=#EAN-8,EAN-13
...
-#endkeyvals
\ No newline at end of file
+#endkeyvals
+
+# expl3 interface
+\GS_cut_EAN_control_digit:N %<⟨seq var⟩%>#/%expl3
+\GS_set_code_digit_seq:Nn %<⟨seq var⟩%> {%<⟨tokens⟩%>}#/%expl3
+\GS_set_EAN_control_digit:N %<⟨seq var⟩%>#/%expl3
+\GS_use_as_EAN_barcode:N %<⟨seq var⟩%>#/%expl3
+\int_set_to_EAN_control_digit:NN %<⟨integer⟩ ⟨seq var⟩%>#/%expl3
```
muzimuzhi, line 28: Maybe
```
\int_set_to_EAN_control_digit:NN %<⟨integer⟩%> %<⟨seq var⟩%>#/%expl3
```
is better, but I didn't check if it's supported or not.

mbertucci47: If the user does not use placeholders, that syntax leaves an unnecessary extra space after the command

PR #3927, opened 2024-12-30
move action continuous to submenu Grid (PDF-Viewer)

octaeder:
grafik

The group of actions from `Zoom in` up to `Fit to Window` are all related to zooming/size. On the other hand action `continuous` is related to the way the pages are organised as it is for all other actions in menu `Grid`. Furthermore I want to use the new `Grid` menu for the embedded viewer (s. #222).


sunderme:
there is no real gain here.
I agree that logically the menu item may be placed in the sub-menu. 
But since it has been there for years and there is no real advancement by moving it (actually making it a bit harder to find), it should stay where it is.

PR #3928, opened 2024-12-31
fix pdf-viewer scroll bar

### LINKED ISSUE #2463: vert. scrollbar empty or missing in pdf-viewer when toggling continuous mode
### How to reproduce

1. Use a new setup of txs (i.e. remove files in config folder). Open a document with several pages and open the pdf in the windowed (internal) pdf-viewer. By Default options _continuous_ and _fit to width_ are set and the viewer shows a scrollbar and its slider allows scrolling through all pages. Narrow down the width of the viewer such that at least a full page fits the window (for me this is already the case). Uncheck continuous in the View menu.
2. Be aware that now any scrolling is used if a single page doesn't fit the height of the viewer to adjust the pages position (even so we could need one more to scroll through the pages as with continuous mode, but I think we can't have both). Please observe that the scrollbar is still present but _the slider is missing_. It is correct that there is no slider (and thus no scrollbar is needed at all) since we can see a full page (s. 1.).
3. If we would check continuous option again we would end up as at the beginning. But instead press button fit to window. The scrollbar disappears as we would expect. But now check continuous and see that the scrollbar is missing. If you press fit to width the scrollbar appears again (with slider). Now you can repeat this (uncheck continuous, slider is missing, ...).

If you would use fit to page in 1. then use fit to width then fit to page again in 3. to reproduce.

### Environment
- TeXstudio: 4.2.3 and releases after that
- Qt: 6
- OS: Windows 10
- TeX distribution: miktex

### Expected behavior
scrollbar should be present with slider, otherwise no scrollbar is needed. If a scrollbar is needed it should be present.

octaeder:
This fixes following (and may also resolve #2463):

1.	Start windowed viewer showing a pdf with 5 or more pages
2.	Set checked 100%
3.	Clear continuous
4.	Set checked fit to width
5.	Choose window size such that no vertical scroll bar is necessary:
    grafik
6.	Set checked continuous, as expected a vert. scroll bar appears:
    grafik
7.	Clear continuous, other than expected the scroll bar isn’t removed:
    grafik
8.	Clear fit to width, size doesn’t fit exactly (but as expected):
    grafik
9.	Set checked fit to width, now result is as expected in 7.:
    grafik

Repeating the procedure from 1. to 8. but using fit to window instead of fit to width is even worth:
At 6., 7. and 8. no vert. scroll bar appears (but size always fits). Skip 9.

10. Set checked  continuous, scroll bar appears:
     grafik
11. Set checked fit to window, some resizing happens:
     grafik
12. Clear continuous, other than expected the scroll bar isn’t removed:
     grafik
13. Same as 8.
14. Same as 9.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -2149,6 +2149,7 @@ void PDFWidget::setGridSize(int gx, int gy, bool setAsDefault)
 	getScrollArea()->goToPage(realPageIndex);
 	if (pi == realPageIndex)
 		reloadPage();
+	emit changedScaleOption(scaleOption);
 	//update();
 }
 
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

PR #3929, opened 2025-01-01
pdf-viewer: fix Custom Grid preset values

octaeder:
since continuous mode changes current values the dialog needs to be preset with the original (config) values


sunderme:
thanks

PR #3931, opened 2025-01-03
make menu entries Show embedded PDF large/small consistent

octaeder:
This PR adresses following two menu entries:

grafik

Both actions are always enabled even when there is no pdf-viewer at all or the viewer is not embedded. When the viewer is emedded exactly one of the entries should be enabled. So we have three cases:

+ no (embedded) viewer:
  grafik
+ small embedded viewer:
  grafik
+ large embedded viewer
  grafik

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -1403,9 +1403,8 @@ void Texstudio::setupMenus()
         }
     }
 
-	newManagedAction(menu, "enlargePDF", tr("Show embedded PDF large"), SLOT(enlargeEmbeddedPDFViewer()));
-	newManagedAction(menu, "shrinkPDF", tr("Show embedded PDF small"), SLOT(shrinkEmbeddedPDFViewer()));
-
+	newManagedAction(menu, "enlargePDF", tr("Show embedded PDF large"), SLOT(enlargeEmbeddedPDFViewer()))->setVisible(false);
+	newManagedAction(menu, "shrinkPDF", tr("Show embedded PDF small"), SLOT(shrinkEmbeddedPDFViewer()))->setVisible(false);
 	newManagedAction(menu, "closeelement", tr("Close Element"), SLOT(viewCloseElement()), Qt::Key_Escape);
 
 	menu->addSeparator();
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -1404,9 +1404,9 @@ void Texstudio::setupMenus()
     }
 
 	act=newManagedAction(menu, "enlargePDF", tr("Show embedded PDF large"), SLOT(enlargeEmbeddedPDFViewer()));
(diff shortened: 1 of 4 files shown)


src/texstudio.cpp
```diff
@@ -1403,9 +1403,8 @@ void Texstudio::setupMenus()
         }
     }
 
-	newManagedAction(menu, "enlargePDF", tr("Show embedded PDF large"), SLOT(enlargeEmbeddedPDFViewer()));
-	newManagedAction(menu, "shrinkPDF", tr("Show embedded PDF small"), SLOT(shrinkEmbeddedPDFViewer()));
-
+	newManagedAction(menu, "enlargePDF", tr("Show embedded PDF large"), SLOT(enlargeEmbeddedPDFViewer()))->setVisible(false);
```
sunderme, line 1406: please make two lines out of this, i.e 
act=...
act->setVi...

src/texstudio.cpp
```diff
@@ -1403,9 +1403,8 @@ void Texstudio::setupMenus()
...
+	newManagedAction(menu, "shrinkPDF", tr("Show embedded PDF small"), SLOT(shrinkEmbeddedPDFViewer()))->setVisible(false);
```
sunderme, line 1407: please make two lines out of this, i.e 
act=...
act->setVi...

src/texstudio.cpp
sunderme: and the approach seems to be a bit indirect.
You want to change menu entries in texstudio.cpp level, the changes are triggered there (even embedding/windowing over Texstudio::runInternalPdfViewer)

src/texstudio.cpp
octaeder: Just in case that there was  no automatic notification, I want to note that I changed the two lines.

sunderme:
Did you see my other comment ?
` ` `
and the approach seems to be a bit indirect.
You want to change menu entries in texstudio.cpp level, the changes are triggered there (even embedding/windowing over Texstudio::runInternalPdfViewer)
` ` `

octaeder:
There are two pairs of similar actions, one of them connected to a button of PDFdocument. So it seems best to me to use the ui changes to setup the menu and the button.

sunderme:
I have looked through the code.
pdfdocument emits triggeredShrink/triggeredEnlarged which are then handled in TexStudio::

Please rewrite the code that the menu change is done in the TexStudio class only.

Secondly, the menu should be disabled, not hidden. The menu entries are always a good learning source for shortcuts.

octaeder:
Thanks for the hints. I made a new commit.

octaeder:
and updatet first commit (images etc.) of th PR.

sunderme:
thanks

PR #3933, opened 2025-01-06
load expl3-commands.cwl in latex-dev.cwl

mbertucci47:
With the addition of the extended environment cwl checking, it now makes sense to load expl3-commands.cwl by default. I picked latex-dev.cwl since this is on the programming side of things but that's loaded by latex-document.cwl anyways. Also updated simpleicons.cwl.


sunderme:
thanks

PR #3942, opened 2025-01-10
make grid options and two others available to the embedded pdf-viewer

### LINKED ISSUE #222: Enhancement: Access to grid view, while viewer is embedded.

octaeder:
Happy new year!

This PR resolves #222 (enhancement request). The changes add tool button `Grid` to the tool bar of the embedded pdf-viewer. The tool button offers known options from the main menu `View` of the windowed pdf-viewer. The button menu looks like this:

grafik

This may often allow working without the windowed pdf-viewer. For example, you could choose grid `2x1` and option `Fit to Window`. If you want to take a closer look at the pdf, you can expand the viewer to the left (and later on shrink it to the right) with the button Enlarge Viewer grafik to the right of the tool bar. On large screens it may suffice to enlarge the texstudio window.

gridAnimation

Grid4x1continuous

Following assertions hold:

1. A new embedded viewer (or a windowed viewer getting embedded) always sets the options as shown in the image (including Fit to Width).  These are the ones used before this PR (c.f. [#3554#issuecomment-2001989472](https://github.com/texstudio-org/texstudio/pull/3554#issuecomment-2001989472)).
2. Only the windowed viewer saves and restores previously set options.
3. These options of embedded and windowed viewer do not interfere.

Previous PR is #3554.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -896,7 +896,7 @@ void PDFWidget::setPDFDocument(PDFDocument *docu)
 	pdfdocument = docu;
 }
 
-void PDFWidget::setDocument(const QSharedPointer<Poppler::Document> &doc)
+void PDFWidget::setDocument(const QSharedPointer<Poppler::Document> &doc, bool embedded)
 {
 	pages.clear();
 	document = doc;
@@ -906,7 +906,10 @@ void PDFWidget::setDocument(const QSharedPointer<Poppler::Document> &doc)
 
 	if (!document.isNull()) {
 		docPages = document->numPages();
-		setSinglePageStep(globalConfig->singlepagestep);
+		if (embedded)
+			setSinglePageStep(false);
+		else
+			setSinglePageStep(globalConfig->singlepagestep);
 	} else
 		docPages = 0;
 #ifdef MEDIAPLAYER
@@ -2098,10 +2101,14 @@ void PDFWidget::setPageOffset(int offset, bool setAsDefault, bool refresh){
 		pageOffset = offset;
 	else {
 		pageOffset = gridx - 1;
-		globalConfig->pageOffset = pageOffset;
 	}
-	if (!setAsDefault)
-		globalConfig->pageOffset = pageOffset;
+	if (!setAsDefault) {
+		bool embedded = pdfdocument->embeddedMode;
+		if (embedded)
+			pageOffsetEmbedded = pageOffset;
+		else
+			globalConfig->pageOffset = pageOffset;
+	}
 
 	if (!refresh)
 		return;
@@ -2132,16 +2139,23 @@ int PDFWidget::getPageOffset() const
 
 void PDFWidget::setGridSize(int gx, int gy, bool setAsDefault)
 {
+	bool embedded = pdfdocument->embeddedMode;
 	if (gridx == gx && gridy == gy)
 		return;
 	gridx = gx;
 	gridy = gy;
-	if (gridx == 1)
+	if (gridx == 1) {
 		setPageOffset(0, true, true);
-	else if (gridx == 2 && gridy == 1)
+	}
+	else if (gridx == 2 && gridy == 1) {
 		setPageOffset(1, false, true);
-	else
+	}
+	else if (!embedded) {
 		setPageOffset(globalConfig->pageOffset, true, true);
+	}
+	else {
+		setPageOffset(pageOffsetEmbedded, true, true);
+	}
 
 	if (setAsDefault)
 		return;
@@ -2199,7 +2213,15 @@ int PDFWidget::pageStep()
  */
 int PDFWidget::gridCols(bool fromConfig) const
 {
-    int result= fromConfig ? globalConfig->gridx : gridx;
+    int result;
+    if (!fromConfig)
+        result = gridx;
+    else {
+        if (pdfdocument->embeddedMode)
+            result = gridxEmbedded;
+        else
+            result = globalConfig->gridx;
+    }
     return result;
 }
 /*!
@@ -2209,7 +2231,15 @@ int PDFWidget::gridCols(bool fromConfig) const
  */
 int PDFWidget::gridRows(bool fromConfig) const
 {
-    int result= fromConfig ? globalConfig->gridy : gridy;
+    int result;
+    if (!fromConfig)
+        result = gridy;
+    else {
+        if (pdfdocument->embeddedMode)
+            result = gridyEmbedded;
+        else
+            result = globalConfig->gridy;
+    }
     return result;
 }
 
@@ -2453,7 +2483,7 @@ void PDFWidget::fitTextWidth(bool checked)
 			if (!textRect.isValid()) return;
 			qreal targetWidth = maxPageSizeFDpiAdjusted().width() * (gridx - 1) + textRect.width() * dpi / 72.0;
 			//qreal targetWidth = textRect.width() * dpi / 72.0;
-			// total with of all pages in the grid - textMargin of a single page
+			// total width of all pages in the grid - textMargin of a single page
 			// for a 1x grid, targetWith is the same as textRect.width()
 			scaleFactor = portWidth / ((targetWidth ) + 2 * margin);
 			if (scaleFactor < kMinScaleFactor)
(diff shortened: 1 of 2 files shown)


sunderme:
I have two points
1) the button should be a symbol, not "grid". It works in English but less so in other languages.
2) I am really not convinced that it should be placed prominently.
How about adding it to context menu (next to show or in show) ?

octaeder:
context menu wuroud be ok. So we would have a sequence of seperator, Grid, Show, right?

sunderme:
> context menu would be ok. So we would have a sequence of seperator, Grid, Show, right?

correct

octaeder:
Grid to context menu only for embedded mode?

octaeder:
> Grid to context menu only for embedded mode?

I would prefere always having Grid in both modes. On large monitors, this can reduce mouse pointer movements.

octaeder:
and for continuous I suggest the same. So you don't need to think about which action where to find in different modes.

sunderme:
I don't see a benefit of having non-continuous or single step in the embedded viewer. But if it must be, maybe part of the grid sub menu
```
----
Grid
  1x1
  ...
  nxm
  ----
  continuous
  etc.
Show
  ...
```

octaeder:
well that structure is already as you describe it. Non-continous allows faster and precise scrolling with the wheel. The question is should continous also be there in the windowed case. As said, I would prefere it.

octaeder:
so windowed and embedded viewer:

grafik

sunderme:
that is fine

octaeder:
and that's why I suggested to move continuous from View to View/Grid in last PR. Would you change your mind now?

sunderme:
> and that's why I suggested to move continuous from View to View/Grid in last PR. Would you change your mind now?

The previous suggestion was about the pdf main menu , yet here we are talking about context menu , right ?
I still see no reason to change the pdf main menu.

octaeder:
I think nothing more to do from my side

src/pdfviewer/PDFDocument.h
```diff
@@ -232,6 +232,14 @@ class PDFWidget : public QLabel
 	Q_INVOKABLE void zoom(qreal scale);
 
 	virtual void wheelEvent(QWheelEvent *event);
+	int getGridxEmbedded() {return gridxEmbedded;};
```
sunderme, line 235: this concepts does not comply with the general idea of object oriented programming.
You should not have special functions to set normal parameters when the object is just in a different state (i.e. embedded)
This way the state needs to be tracked in the caller & the receiver and you can make mistake on both ends.

So the issue is that you want to overwrite the global config for the grid.
The simplest way, is to keep it as it is:
- grid 1/1 does not overwrite global config
- other grid settings set the global config and thus stay when you switch from embedded to windowed.
- This seems like a useful behavior

src/pdfviewer/PDFDocument.h
```diff
@@ -675,8 +682,10 @@ private slots:
 
     QStatusBar *statusbar;
     QToolBar *toolBar;
+    QToolBar *tbPdfView;
```
sunderme, line 685: unused, please remove

src/pdfviewer/PDFDocument.h
```diff
@@ -363,9 +370,10 @@ public slots:
 #endif
 	int		currentTool;	// the current tool selected in the toolbar
 	int		usingTool;	// the tool actually being used in an ongoing mouse drag
-	bool		singlePageStep;
+	bool	singlePageStep, singlePageStepEmbedded;
 
 	int gridx, gridy, pageOffset;
+	int gridxEmbedded, gridyEmbedded, pageOffsetEmbedded;
```
sunderme, line 376: I don't see the need for extra variables to keep a state (grid setting)

src/pdfviewer/PDFDocument.cpp
```diff
@@ -1586,9 +1589,16 @@ void PDFWidget::contextMenuEvent(QContextMenuEvent *event)
...
+	if (pdfDoc) {
+		if (pdfDoc->menuGrid || pdfDoc->menuShow) {
+			menu.addSeparator();
+			if (pdfDoc->menuGrid) {
+				menu.addMenu(pdfDoc->menuGrid);
+			}
+			if (pdfDoc->menuShow) {
+				menu.addMenu(pdfDoc->menuShow);
+			}
+		}
```
sunderme, line 1601: overly complicated code
apart from the fact the the pointers are not zeroed in the first place, there is no case when only one of the two appears.

src/pdfviewer/PDFDocument.cpp
```diff
@@ -2098,10 +2108,14 @@ void PDFWidget::setPageOffset(int offset, bool setAsDefault, bool refresh){
 		pageOffset = offset;
 	else {
 		pageOffset = gridx - 1;
-		globalConfig->pageOffset = pageOffset;
 	}
-	if (!setAsDefault)
-		globalConfig->pageOffset = pageOffset;
+	if (!setAsDefault) {
```
sunderme, line 2112: the setAsDefault was the way of distinguishing embedded/windowed
Having now an additional "embedded" makes a mess of the concept.

src/pdfviewer/PDFDocument.cpp
```diff
@@ -2199,7 +2220,15 @@ int PDFWidget::pageStep()
  */
 int PDFWidget::gridCols(bool fromConfig) const
```
sunderme, line 2221: like above
the concept from/to config and embedded/windowed needs to cleaned up

sunderme (changes requested):
There needs to be a cleaner concept about embedded/windowed handling of the grid setting.

And don't leave obsolete/dead code in pull-requests

src/pdfviewer/PDFDocument.cpp
```diff
@@ -2919,7 +2959,7 @@ PDFDocument::~PDFDocument()
 /*!
  * \brief setup ToolBar
  */
-void PDFDocument::setupToolBar(){
```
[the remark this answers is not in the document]
octaeder: removing argument `embedded` of method done

src/pdfviewer/PDFDocument.cpp
```diff
@@ -3072,6 +3112,9 @@ void PDFDocument::setupMenus(bool embedded)
     actionGroupGrid->addAction(actionCustom);
 	menuGrid->addSeparator();
     actionSinglePageStep=configManager->newManagedAction(menuroot,menuGrid, "singlePageStep", tr("Single Page Step"), pdfWidget, SLOT(setSinglePageStep(bool)), QList<QKeySequence>());
+//    if (embedded) {
```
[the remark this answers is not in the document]
octaeder: done

src/pdfviewer/PDFDocument.h
octaeder: indeed, and change to definition of setupToolBar must also be removed. This is done.

src/pdfviewer/PDFDocument.h
octaeder: This is a missunderstanding. The `Embedded` ones are used similar to the ones from the config. Why this? Well, option  continuous changes temporarily gridy (and similar single page step changes offset temporarily). So, turning of this option makes it necessary to retrieve the original values (I do so for gridx also, even so it is not needed). Until now this has been gridy=1. But now we need the user supplied value. This could be the config. Maybe I didn't get right, what you said here https://github.com/texstudio-org/texstudio/pull/3554#issuecomment-2001989472. Anyway, I still think that the user may want to have two non-shared settings, because one preferes other settings in the smaler area of the embedded viewer.  If a future decision is to save the embedded settings, too, then this could be done easier.

src/pdfviewer/PDFDocument.cpp
octaeder: no. as explained above there are cases where `pageOffset` needs to be changed temporarily, i.e. without updating the profile. That's what `setAsDefault` is used.

src/pdfviewer/PDFDocument.h
octaeder: The idea of setting grid to 1x1 without changes to the original config turns out to be problematic. When continuous mode is switched off, the grid size is restored from the config, resulting in unexpected grid changes for the user. This is independent of whether the imbedded and windowed viewer use the same of different configurations.  
Both embedded and windowed viewer initialize the config to grid 1x1 when startet first time ever. Regarding your concerns about changes to the grid, I would say that this is not such a problem. An unexperienced user might do grid changes and others with the main menu in the windowed viewer. This would have no effect on the embedded viewer (assuming two configs). Only when he discovers the context menu can he make changes in this case. I think this is acceptable. Restoring the settings of the embedded viewer will certainly please many users.  
This is now the current PR.

sunderme:
Thanks

PR #3946, opened 2025-01-13
fix styleHints not used for Fusion+Windows Style

octaeder:
if `c1 != c2` then `a==c1 && a==c2` is alway false


sunderme:
thanks

PR #3957, opened 2025-01-19
fix pages/grid after switching single page step

octaeder:
The PR fixes following two cases (alway start with non singlePageOffset). Left half of the images from master, right half fixed (s. details in txs window title).

1. Set non continuous, pageOffset>0. Scroll to top. Now setting singlePageStep moves pages and leaves empty grid faces at lower right of the grid:

    grafik
    click to enlarge

    grafik
    click to enlarge

    This issue is a direct consequence of fixing calculation of visible pages/pages list #3952. This is fixed with `reloadPage`.

3. Set continuous, pageOffset>0 and last grid row has at most pageOffset many pages (f.ex. gridx=3 and 11 pages). Scroll to end. Then setting singlePageStep moves last pages to previous line and last grid row is empty. Have a close look at the different scrollbars to the left and right. To the left you can sroll down to the empty row. But overscroll should't be possible with continuous. Also, returning back works with fixed version: Switching to non singlePageStep brings back the row on the right side:

    grafik
    click to enlarge

    grafik
    click to enlarge

    This is fixd with `updateScrollBars`. I see no difference when removing `delayedUpdate`.

### DIFF
--- a/src/pdfviewer/PDFDocument.cpp
+++ b/src/pdfviewer/PDFDocument.cpp
@@ -2265,7 +2265,8 @@ void PDFWidget::setSinglePageStep(bool step)
 		return;
 	singlePageStep = step;
 	getScrollArea()->goToPage(realPageIndex);
-    delayedUpdate();
+	reloadPage();
+	getScrollArea()->updateScrollBars();
 }
 
 void PDFWidget::goFirst()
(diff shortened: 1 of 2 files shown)


sunderme:
thanks

PR #3963, opened 2025-01-27
faster search of a visible page for the pdf viewer

octaeder:
This PR improves the algorithm of how the visible page with the highest page index is found. It should be nearly a factor of gridx faster. The algorithm uses the fact that all pages of the same row have the same vertical position, i.e. they are aligned vertically. So, instead of checking all pages of each row, the algorithm checks one page per row only. For this pages in the same column are checked (and the first page of the grid if necessary).


sunderme:
though the code may be correct, it does not improve readability (rather the opposite) and for any practical document sizes, there is no real life benefit in speed optimization here.

octaeder:
ok

PR #3978, opened 2025-02-05
use same cmd for build and preview

### LINKED ISSUE #3851: Using preview tool corresponding to compiler
### Environment
- TeXstudio 4.8.1 (git 4.8.1)
- Qt version 6.7.1, compilé avec Qt 6.7.1 R
- OS: MacOS Sonoma 14.2.1 (23C71)
- TeX distribution: TeX Live 2023

This issue is largely documented [here](https://tex.stackexchange.com/questions/718530/pdftex-log-file-eating-up-disk-space), it seems that TeXStudio does not automatically adapt its preview tool to the compiler used. This apparently caused a bug which ended up filling up the OP's hard-drive. The issue was not easily reproducible though, so it may be hard to solve.

### Expected behavior

Preview command changes with compiler, *i.e.* compiling with pdflatex imposes a pdflatex preview, lualatex a lualatex preview, *etc.*

### Actual behavior

Preview and compilation commands are independent, which may cause the aforementioned hard drive filling up process.

octaeder:
A new selection available in the Command combobox (s. Config/Preview and images below) tells the build system that the compiler used for previews should be the one used for builds. This includes the Default Compiler setup in Config/Build or given in a magic comment in the document. For this the build system tries to figure out which dvi2pngMode (i.e. one of the other selections available in the combobox) should be used internally. If this is not possible then a default selection from the combobox list is used (Preview with dvipng).

previewCmdComboboxAnimation

This PR closes #3851.

Note: Comment updated

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1801,6 +1801,16 @@ BuildManager::Dvi2PngMode BuildManager::guessDvi2PngMode() {
 	return dvi2pngModeDerived;
 }
 
+int BuildManager::index(BuildManager::Dvi2PngMode dvi2pngMode) {
+	if (dvi2pngMode==DPM_BUILD_COMPILER) return 0;
+	return static_cast <int>(dvi2pngMode) + 1;
+}
+
+BuildManager::Dvi2PngMode BuildManager::dvi2PngMode(int index) {
+	if (index==0) return DPM_BUILD_COMPILER;
+	return static_cast <BuildManager::Dvi2PngMode>(index - 1);
+}
+
 //there are 3 ways to generate a preview png:
 //1. latex is called => dvipng is called after latex finished and converts the dvi
 //2. latex is called and dvipng --follow is called at the same time, and will manage the wait time on its own
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1341,8 +1341,7 @@ void BuildManager::registerOptions(ConfigManagerInterface &cmi)
 	cmi.registerOption("Tools/Quick Mode", &deprecatedQuickmode, -1);
 	cmi.registerOption("Tools/Max Expanding Nesting Deep", &maxExpandingNestingDeep, 10);
 	Q_ASSERT(sizeof(dvi2pngMode) == sizeof(int));
-    cmi.registerOption("Tools/Dvi2Png Mode", reinterpret_cast<int *>(&dvi2pngMode), 3);
-	cmi.registerOption("Tools/AutoPreviewCmd", &autoPreviewCmd, true);
+	cmi.registerOption("Tools/Dvi2Png Mode", reinterpret_cast<int *>(&dvi2pngMode), 6);
     cmi.registerOption("Files/Save Files Before Compiling", reinterpret_cast<int *>(&saveFilesBeforeCompiling), static_cast<int>(SFBC_ONLY_NAMED));
 	cmi.registerOption("Preview/Remove Beamer Class", &previewRemoveBeamer, true);
 	cmi.registerOption("Preview/Precompile Preamble", &previewPrecompilePreamble, true);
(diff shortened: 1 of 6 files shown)


src/configdialog.ui
```diff
@@ -4046,7 +4046,23 @@ them here.</string>
...
                    </widget>
                   </item>
-                  <item row="1" column="0">
+                  <item row="1" column="1">
+                   <widget class="QCheckBox" name="checkBoxAutoPreviewCmd">
+                    <property name="text">
+                     <string notr="true">Preferre Preview with Build Compiler</string>
+                    </property>
+                    <property name="toolTip">
+                     <string>The option applys when the build compiler is pdflatex, lualatex, xelatex, or latex.</string>
```
mbertucci47, line 4055: "applies"

octaeder: thanks

src/configdialog.ui
```diff
@@ -4046,7 +4046,23 @@ them here.</string>
                     </item>
                    </widget>
                   </item>
-                  <item row="1" column="0">
+                  <item row="1" column="1">
+                   <widget class="QCheckBox" name="checkBoxAutoPreviewCmd">
+                    <property name="text">
+                     <string notr="true">Preferre Preview with Build Compiler</string>
```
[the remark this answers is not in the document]
octaeder: thanks

src/buildmanager.cpp
```diff
@@ -1771,6 +1772,32 @@ void addLaTeXInputPaths(ProcessX *p, const QStringList &paths)
 	p->setOverrideEnvironment(env);
 }
 
+/*!
+ * \brief find a Preview Mode which uses the Build Compiler (Default or by Magic Comment)
+ * it was reported (#3851) that external storage was filled up when running a preview with a different compiler than the Build Compiler
+ * from the Build setup. To prevent this check option autoPreviewCmd. This is also usefull when you switch between documents which can't
+ * be compiled with the same compiler and a manuell switch would also be necessary for previews.
```
muzimuzhi, line 1779:
```suggestion
 * from the Build setup. To prevent this check option autoPreviewCmd. This is also useful when you switch between documents which can't
 * be compiled with the same compiler and a manual switch would also be necessary for previews.
```

src/buildmanager.h
```diff
@@ -181,7 +181,12 @@ public slots:
 	QStringList deprecatedUserToolCommands, deprecatedUserToolNames;
 	QStringList userToolOrder, userToolDisplayNames;
 	enum Dvi2PngMode { DPM_DVIPNG, DPM_DVIPNG_FOLLOW, DPM_DVIPS_GHOSTSCRIPT, DPM_EMBEDDED_PDF, DPM_LUA_EMBEDDED_PDF, DPM_XE_EMBEDDED_PDF};
+	// following Dvi2PngModes add tight page modifications for to the document preamble preview
```
muzimuzhi, line 184:
```suggestion
	// following Dvi2PngModes add tight page modifications for the document preamble preview
```

src/configdialog.ui
```diff
@@ -4046,7 +4046,23 @@ them here.</string>
...
                    </widget>
                   </item>
-                  <item row="1" column="0">
+                  <item row="1" column="1">
+                   <widget class="QCheckBox" name="checkBoxAutoPreviewCmd">
+                    <property name="text">
+                     <string notr="true">Prefer Preview with Build Compiler</string>
+                    </property>
+                    <property name="toolTip">
+                     <string>The option applies when the build compiler is pdflatex, lualatex, xelatex, or latex.</string>
```
muzimuzhi, line 4055: Would "is one of ..." be better? I'm not native speaker.
```suggestion
                     <string>The option applies when the build compiler is one of pdflatex, lualatex, xelatex, or latex.</string>
```

muzimuzhi (changes requested):
Minor spelling and wording improvements.

src/buildmanager.cpp
octaeder: thanks

src/buildmanager.h
octaeder: @muzimuzhi I think I ment 'add ... to the document preamble for preview''. Ok?

muzimuzhi: Sure. It's "for to" that confused me. So you meant
```suggestion
	// following Dvi2PngModes add tight page modifications to the document preamble preview

```

octaeder: I suggest following Dvi2PngModes add tight page modifications to the document preamble used for preview

src/configdialog.ui
octaeder: I'll wait for suggestions from sunderme

sunderme:
I would suggest to simplify the design. Just add a "Autoselect preview command" (or similar) in the combobox and set it as default.
Then you can decide what to do if the logic can't find an adequate command (fail with error or try something blindly)
From user perspective, there is no advantage in having extra choices which are typically not needed.

octaeder:
This was my first try. I added it at top of the list, but this is not sensible because it shifts current user setting to the previous item. Another advantage of the checkbox would have been that we can force this option as default for anyone (not jsut for new installations). I'm going to change it to your solution.

octaeder:
grafik

grafik

I wouldn't say that there is some auto selection.
Adding a new Preview item in the future will not look very nice since it needs to be added at the end of the list.

src/configmanager.cpp
```diff
@@ -1465,11 +1465,15 @@ bool ConfigManager::execConfigDialog(QWidget *parentToDialog)
...
+    int m_dvi2pngModeIndex = 0;
+    if (buildManager->dvi2pngMode!=BuildManager::DPM_BUILD_COMPILER){
+        m_dvi2pngModeIndex = static_cast <int>(buildManager->dvi2pngMode) + 1;
+    }
+    confDlg->ui.comboBoxDvi2PngMode->setCurrentIndex(m_dvi2pngModeIndex);
 #ifdef NO_POPPLER_PREVIEW
 	int l = confDlg->ui.comboBoxDvi2PngMode->count();
 	for (int index=l-1; index>=0; index--) {
-		if (buildManager->modifyHeader.contains(buildManager->dvi2PngMode(index)))
+        if (buildManager->dvi2PngMode(index)>=BuildManager::DPM_EMBEDDED_PDF)
```
octaeder, line 1476: this will also remove the new option dpm_build_compiler

PR #3983, opened 2025-02-10
remove redundant code + changelog

octaeder:
for a proof it suffices to check the case index==0


sunderme:
grafik

? 
"preview with default compiler" should not be there in no_poppler mode (because pdflatex is a valid and likely compiler, just not for previewing)

octaeder:
sorry, thought this is function `Dvi2PngMode`

PR #3989, opened 2025-02-12
shrink embedded PDF-viewer on clicking a structure view item

octaeder:
If the embedded PDF-viewer is enlarged, you can't see the tex document. A click on the pdf shrinks the embedded viewer, so you can see the line to which txs syncted from the pdf. This PR implements a similar behavior when you click an item in the strucure or TOC view.

structureViewClickShrinksEmbeddedPDF


sunderme:
thanks

PR #3992, opened 2025-02-16
signal checkStateChanged replaces stateChanged which is depricated for Qt 6.9

octaeder:
From [here](https://wiki.qt.io/QtReleasing) I conclude that txs will be built with Qt 6.8 regardless of the OS.

[QCheckbox signals](https://doc.qt.io/qt-6/qcheckbox.html#signals)
[checkStateChanged](https://doc.qt.io/qt-6/qcheckbox.html#checkStateChanged)
[stateChanged](https://doc.qt.io/qt-6/qcheckbox.html#stateChanged)

I found the only place in the code that uses this in `texdocdialog.cpp`.

Qt 6.8.3 and 6.9 are planned to be released in March 2025.


sunderme:
we will handle deprecation when they become effective.
txs is still built for qt5.15

octaeder:
If you think we don't need it, then you may close the PR.

PR #3994, opened 2025-02-17
fix enlarge/shrink not available when windowed and embedded viewer exist

octaeder:
Explanation of the PR:

old line 6220: The line was executed for each pdf in the loop. If the first viewer opened is imbedded the menu actions are set correctly. If now a windowed viewer is opened (with option preserve-existing) then the second execution of the line will disable both options. It is not necessary to use the embeddedMode of the embedded viewer. Variable embedded reflects the state.

old lines 11374ff and 11402ff: To set the viewers stateEnlarged icons we need to now the viewer. The code relies on the fact that the embedded viewer is the first viewer opened. If you open a windowed viewer and then the embedded viewer (with option preserve-existing) then this is not true.  Even so the enlarge menu action is enabled, clicking it can not set the splitter hidden, so nothing happens. When you click on the enlarge icon, it changes state but again the enlarge method fails. Therefore we need to search the embedded viewer in the list. The foreach loop can detect if the list is empty or that there is no embedded viewer.

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -6217,10 +6217,9 @@ void Texstudio::runInternalPdfViewer(const QFileInfo &master, const QString &opt
 			viewer->setStateEnlarged(true);
             centralVSplitter->hide();
 		}
-		setEnabledMenusEnlargeShrink(viewer->embeddedMode && !configManager.viewerEnlarged, viewer->embeddedMode && configManager.viewerEnlarged);
-
 		if (preserveDuplicates) break;
 	}
+	setEnabledMenusEnlargeShrink(embedded && !configManager.viewerEnlarged, embedded && configManager.viewerEnlarged);
 #if defined Q_OS_MAC
 	if (embedded)
 		setMenuBar(configManager.menuParentsBar);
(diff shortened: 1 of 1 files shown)


src/texstudio.cpp
```diff
@@ -11371,10 +11370,15 @@ void Texstudio::enlargeEmbeddedPDFViewer()
 {
 #ifndef NO_POPPLER_PREVIEW
 	QList<PDFDocument *> oldPDFs = PDFDocument::documentList();
-	if (oldPDFs.isEmpty())
-		return;
-	PDFDocument *viewer = oldPDFs.first();
-	if (!viewer->embeddedMode)
```
sunderme, line 11377: txs does not really support multiple open pdfs. So the original code should already be fine to find the open pdf.

octaeder: txs is ok with showing multiple pdfs. You can even use muliple embedded viewers (s. image), but I wouldn't recommend it. I think that the PDF Viewer (in Build config) would need some improvements. Currently it's best when you open the viewers manually and use seperate compile commands (not Compile `&` Build). Then the viewers will update their contents as needed. If doing so, using several windowed and one embedded viewer is fine. The loop mentioned above and the definition of `documentList` already exist to handle multiple viewers.  The small code change you mentioned just ensures that you see no issues when the embedded viewer wasn't the first created. Hopefully my changes are a step further to multiple pdfs. But if you don't like these ideas, I can cleanup my code.

grafik

sunderme: I would stick with the old code.

If multi pdf is to be improved, a concept is needed first.

sunderme:
Thanks

PR #3999, opened 2025-02-20
Cwls

mbertucci47:
New cwls for `ebgaramond-maths`, `hittype`, `marginalia`, `pegmatch`, and `timechart`. Several updated cwls


sunderme:
Thanks

PR #4016, opened 2025-03-04
Cwls

mbertucci47:
New cwls for `eqnlines`, `indextra`, `math-operator`, and `outoruby`. Several updated cwls.


sunderme:
Thanks

PR #4055, opened 2025-04-25
Fix: made it so there is only one system prompt per conversation with chatbot

joaopaulo7:
LLMs work best when provided a single system prompt, usually as the first message. The previous code was adding a new system prompt message for every query in a conversation, which would probably cause issues and increase token count.

So I added code to ensure there's always a system prompt, which can be empty, at the start of the conversation. Then, every new query with a system prompt updates this one message.

### DIFF
--- a/src/aichatassistant.cpp
+++ b/src/aichatassistant.cpp
@@ -203,16 +203,12 @@ void AIChatAssistant::slotSend()
         m_timer->setInterval(100);
         connect(m_timer,&QTimer::timeout,this,&AIChatAssistant::slotUpdateResults);
     }
-    if(!config->ai_systemPrompt.isEmpty()){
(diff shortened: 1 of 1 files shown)


sunderme:
I would prefer no system prompt, when none is defined.

sunderme:
now the corner case is that you start the conversation without system prompt and change it then.

joaopaulo7:
Do you want the prompt to be defined at the start of a conversation and then be fixed for the rest of it?

sunderme:
that actually sounds reasonable. Otherwise the conversation does not really reflect question/answer anymore.

joaopaulo7:
Also, since it will be fixed per convo, there probably should be a way for the user to check it when retrieving a recorded conversation.

sunderme:
thanks

Issue #4052, opened 2025-04-24
AI chatbot stored conversations weirdness


joaopaulo7:
The stored AI conversations tree has been really awkward when I try using it. I tried selecting/deselecting the 'Record Conversation' option a couple of times, so that could have something to do with it.

There are multiple "fake empty" conversations, that have no labels, but contain text:
![Image](https://github.com/user-attachments/assets/31477d01-13e4-426b-b514-e7f05a6083c4)
 
Also, multiple conversations with labels that don't show actual conversations when selected :
![Image](https://github.com/user-attachments/assets/e075a80f-b8dd-4bbe-b282-4408a552e60e)

But, then, the past conversation "shows up" after sending a query:
![Image](https://github.com/user-attachments/assets/a33d9287-efa7-4568-989b-7794e54f9e2f)

### Environment
- TeXstudio: 4.8.6 
- Qt: 5.15.13
- OS: Ubuntu 24.04.2 LTS
- TeX distribution: texlive
- AI assistant: gemma3-4B running locally with ollama

### Expected behavior
- Stored conversations probably should have a label, with an empty one for a fresh conversation.
- Conversations probably should show their contents when selected

### Actual behavior
- A new "fake empty" conversation shows up every time the AI Chat interface opens.
- Some conversations only show contents  after a query has been sent.

### How to reproduce
In my environment it has been a consistent issue every time I try to use the interface.

PR #4057, opened 2025-04-25
Clearer ai options and small fix in recorded conversations

joaopaulo7:
### Small fix
Switching from a non-empty conversation to an empty one would not clear the `ja_messages` variable, so they would appear empty, but actually contain text when a query was made. This probably what caused the "fake empty" conversations I mentioned in #4052. Now the variable resets together with the text browser.

---

### Clearer options
Made the ai options clearer by:
- showing the system prompt that's being used in the selected conversation.
- showing that the system prompt can't be changed mid-conversation.
- showing that the temperature can't be changed for local models.


sunderme:
thanks

Issue #4117, opened 2025-07-21
Project structure complains that the input glyphtounicode file cannot be found.


kcexn:
### Environment
- TeXstudio: 4.8.8
- Qt: 6.9.1
- OS: Windows 11
- TeX distribution: Miktex

### Expected behavior
```latex
\input{glyphtounicode}
\pdfgentounicode=1
```
Should not generate any errors in the project structure as glyphtounicode is defined by pdftex.

### Actual behavior
```latex
\input{glyphtounicode}
\pdfgentounicode=1
```
generates an error in the project structure that reports file not found:

<img width="417" height="33" alt="Image" src="https://github.com/user-attachments/assets/28814f0b-25d3-41fd-b8db-59032b241fd6" />

### How to reproduce
```latex
\input{glyphtounicode}
\pdfgentounicode=1
```

sunderme:
txs assumes local tex files for `\input` file checking.

kcexn:
I think that keeping a list of exceptions of files that txs excludes from error reporting is a good enough solution for this? Is there any reason it needs to be more generic?

I'll take a look and see if I can submit a pull request.

sunderme:
I am not keen on hard coded special work arounds.
After all it is kind of cosmetic.
You could deactivate/silence the check by setting it like this:
```
\newcommand{\loadGlyphs}{\input{glyphtounicode}}
\loadGlyphs
\pdfgentounicode=1
```

kcexn:
What if we added an extra configuration item? Similar to the Bib and Image paths?

<img width="1524" height="58" alt="Image" src="https://github.com/user-attachments/assets/bc0ca38b-5b27-41d3-8123-af54d6eef995" />

This can be configured to silence other errors too.

The new command approach seems to work fairly well too.

sunderme:
the extra path definition is way too complicated for the goal to achieve.
again, you can work around the issue as described above.

kcexn:
I don't think that the extra path definition is too complicated. The `\input` command already searches the `\input@path` variable and the TEXINPUTS environment variable.

The input path approach is also useful in large projects where I want to use `\input` relative to a base directory. I.e.:

```latex
\makeatletter
\def\input@path{{/path/to/chapter/}}
\makeatother

\input{sections/introduction}
\input{examples/example1}
```

and also from inside `introduction.tex`

```latex
\makeatletter
\def\input@path{{/path/to/chapter/}}
\makeatother

\section{Introduction}
[...]

PR #4121, opened 2025-07-24
Issue #4117: Added kpsewhich search paths to getAbsoluteFilePath.

kcexn:
getSearchPath shells out to kpsewhich to populate the search paths used by getAbsoluteFilePath. On failure (let's say because kpsewhich can't be found), getSearchPath will log an error and getAbsoluteFilePath will fall back to the old behavior.

### DIFF
--- a/src/latexdocument.cpp
+++ b/src/latexdocument.cpp
@@ -3208,10 +3208,44 @@ LatexDocument *LatexDocuments::getRootDocumentForDoc(LatexDocument *doc,bool bre
     return const_cast<LatexDocument *>(current->getRootDocument(nullptr,breakAtSubfileRoot));
 }
 
+static QStringList getSearchPath(const QString &extension) {
+    static QMap<QString, QStringList> kpsePaths;
+    const auto handle_errors = [&](QProcess::ProcessError error, int lineno) {
+        static qint64 silent_until{0};
+        const auto now = QDateTime::currentSecsSinceEpoch();
+        if(now > silent_until) {
+            QString filename{QFileInfo(__FILE__).fileName()};
+            qCritical() << error << "in" << qPrintable(filename) << "at line" << lineno;
+            silent_until = now + 300;
+        }
+        kpsePaths.remove(extension);
+        return QStringList();
+    };
+
+    auto it = kpsePaths.find(extension);
+    if (it == kpsePaths.end()) {
+        it = kpsePaths.insert(extension, QStringList());
+        QProcess kpsewhich;
+        kpsewhich.start("kpsewhich", QStringList() << "-show-path="+extension, QProcess::ReadOnly);
+        if(!kpsewhich.waitForFinished())
+            return handle_errors(kpsewhich.error(), __LINE__);
+        QString path = QString(kpsewhich.readAllStandardOutput()).trimmed();
+        kpsewhich.start("kpsewhich", QStringList() << "-expand-path="+path, QProcess::ReadOnly);
+        if(!kpsewhich.waitForFinished())
+            return handle_errors(kpsewhich.error(), __LINE__);
+        path = QString(kpsewhich.readAllStandardOutput()).trimmed();
+#ifdef Q_OS_WIN
+        path.replace("/", "\\");
+#endif
+        it.value() = path.split(getPathListSeparator());
+    }
+    return it.value();
+}
+
 QString LatexDocument::getAbsoluteFilePath(const QString &relName, const QString &extension, const QStringList &additionalSearchPaths) const
 {
-	QStringList searchPaths;
-    const LatexDocument *rootDoc = getRootDocument(nullptr,true);
+	QStringList searchPaths = getSearchPath(extension);
+	const LatexDocument *rootDoc = getRootDocument(nullptr,true);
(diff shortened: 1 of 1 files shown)


sunderme:
This is not the way to go.

1) no pop-up from a utility function, especially if the failure is basically not impacting anything (qcritical)
2) no call of external processes from internal utility function especially if they are used in time critical code (syntax analysis).
 - that they are called only every 5 minutes does not make it better. That leads to unexplainable lag.
3) no "local" global variables (static kpsePath,silent_unit). Information on it gets lost, completely undocumented.

sunderme:
and by the way, here it generates 7000+ folders to search in, so no.

kcexn:
Is the problem just style? The static variables and QProcess calls can be moved elsewhere in the code, I have based the logic on KpathSeaParser.cpp. I was just hoping to avoid touching too many files to silence the glyphtounicode error.

```cpp
QString KpathSeaParser::kpsewhich(const QString &arg)
{
	ExecProgram execProgram(kpseWhichCmd + " " + arg, m_additionalPaths);
	execProgram.execAndWait();
	return execProgram.m_standardOutput;
}
```

The performance impact of this change can be completely eliminated by changing the `searchPath` logic in `LatexDocument::getAbsoluteFilePath` from appending to prepending.

sunderme:
This is not a style issue.
This introducing delays in an optimized code path.

kcexn:
After initialization, the paths are memoized in the QMap so the new code path only executes these three instructions.

```cpp
auto it = kpsePaths.find(extension);
if ( it == kpsePaths.end() ) { ... }
return it.value();
```

Do you have some specific performance regressions in mind?

sunderme:
Initialization happens in normal operation.
So that is not good.

The search path gets extended by thousands of entries. This impacts typing when entering a path because it checks its existence directly.

And for what ? That one special case does not show up red because the work-around is too much effort ?
It is not a common occurrence that tex files are loaded from texmf.

sunderme:
basically the similar issue exists for package detection.
Here the fndb (file name data base) is used to cache all available files (in code look for latexPackageList/packageListReader)

But here this seems to be excessive for such a fringe case. From my understanding lualatex and similar don't even need this.

kcexn:
Ah okay, I see that `findFileName` is being called from within `interpretCommandArguments` on every key press. I think the simplest solution for this would be to debounce the `patchStructure` slot.

```cpp
// Debounce SLOT(patchStructure(int,int)) so that it runs after the user has stopped typing.
connect(edit->editor->document(), SIGNAL(contentsChange(int,int)), edit->document, SLOT(patchStructure(int,int)));
```

This would improve typing responsiveness since TeXstudio won't run complicated logic on every keypress. It should be straightforward to write a generic debouncer that you can use anywhere in TeXstudio. It depends on whether you think updating the structure on every keypress or updating after the user is done typing is a better experience though.

If you'd like to debounce `patchStructure` I can look at adding another commit later in the week.

> But here this seems to be excessive for such a fringe case. From my understanding lualatex and similar don't even need this.

pdflatex is the default compiler for TeXstudio, so it's likely that a significant portion of users rely on it. As `glyphtounicode` is needed for proper unicode support, many users will copy and paste it into their document preamble as part of a template. I do not think this problem is a fringe issue, just that most authors do not report errors in their document structure to the TeXstudio project.

More importantly, reporting an error when there isn't one is confusing for new and less technical users and simplifying LaTeX is one of the reasons authors choose to use TeXstudio over TeXmaker or a text editor like VSCode.

sunderme:
This is where benefit and effort really spin out of control.
Basically "just change everything on the real time processing, so that my red line disappears"
(Which disappears by adding *one* line in your latex document.)

kcexn:
There is no need to change any logic. It will simply be a wrapper function that debounces the slot.

Usage will be something like:

```cpp
connect(edit->editor->document(), SIGNAL(contentsChange(int,int)), edit->document, debounce(SLOT(patchStructure(int,int))));
```

Maybe it needs to be slightly more complicated than this to make sure that errors and edge cases are all properly handled.

kcexn:
This is how I would implement the debouncer. We just wrap the call to `doc->patchStructure()` in a call to debounce. There's a little extra logic to deal with tracking changes across every keypress. This dramatically improves the keypress responsiveness as well.

sunderme:
I would like to see some convergence on technical discussions.
Unfortunately here that is not the case.
The basic principle stands that benefit/effort/risk should align, which is not the case.
So I will reject this pull request.

kcexn:
That's fair. It's not clear to me what an alternative solution might be without a significant rewrite of portions of the main loop.

The FNDB approach doesn't work here either because the number of paths is too many, and the file not found case is the common case.

I think that just leaves hardcoding the path for `glyphtounicode` or adding a configuration line item for it. Unless there is something else I haven't thought of.

sunderme:
I will not change txs for this.
Your options are:
1) use workaround
```
\newcommand{\loadGlyphs}{\input{glyphtounicode}}
\loadGlyphs
\pdfgentounicode=1
```
2) use lualatex
3) ignore the marking in TOC
4) patch txs and compile it locally.

PR #4131, opened 2025-08-04
Cwls

mbertucci47:
New cwls for `broydensolve`, `char2path`, `intexgral`, `longtwocolumn`, `optikz`, and `qworld`. Several updated cwls.


sunderme:
thanks

PR #4146, opened 2025-08-19
Title option

octaeder:
A recent question (#4013) asked for ways to change the window title. In particular one wish is that it should be possible that the title shows a 'project name'.

This PR adds an option to the command line inteface of txs which allows the user to pass a string which will be appended to the text in the window title of 1) txs main window and 2) of the  windowed pdf-viewer. As an example, the name `Project Alpha` can be seen in the following screenshot:

grafik

The option given in the current implementation (other suggestions?) would be `--titleExt "Project Alpha"`. The `--help` option shows this:

grafik

Me too uses several instances of txs. So the idea of changing the window title came in mind a long time ago. Ever since I used the simple way of changing the title string in the code because I build my own versions anyway. But this is not suitable for general users.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -353,6 +353,7 @@ bool ManagedProperty::readFromObject(const QObject *w)
 const int ConfigManager::MAX_NUM_MACROS;
 QTextCodec *ConfigManager::newFileEncoding = nullptr;
 QString ConfigManager::configDirOverride;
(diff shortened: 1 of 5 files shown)


sunderme:
If there is any useful title extension, it needs to be set automatically.
Any manual interaction, especially per command line option, will get forgotten and not used. (that would also include config settings)

A meaningful way would be root file name or root path.

PR #4151, opened 2025-08-20
show tooltip only for specific checkbox cbAIRecordConversation

octaeder:
The tooltip shown here

grafik

shows everywhere in the AI-groupbox. But the specific text fits best to the checkbox at the end of the group.


sunderme:
Thanks

PR #4152, opened 2025-08-20
set API Url for Mistral, OpenAI, and Custom

### LINKED ISSUE #4150: LLM configuration
### Environment
- TeXstudio: 
- Qt: 
- OS: 
- TeX distribution: 

### Expected behavior
modify URL  

### Actual behavior

the mistral and openia URL becomes not   editable when you switch from one to other 
(even at the first  texstudio run )

### How to reproduce

launch texstudio -> options -> configuration -> language model 
IA assistant  -> API   
switch from mistral to openAI or localmodel
you cannot edit url

octaeder:
Choosing Mistral or OpenAI should set the API Url to the appropriate value. Currently the same value as for Custom provider (localhost) ist used. The PR fixes #4150 (s. screenshots there).

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -713,6 +713,7 @@ void ConfigDialog::aiProviderChanged(int provider)
     default:
         ui.cbAIPreferredModel->clear();
         ui.cbAIPreferredModel->setPlaceholderText("Enter model name (e.g., llama-3.3-8B-Instruct)");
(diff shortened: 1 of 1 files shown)


sunderme:
The url should only show custom url.
It can be changed then to adapt.
It should be greyed out in mistral/openai mode

sunderme:
API URL is only for custom provider.
A fix has been checked in that disables the URL on dialog start-up when mistral/openai is selected

PR #4155, opened 2025-08-20
set inital model on switch between Mistral and Open AI

octaeder:
Starting txs with empty profile sets default provider Mistral and a default model (`open-mistral-7b`). Since txs doesn't know which model to choose when the user switches between those providers, txs should also set a default here. Currently, txs leaves the field blank. With this PR, the first entry in the model list of the provider is set as default.

grafik

grafik


sunderme:
This change is breaking code.
The function is a slot for an ui element.
Number of parameters must not be changed.

`connect(ui.cbAIProvider, SIGNAL(currentIndexChanged(int)), this, SLOT(aiProviderChanged(int)));`

sunderme:
thanks

PR #4156, opened 2025-08-20
fix missing AI model placeholderText

octaeder:
In AI-Group configuration the placeholderTexts (depending on the provider) do not show even when the Preferred Model field is cleared. This PR fixes this.

For an explanation of this see [qt/placeholderText](https://doc.qt.io/qt-6/qcombobox.html#placeholderText-prop) (When the QComboBox ...)


sunderme:
implemented directly to avoid conflict

thanks

PR #4164, opened 2025-08-30
cwl: correct keys names in tikz lib `shapes.symbols`

muzimuzhi:
Sync with https://github.com/pgf-tikz/pgf/commit/e0f72950b3a2351818ccfbbb030b480343f6cd2b.


sunderme:
thanks

PR #4198, opened 2025-10-03
Cwls

mbertucci47:
New cwls for `bahaistar`, `fretplot`, and `xpeekahead`. Several updated cwls.

### DIFF
--- a/completion/tcolorbox.cwl
+++ b/completion/tcolorbox.cwl
@@ -2,7 +2,7 @@
 # thatlittleboy/2018-06-07 for tcolorbox v4.13
 # muzimuzhi/30 Aug 2019 for tcolorbox v4.20
 #     reduce file size by using clist in "#keyvals:\cmdA,\cmdB"
-# Matthew Bertucci updated 2024/07/10 for v6.3.0
+# Matthew Bertucci updated 2025/09/09 for v6.8.0
 
 #include:environ
 #include:etoolbox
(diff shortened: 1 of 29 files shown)


completion/tcolorbox.cwl
```diff
@@ -2,7 +2,7 @@
 # thatlittleboy/2018-06-07 for tcolorbox v4.13
 # muzimuzhi/30 Aug 2019 for tcolorbox v4.20
 #     reduce file size by using clist in "#keyvals:\cmdA,\cmdB"
-# Matthew Bertucci updated 2024/07/10 for v6.3.0
+# Matthew Bertucci updated 2025/09/09 for v6.8.0
```
muzimuzhi, line 5: According to the `tcolorbox` [Changelog](https://github.com/T-F-S/tcolorbox/blob/master/doc/latex/tcolorbox/CHANGELOG.md), there are more options and commands added in v6.8.0:

- init option `number within from`
- commands `\tcbcounterof` and `\thetcbcounterof`

mbertucci47: @muzimuzhi You're right, I missed the commands. For the init option, the problem is that TXS does not support multiple keyval arguments. That's why there is a commented chunk at the end of tcolorbox.cwl (which I believe you may have written?)

muzimuzhi: > That's why there is a commented chunk at the end of tcolorbox.cwl (which I believe you may have written?)

Ah yes. My memory fades quickly.

Nonetheless, it doesn't hurt to add new options to that commented chunk, as you already did for `blend before title` in https://github.com/texstudio-org/texstudio/commit/25866c58d8a4c0035f8ee3c8aeda7250e6ca6b93.

mbertucci47: True. Added just now

sunderme:
thanks

PR #4200, opened 2025-10-05
Allow multiple independent keyval args for commands

sunderme:
Extend txs to handle multiple keyValArg within one command
the keyvals need to be defined for command#o1 etc
Starting from index 0 as the right most.

Example: init option in tcolorbox.cwl

### DIFF
--- a/completion/tcolorbox.cwl
+++ b/completion/tcolorbox.cwl
@@ -134,59 +134,59 @@ library/all
 \newtcolorbox{envname}{options%keyvals}#N
 \newtcolorbox{envname}[args]{options%keyvals}#N
 \newtcolorbox{envname}[args][default]{options%keyvals}#N
-\newtcolorbox[init options]{envname}{options%keyvals}#N
-\newtcolorbox[init options]{envname}[args]{options%keyvals}#*N
-\newtcolorbox[init options]{envname}[args][default]{options%keyvals}#*N
+\newtcolorbox[init options%keyvals]{envname}{options%keyvals}#N
+\newtcolorbox[init options%keyvals]{envname}[args]{options%keyvals}#*N
+\newtcolorbox[init options%keyvals]{envname}[args][default]{options%keyvals}#*N
 \renewtcolorbox{envname}{options%keyvals}
 \renewtcolorbox{envname}[args]{options%keyvals}
 \renewtcolorbox{envname}[args][default]{options%keyvals}
-\renewtcolorbox[init options]{envname}{options%keyvals}
-\renewtcolorbox[init options]{envname}[args]{options%keyvals}#*
-\renewtcolorbox[init options]{envname}[args][default]{options%keyvals}#*
+\renewtcolorbox[init options%keyvals]{envname}[args]{options%keyvals}#*
+\renewtcolorbox[init options%keyvals]{envname}{options%keyvals}
+\renewtcolorbox[init options%keyvals]{envname}[args][default]{options%keyvals}#*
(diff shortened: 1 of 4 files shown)


completion/tcolorbox.cwl
```diff
@@ -642,33 +642,22 @@ tcbox width=#auto,auto limited,forced center,forced left,forced right,minimum ce
...
-# crefname={%<singular%>}{%<plural%>}
-# Crefname={%<singular%>}{%<plural%>}
-# blend into=#figures,tables,listings
-# blend before title=#colon,dash,colon hang,dash hang#c
-# blend before title code=%<code%>
-# list inside=%<name%>
-# list type=%<type%>
-# #endkeyvals
+#keyvals:\newtcolorbox#o1,\renewtcolorbox#o1,\newtcbox#o1,\renewtcbox#o1,\newtcblisting#o1,\renewtcblisting#o1,\newtcbinputlisting#o1,\renewtcbinputlisting#o1,\newtcbtheorem#o1,\renewtcbtheorem#o1,\newtcboxfit#o1,\renewtcboxfit#o1,\DeclareTColorBox#o1,\NewTColorBox#o1,\RenewTColorBox#o1,\ProvideTColorBox#o1,\DeclareTotalTColorBox#o1,\NewTotalTColorBox#o1,\RenewTotalTColorBox#o1,\ProvideTotalTColorBox#o1,\DeclareTCBox#o1,\NewTCBox#o1,\RenewTCBox#o1,\ProvideTCBox#o1,\DeclareTotalTCBox#o1,\NewTotalTCBox#o1,\RenewTotalTCBox#o1,\ProvideTotalTCBox#o1,\DeclareTCBListing#o1,\NewTCBListing#o1,\RenewTCBListing#o1,\ProvideTCBListing#o1,\DeclareTCBInputListing#o1,\NewTCBInputListing#o1,\RenewTCBInputListing#o1,\ProvideTCBInputListing#o1,\DeclareTCBoxFit#o1,\NewTCBoxFit#o1,\RenewTCBoxFit#o1,\ProvideTCBoxFit#o1,\DeclareTotalTCBoxFit#o1,\NewTotalTCBoxFit#o1,\RenewTotalTCBoxFit#o1,\ProvideTotalTCBoxFit
```
mbertucci47, line 645: `\ProvideTotalTCBoxFit` is missing the `#o1`

muzimuzhi:
> the keyvals need to be defined for command#o1 etc
> Starting from index 0 as the right most.

This would introduce ambiguity if a command accepts two optional arguments which accept two different list of key-value pairs. Don't have a specific example at hand, just worrying.

```
\foo[bar options%keyvals]{arg}
\foo[bar options%keyvals]{arg}[baz options%keyvals]
\foo{arg}
\foo{arg}[baz options%keyvals]

## \foo#o1 refers to which %keyvals arg?
#keyvals:\foo#o1
#endkeyvals
```

Named `%keyvals` can solve the ambiguities.

```
\foo[bar options%keyvals:o1]{arg}
\foo[bar options%keyvals:o1]{arg}[baz options%keyvals:o0]
\foo{arg}
\foo{arg}[baz options%keyvals:o0]
```

sunderme:
There is no ambiguity.
Txs takes only the command definition with the most numbers of arguments into account for syntax parsing.(Rest is only for completion)
Leaving an optional argument blank does not mean it does not exist.

muzimuzhi:
> Txs takes only the command definition with the most numbers of arguments into account for syntax parsing.(Rest is only for completion)

Great! Thank you!

PR #4219, opened 2025-10-19
Round out tcolorbox init options

mbertucci47:
This completes the list of tcolorbox commands taking an `init options` keyval argument. It also adds `%keyvals` to the library cwls where it was missing. Lastly, it updates a few unrelated cwls.


sunderme:
thanks

PR #4274, opened 2025-12-13
Fix: Improved editEraseWordCmdEnv() behavior for bracket environments

miaoler:
The implementation of editEraseWordCmdEnv() for bracket environments was leaving behind unnecessary whitespace after deleting the delimiters, which caused lines to pop out. Furthermore, the caret position was not adjusted, causing it to land mid-content.

This change modifies the behavior to:
1. Automatically remove consecutive spaces following the starting delimiter and preceding the ending delimiter.
2. Adjust the carat position to the beginning or end of the now-unbracketed content, depending on which delimiter the carat was near when the command was invoked.

This provides a much cleaner and more intuitive editing experience, particularly when undoing the automatic bracket insertion.

Example:

(caret right after "\[", and press alt+del)
"\\[ abc \\]"

old behavior:
(space before 'a' and after 'c', and caret between 'a' and 'b')
" abc "

new behavior:
(no space before 'a' or after 'c', and caret right before 'a')
"abc"

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -4072,6 +4072,8 @@ void Texstudio::editEraseWordCmdEnv()
 
     if(!handled){
         // remove matching brackets and trailing spaces, move cursor to either start or end of the content within
+        // in case of within one line, remove spaces and tabs from both sides
+        // in case of separate lines, do not remove tabs on to side (indentation)
         QDocumentCursor orig, to;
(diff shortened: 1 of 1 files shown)


sunderme:
any objections to the changed code ?

miaoler:
No objections.

sunderme:
Thanks

PR #4275, opened 2025-12-14
Update packageDatabase.json

octaeder:


octaeder:
as an example take package [pkgcheck](https://www.ctan.org/ctan-ann/id/fcf12269-0713-5be0-36cf-1edfc730b75a@ctan.org) which has been updated on CTAN recently.

sunderme:
thanks

PR #4283, opened 2025-12-22
translatable items for Find Word Repetitions tool

octaeder:
This PR makes the items shown in the combobox translatable and thus more readable. This is done by hiding the technical values:

grafik


sunderme:
thanks

PR #4289, opened 2026-01-05
Cwls

mbertucci47:
New cwls for `beamer-reveal`, `linguistix-languages`, `lua-unicode-math`, `pst-kepler`, and `pst-sphericaltrochoid`. Several updated cwls.


sunderme:
thanks

PR #4294, opened 2026-01-06
cleanup Package Help UI when filtered list gets empty

octaeder:
test with filter string `biblatex-dw` by typing a second `w` at the end


sunderme:
thanks

PR #4304, opened 2026-01-17
remove \ldots (available in plaintex.cwl)

octaeder:
c.f. q&a #4303

PR #4305, opened 2026-01-17
fix always math for \lbrack and \rbrack

octaeder:
s. PR https://github.com/texstudio-org/texstudio/pull/4304#issuecomment-3764223557

### DIFF
--- a/completion/plaintex.cwl
+++ b/completion/plaintex.cwl
@@ -381,7 +381,7 @@
 \lastpenalty
 \lastskip
 \lbrace#m
-\lbrack#m
+\lbrack
 \lccode%<⟨char code⟩%>
 \lceil#m
 \ldotp#m
@@ -652,7 +652,7 @@
 \raise%<⟨dimen⟩%>
 \rangle#m
 \rbrace#m
-\rbrack#m
+\rbrack
 \rceil#m
 \Re#m
 \read%<⟨number⟩%> to %<⟨control sequence⟩%>
--- a/completion/tex.cwl
+++ b/completion/tex.cwl
@@ -566,7 +566,7 @@
 \land#m
 \langle#m
 \lbrace#m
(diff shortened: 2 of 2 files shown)


octaeder:
trying to clarify if changes are necessary for `\lBrack#m` und `\rBrack#m`  (https://github.com/texstudio-org/texstudio/pull/4304#issuecomment-3764379754)

muzimuzhi:
I (re)checked how `\(lr)brack` (`[]`, square bracket) and `\(lr)Brack` (`⟦⟧`, double/white square bracket) are defined in OpTeX.

Technically, they are all simple macros which contain the corresponding single (Unicode) character, thus can be used in both text and math modes, with appropriate font support. And they are not the only ones, for example `\(lr)paren`.

But practically, compared to `\(lr)brack` inherited from plainTeX, maybe `\(lr)Brack` (uppercase B) are mainly for uses in math.

```
$ cat optex-test.tex
\tracingoutput=1
\fontfam[LMfonts]
$\lBrack \lbrack x \rbrack \rBrack$ \par
\lBrack \lbrack x\rbrack \rBrack
\bye

$ optex optex-test.tex
$ rg '⟦' -A4 optex-test.log
61:...\_mF ⟦
62-...\_mF [
63-...\_mF 𝑥
64-...\_mF ]
65-...\_mF ⟧
--
79:...\_ten_tenrm ⟦
80-...\_ten_tenrm [
81-...\_ten_tenrm x
82-...\_ten_tenrm ]
83-...\_ten_tenrm ⟧
```

I have no strong opinion here. We can even leave `optex.cwl` as-is --- not even drop `#m` from `\(lr)brack` entries. (This leaves the enhancement possibilities to future OpTeX + TeXstudio users, if there would be any.)

sunderme:
thanks

Issue #1031, opened 2020-04-07
Sumatra command for forward and inverse search


aloispichler:
May I recommend to modify the manual as follows:
The first command in the section Sumatra (Section 4.10) should be replaced by
"(your sumatra path)" -forward-search "?c:am.tex" @ -inverse-search "\"(your TeXstudio path)\" \"%%f\" -line %%l" "?am.pdf"
because it enables both, inverse and forward search.

As well I want to recommend to include both paths, as follows:
A possible value for (your Sumatra path) is
C:/Program Files/SumatraPDF/SumatraPDF.exe
A possible value for (your TeXstudio path) is
C:/Program Files/TeXstudio/texstudio.exe

Thank you!

atiq4pk:
I have used the above command in TeXstudio 3.0.1 using SumartraPDF  3.2 64bit under WIndows 10 64bit. On my system, in inverse search it doesn't lead to the document in TeXstudio, it open some new blank document with other name.
I have noticed, in SumartraPDF, under "Set inverse search command-line" it put the following command:
`(your TeXstudio path) %f -line %l `
instead of 
`"(your TeXstudio path)" "%f" -line %l `

In the previous versions 2.x of TeXstudio,  this was working fine. I am very thankful for any suggestion, which fix the issue.

mauriciolp:
I had to change some of the double quotes to single quote to get the inverse search to work properly:

`"(your sumatra path)" -forward-search "?c:am.tex" @ -inverse-search '"(your TeXstudio path)" "%%f" -line %%l' "?am.pdf"`

Otherwise it was setting the inverse search only with "C:/Program", and trying to open the other blocks of text as pdf files.

atiq4pk:
With the latest MiKTeX 20.10, I have noticed that there is no need to give any extra command with Sumatra. Just browse to the executable file of Sumatra in External PDF command line, it will work with forward and inverse search.  (at least it is working for me).

ExpPhysU:
In general it works, 

The first time when SumatraPDF is opened there are no issues.

However with SumatraPDF open and a Build-Run there are errors

[...]

PR #4307, opened 2026-01-20
docs: fix escaping of " in sumatra command #1031

TeXhackse:
Tiny PR adjusting the escaping of `"` in the sumatra command. 
Tested with latest release of Sumatra. 

I also moved the comment, as Sumatra itself is Windows I guess it's not necessary to mention that multiple times.


sunderme:
thanks

PR #4310, opened 2026-01-23
Create texstudio128x128.webp

octaeder:
Running `make html` shows `WARNING: logo file '../../texstudio128x128.webp' does not exist`. Thus, this file cannot be copied to the destination folder `manual/build/html/_static`. When you open `getting_started.html` the logo is missing at top left. This can also be seen at [https://texstudio-org.github.io/getting_started.html](https://texstudio-org.github.io/getting_started.html).
The config file `utilities/manual/source/conf.py` states that the logo file is `html_logo = '../../texstudio128x128.webp'`. So it should be in folder `utilities`. I used `utilities/texstudio128x128.png` to create the .webp file with mogrify.

grafik


sunderme:
thanks

PR #4317, opened 2026-02-01
improve error message in update checker

octaeder:
I opened Config dialog/General to check for new updates. When I pressed button `Check Now` I faced following message:

image

I wondered what it could be. I changed the code such that the message shows more details.

image

Then I understood what's going on. This may be usefull in similar cases.

Not all messages have extended information attached:

image


sunderme:
thanks

PR #4322, opened 2026-02-05
replace \AtBeginDocument and \AtEndDocument in extra LaTeX code used in documents to be previewed

### LINKED ISSUE #4320: Preview selection (Alt+P) compile with error, no display
### Environment
- TeXstudio:  4.9.1
- Qt:  6.10.1
- OS:  OpenSuSe Tumbleweed
- TeX distribution:  texlive 2025.20250308-20120986.5 (rpm)

### Expected behavior
To display either in the Preview panel, or inline, the selection compiled and updated automatically.

### Actual behavior

All the libraries are installed on the machine (e.i. dvipng, dvips, gs, pdflatx, lualatex, xetex...). The compilation starts and ends with error as shown in the messages tab. However, there is no information on the error to identify the problem. (log file is empty).
Nothing is displayed.

This functionality used to work well, but it doesn't for some time and I wonder if it still does. Does anyone have also this issue with it?

### How to reproduce
Reproducible with two different PC and with different versions starting from 4.0.0. Also with the flatpack image, with the AppImage and with a self-build version.

octaeder:
This PR resolves #4320. The problem described arises because the document class being used and TeXstudio’s PDF-preview both attempt to use the same LaTeX mechanism to achieve certain modifications, and in doing so they interfere with each other. They use the commands `\AtBeginDocument` and `\AtEndDocument` to allow code to be specified in the document preamble that should be inserted at the beginning or end of the text written or selected for preview. The issue is resolved by having TeXstudio place the required modifications directly at the start and end of a temporary text used for preview, in exactly the same way these commands would.

Description of the changes (line numbers from changed files):
`buildmanager.cpp` builds up the `document` env. But here it's not clear if any changes should be added. This is well known in `texstudio.cpp` (l. 9432 and 9474). Thus we build necessary changes there and pass them as new parameters to `buildmanager.cpp` (s. `buildmanager.h`). The code in l. 9474 in fact was the final chapter of `Texstudio::makePreviewHeader` which is moved to `Texstudio::showPreview`.

Special behavior for Beamer class is not changed in any way.

Note: The discussion in comments below may appear somewhat outdated, as the description above has changed quite a lot (including PR's title).

<details open>
<summary>Example of LaTeX document</summary>

```Latex
\documentclass[aip,apl,amsmath,amssymb,reprint,linenumbers]{revtex4-2}
\begin{document}
Several mechanisms can lead to significant Doppler broadening of hydrogen Balmer
\end{document}
```
</details>

Assume that the whole text line has been selected for preview.

<details open>
<summary>Code finally compiled for **inline** preview</summary>

```Latex
\documentclass[aip,apl,amsmath,amssymb,reprint,linenumbers]{revtex4-2}
\usepackage[active,tightpage]{preview}
\usepackage{varwidth}
\pagestyle{empty}
\begin{document}
\begin{preview}\begin{varwidth}{\linewidth}
Several mechanisms can lead to significant Doppler broadening of hydrogen Balmer
\end{varwidth}\end{preview}
\end{document}
```
</details>

<details open>
<summary>Code finally compiled for preview in embedded PDF-viewer</summary>

```Latex
\documentclass[aip,apl,amsmath,amssymb,reprint,linenumbers]{revtex4-2}
\pagestyle{empty}
\begin{document}
Several mechanisms can lead to significant Doppler broadening of hydrogen Balmer
\end{document}
```
</details>

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1806,23 +1806,23 @@ BuildManager::Dvi2PngMode BuildManager::guessDvi2PngMode() {
 //2. latex is called and dvipng --follow is called at the same time, and will manage the wait time on its own
 //3. latex is called => dvips converts .dvi to .ps => ghostscript is called and created final png
 //Then ghostscript to convert it to
-void BuildManager::preview(const QString &preamble, const PreviewSource &source, const QString &masterFile, QTextCodec *outputCodec)
+void BuildManager::preview(const QString &preamble, const PreviewSource &source, const QString &masterFile, QTextCodec *outputCodec, const QString &atBeginDocument, const QString &atEndDocument)
 {
 	QString tempPath = QDir::tempPath() + QDir::separator() + "." + QDir::separator();
 	Dvi2PngMode dvi2pngModeDerived = guessDvi2PngMode();
 
 	//process preamble
 	QString preamble_mod = preamble;
-    static const QRegularExpression beamerClass("^(\\s*%[^\\n]*\\n)*\\s*\\\\documentclass(\\[[^\\]]*\\])?\\{beamer\\}"); //detect the usage of the beamer class
+	static const QRegularExpression beamerClass("^(\\s*%[^\\n]*\\n)*\\s*\\\\documentclass(\\[[^\\]]*\\])?\\{beamer\\}"); //detect the usage of the beamer class
 	if (previewRemoveBeamer && preamble_mod.contains(beamerClass)) {
 		//dvipng is very slow (>14s) and ghostscript is slow (1.4s) when handling beamer documents,
 		//after setting the class to article dvipng runs in 77ms
 		preamble_mod.remove(beamerClass);
 		preamble_mod.insert(0, "\\documentclass{article}\n\\usepackage{beamerarticle}");
-        // remove \mode... as well (#1125)
-        QRegularExpression beamerMode("\\\\mode.*\n");
-        //beamerMode.setMinimal(true);
-        preamble_mod.remove(beamerMode);
+		// remove \mode... as well (#1125)
+		QRegularExpression beamerMode("\\\\mode.*\n");
+		//beamerMode.setMinimal(true);
+		preamble_mod.remove(beamerMode);
 	}
 
 	QString masterDir = QFileInfo(masterFile).dir().absolutePath();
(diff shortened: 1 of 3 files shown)


sunderme:
I think, this needs to be discussed in more details first.
The current solution replaces only beamer and nothing else. It is on by default.

This can not happen with a general replacement, especially as classes tend to load relevant packages which are now disabled in preview.

octaeder:
so it would me better to have a second option for this, which is not on by default (and maybe more things)

sunderme:
Yes, i would leave the beamer part as is.
And then we can think about another option.

However, are there any other cases than revtex which fail and need this ?

octaeder:
I use only book, article or report. So I have no specific knowledge here and I can't remember any similar issue opened here in the past. Maybe @muzimuzhi has heard about it? Are there known issues or usage restrictions in the documentation of the packages added for preview?

muzimuzhi:
> there a number of lua related things that can not be dumped in a format, see the answer and the comments here [tex.stackexchange.com/questions/324559/…](https://tex.stackexchange.com/questions/324559/guidelines-for-using-mylatexformat-with-luatex)
>
> _[TeX-SX comment by Ulrike Fischer](https://tex.stackexchange.com/questions/543586/the-precompilation-with-mylatexformat-of-a-beamer-document-with-lualatex-fails#comment1373105_543586)_

So the list of unsupported document classes would be long. Also it's hard to determine if a class supports format dumping, without checking its implementation.

For example, if `beamer` is unsupported, I suppose its successor `ltx-talk` is unsupported too.

Related TeX-SX post(s)
- https://tex.stackexchange.com/q/569408
  Marcel Krüger explained in comment why OpenType fonts are only partially supported when dumping a format using LuaLaTeX

If I remember correctly, XeLaTeX also doesn't support dumping OpenType fonts.

octaeder:
@muzimuzhi Where comes dumping into place? I remember having issues when implementing preview for lualatex. The reason was that option Precompile Preamble (from Preview) was set and dumping (the result of the precompile) didn't work with tikz. But preview worked when the option was not set. Is it this you are talking about?

sunderme:
thanks

muzimuzhi:
> Where comes dumping into place?

@octaeder It seems the preview utility is implemented by dumping and using customized formats, noted the `-ini` option for pdflatex and friends.

https://github.com/texstudio-org/texstudio/blob/56512c41529e749e58222ddd3e2d54ddc394141e/src/buildmanager.cpp#L1866-L1874

PR #4336, opened 2026-02-14
disable zoom buttons when reaching slider ends

octaeder:
The zoom buttons now appear disabled when the zoom slider is positioned at either end.

grafik

This occurs simultaneously in the same way as with the corresponding menu actions.


sunderme:
thanks

PR #4338, opened 2026-02-17
action that removes entries from list of recently used files

octaeder:
When working with multiple files and using the preview in the PDF-viewer, the “Go to Source” action may open the file generated for the preview instead of the original file. In such cases, unwanted entries can be removed from the list of recently used files.

Open the `File / Open Recent` menu and select `File list`. In the displayed list, right‑click the entry to be removed and choose `Remove selected files` from the context menu. To remove multiple entries at once, select them using mouse clicks combined with `Ctrl` or `Shift` (depending on the OS), then use the context menu.

grafik

grafik

grafik

grafik

grafik

### DIFF
--- a/src/fileselector.cpp
+++ b/src/fileselector.cpp
@@ -1,8 +1,8 @@
 #include "fileselector.h"
 #include "utilsUI.h"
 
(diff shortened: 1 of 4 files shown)


sunderme:
thanks

PR #4354, opened 2026-03-02
use constants TXS_AUTO_REPLACE_QUOTE_OPEN/CLOSE

### LINKED ISSUE #4351: move TXS_AUTO_REPLACE_QUOTE_OPEN/CLOSE
I want to kindly ask for advice. I see in `configmanager.cpp` (l. 41f):
```cpp
const QString TXS_AUTO_REPLACE_QUOTE_OPEN = "TMX:Replace Quote Open";
const QString TXS_AUTO_REPLACE_QUOTE_CLOSE = "TMX:Replace Quote Close";
```
I find in `texstudio.cpp` (l. 663 and 5931):
```cpp
if (m.name == "TMX:Replace Quote Open" || m.name == "TMX:Replace Quote Close" || m.document)
```
Am I right that it seems reasonable to replace the string literals with the constants?

Since `texstudio.h` includes `configmanager.h` I would suggest to move the two lines given above to `configmanager.h` as follows:

<details open>

```cpp
#ifndef Header_Config_Manager
#define Header_Config_Manager

#include "mostQtHeaders.h"

#include "buildmanager.h"
#include "configmanagerinterface.h"
#include "smallUsefulFunctions.h"

class ConfigDialog;
class LatexCompleterConfig;
class LatexEditorViewConfig;
class WebPublishDialogConfig;
class InsertGraphicsConfig;
struct PDFDocumentConfig;
class GrammarCheckerConfig;
class InternalTerminalConfig;

const QString TXS_AUTO_REPLACE_QUOTE_OPEN = "TMX:Replace Quote Open";
const QString TXS_AUTO_REPLACE_QUOTE_CLOSE = "TMX:Replace Quote Close";

#ifdef Q_OS_WIN
const QKeySequence::SequenceFormat SHORTCUT_FORMAT = QKeySequence::PortableText;
#else
const QKeySequence::SequenceFormat SHORTCUT_FORMAT = QKeySequence::NativeText;
#endif

struct ManagedToolBar {
	QString name;
	QStringList defaults;
	QStringList actualActions;
	QToolBar *toolbar;
	ManagedToolBar(const QString &newName, const QStringList &defs);
};

class ConfigManager: public QObject, public ConfigManagerInterface
```
</details>

What do you think?

octaeder:
This PR resolves #4351 .

These constants define names of special (internal) macros reserved for quotes handling:
```
const QString TXS_AUTO_REPLACE_QUOTE_OPEN = "TMX:Replace Quote Open";
const QString TXS_AUTO_REPLACE_QUOTE_CLOSE = "TMX:Replace Quote Close";
```
They are currently used in `configmanager.cpp` to
- save *other* macros only
- identify *other* macros using the same triggers
- add *other* macros only to the Macros menu
- add them to the completer when quote handling is desired

`texstudio.cpp` uses the macro names itself to add
- *other* macros only to the tags list (side panel)
- *other* macros only to the items list of the macro editor

This PR moves the definitions from `configmanager.cpp` to `configmanager.h`. The constants can now replace the macro names used in `texstudio.cpp`.

### DIFF
--- a/src/configmanager.h
+++ b/src/configmanager.h
@@ -16,8 +16,8 @@ struct PDFDocumentConfig;
 class GrammarCheckerConfig;
 class InternalTerminalConfig;
 
-const QString TXS_AUTO_REPLACE_QUOTE_OPEN = "TMX:Replace Quote Open";
-const QString TXS_AUTO_REPLACE_QUOTE_CLOSE = "TMX:Replace Quote Close";
+inline const QString TXS_AUTO_REPLACE_QUOTE_OPEN = "TMX:Replace Quote Open";
+inline const QString TXS_AUTO_REPLACE_QUOTE_CLOSE = "TMX:Replace Quote Close";
 
 #ifdef Q_OS_WIN
 const QKeySequence::SequenceFormat SHORTCUT_FORMAT = QKeySequence::PortableText;


benibela:
Constants in headers need to marked as `inline` (and it is only supported with C++17):

https://www.fluentcpp.com/2019/07/23/how-to-define-a-global-constant-in-cpp/

octaeder:
inline seems not to be possible since most of the builds do not use c++17.

I think duplicate definitions should be no real problem here (simple string comparisons) and the string literal also needs storage. It would simplify locating usages of the macro names. But I understand that this is not a favorite

A clean solution exists but is rather cumbersome: use a header to declare the variables as externals. Create a cpp that includes the header and defines the variables and assigns the values. configmanager.cpp and texstudio.cpp simply use the include.

I think this is a dead horse. Any comments or should the PR be closed?

Thanks, Benito, for clarifying this.

sunderme:
Txs builds should be at least at c++17.

octaeder:
I looked yesterday in the actions running in my fork. Aren't these the version numbers?

https://github.com/octaeder/texstudio/actions/runs/22570308898/job/65376203353

image


https://github.com/octaeder/texstudio/actions/runs/22570308904/job/65376203407

image

sunderme:
You mean gnu 13... ?

Yes, of the compiler , not of the used c++ features.

sunderme:
thanks

PR #4361, opened 2026-03-11
user macros can be "activated" (checked)

### LINKED ISSUE #3971: Feature Request: Add a toggle button for macros
I'm not familiar with there being an option to disable a macro without deleting it for now, and it would be great to disable a macro so as not to use it while working on different documents, and so on, by means of a toggle enable/disable button added to the macros tab.

Practically everyone that wants the convenience of disabling a macro temporarily.

octaeder:
#### Final remarks about results from the comments:
Macros are not really deactivated when they are not checked. They are still visible in the Macro menu as well as in the side panel and can be executed from there. You can also execute them in the Macro Editor by pressing the Exec button. While the macro shortcut and trigger won't work (s. added tooltip), the abbreviation can still be used. No checkbox has been added to the macro browser (repo). Any macro downloaded from there will be "activated" (checked).

#### Original PR
This PR resolves #3971. I made commits (starting with class changes) for each file changed. Some commits give further details. So I hope it's easier to understand. I also added some testing.

Loading/importing macros set checked macros not having checkState element, otherwise they are checked accordingly.
When importing from macro repo you can decide what checkState they should have (s. image below).

Screenshots:

show only checked macros in side bar and menu:

grafik

grafik

Tree view in macro editor:

grafik

Macro browser (macro repo):

grafik

### DIFF
--- a/src/usermenudialog.cpp
+++ b/src/usermenudialog.cpp
@@ -140,11 +140,13 @@ void UserMenuDialog::addMacro(const Macro &m,bool insertRow)
 {
     auto *item=new QTreeWidgetItem();
     item->setFlags(Qt::ItemIsSelectable | Qt::ItemIsUserCheckable | Qt::ItemIsEnabled | Qt::ItemIsDragEnabled);
+    item->setToolTip(0,tr("Disable Shortcut and Trigger"));
     item->setText(0,m.name);
     item->setText(1,m.shortcut());
     item->setText(2,m.trigger);
     item->setText(3,m.abbrev);
     item->setData(0,Qt::UserRole,QVariant::fromValue(m));
+    item->setCheckState(0,m.checkState());
     if(insertRow){
         ui.treeWidget->addTopLevelItem(item);
         ui.treeWidget->setCurrentItem(item);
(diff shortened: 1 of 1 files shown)


sunderme:
1)The issue is about deactivating triggers (and maybe shortcuts).
There is no need to hide those macros.

2) default must be "active"
3) "import as ..." is superfluous, there is no reasonable scenario where the use imports and deactivates (and he could do so afterwards)

src/usermenudialog.cpp
```diff
@@ -330,7 +332,7 @@ void UserMenuDialog::slotOk()
 void UserMenuDialog::slotExecMacro()
 {
     QTreeWidgetItem *item = ui.treeWidget->currentItem();
-    if (item==nullptr) return;
+    if (item==nullptr || item->checkState(0)!=Qt::Checked) return;
```
sunderme: why is the "execute macro" button blocked with active/inactive macro ?

sunderme:
and at last, there should be a tooltip on the checkboxes, so that user may know what they are deactivating

octaeder:
ok, I was thinking about tooltips at the beginning. But I assumed this could be to much UI activity while cursor is moving. Do you have an idea regarding the latexcompleter (s. question above)?

octaeder:
tooltips also for folder items?

sunderme:
thanks

PR #4388, opened 2026-03-27
Update CHANGELOG.md

octaeder:
- fix empty item showing the previous header as an item (add first item when needed)
- use menu item names as is in txs
- punctuation and typo


sunderme:
thanks

PR #4391, opened 2026-03-28
Add checkbox to reset stored answer for large‑file confirmation

### LINKED ISSUE #4389: long error file
I have a very long latex file, and when I compile it (with miktex) and  the error file is longer than 2mb texstudio asks me whether I want to see the error file or not. Texstudio also gives me the option "don't show this message again". At some point I clicked on this, and now it never shows me the error file if its longer than 2mb. I have not yet figured out how to do undo this. I looked at my configuration file and the new downloaded one, but I could not find which of the many entries makes this difference. 
Perhaps one can add a feature that one can easily return to seeing the full error message. 

### Environment
- TeXstudio: 
- Qt: 
- OS: 
- TeX distribution:

octaeder:
This PR resolves #4389 

When you load a large log file, an additional size confirmation dialog is shown. This dialog includes a checkbox that allows txs to remember whether you selected Yes or No. If No is selected and remembered, you will no longer be able to load large log files. In such cases, users previously had to open an issue and manually edit a value in the .ini file.

To avoid this situation, the first confirmation dialog (asking whether you want to load the log as a LaTeX log file) has been enhanced. If the size confirmation is no longer shown, an option is now provided that allows the user to restore the original behavior:

grafik

### DIFF
--- a/src/texstudio.cpp
+++ b/src/texstudio.cpp
@@ -2242,11 +2242,24 @@ LatexEditorView *Texstudio::load(const QString &f , bool asProject, bool recheck
     }
 #endif
 
-    if (f_real.endsWith(".log", Qt::CaseInsensitive) &&
-            UtilsUi::txsConfirm(QString("Do you want to load file %1 as LaTeX log file?").arg(QFileInfo(f).completeBaseName()))) {
-        outputView->getLogWidget()->loadLogFile(f, documents.getTemporaryCompileFileName(), QTextCodec::codecForName(configManager.logFileEncoding.toLatin1()));
-        setLogMarksVisible(true);
-        return nullptr;
+    if (f_real.endsWith(".log", Qt::CaseInsensitive)) {
+        UtilsUi::txsWarningState rememberChoice=static_cast<UtilsUi::txsWarningState>(configManager.getOption("LogView/RememberChoiceLargeFile",0).toInt());
+        bool result;
+        QString question=tr("Do you want to load file %1 as LaTeX log file?").arg(QFileInfo(f).completeBaseName());
+        if (rememberChoice==UtilsUi::DontRemember) {
+            result=UtilsUi::txsConfirm(question);
+        } else {
+            rememberChoice=UtilsUi::DontRemember;
+            result=UtilsUi::txsConfirmWarning(question,rememberChoice,tr("Clear remembered size confirmation"),QMessageBox::Question);
+            if (!rememberChoice==UtilsUi::DontRemember) {
+                configManager.setOption("LogView/RememberChoiceLargeFile",static_cast<int>(UtilsUi::DontRemember));
+            }
+        }
+        if (result) {
+            outputView->getLogWidget()->loadLogFile(f, documents.getTemporaryCompileFileName(), QTextCodec::codecForName(configManager.logFileEncoding.toLatin1()));
+            setLogMarksVisible(true);
+            return nullptr;
+        }
     }
 
     raise();
--- a/src/utilsUI.cpp
+++ b/src/utilsUI.cpp
@@ -37,15 +37,15 @@ bool txsConfirmWarning(const QString &message)
  * \param rememberChoice if true, return true and avoid Msg. rememberChoice is overwritten by checkbox result.
  * \return yes=true
  */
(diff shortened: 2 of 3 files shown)


sunderme:
If I see it correctly, a dialog will always appear.
Either asking if the large log file is to be loaded or if your last choice to remember was correct ?

octaeder:
It works as it was intended (before the change). I can clarify this: First, a __question__ dialog always appears. The reason is that when opening a .log file, txs needs to know whether it should open it in the editor or in the log view. This is the dialog where I want to add the “clear” checkbox.

The clear checkbox is only added when I detect that the __warning__ dialog — which appears only for large files (above the 2 MB limit) — had previously been answered with Yes or No and the option to remember the answer was checked. If the clear checkbox is selected, this dialog can reappear when necessary.

I believe this is a smart solution that avoids modifying the configuration dialog, which I assume is something you would prefer not to change.

octaeder:
I think that the text of the question dialog, i.e.

grafik

is not precise enough. The question suggests that choosing No will prevent the file from being loaded at all. In reality, however, txs still loads it in the editor.

Maybe the text could be: _Open %file in the LaTeX log view? If not, it will be opened in the editor._

sunderme:
This is not the way to go.
The point on "remember choice" is that you don't have to see a blocking pop-up in that scenario again.
Replacing it by another blocking pop-up which basically just asks "are you sure" would be even more annoying.

octaeder:
I can't see your point. The only thing changed in the flow is adding the checkbox (s.  mark 1). Blue boxes are dialogs with user interaction. Flow goes downwards if condition is true, otherwise left or right:

grafik
 
If you still have concerns, I assume they might be about the current behavior rather than the proposed change. Let me know if I misunderstood.

octaeder:
and the new checkbox will only be added when you checked the Remember checkbox once before. If you did not do so, everything is exactly the same as before.

sunderme:
again, from a user guidance perspective, that is not the way to go.
What exactly do you want to achieve ?
My assumption is:
* user falsely set "remember load log >2 MB" to "don't load"
* now he wants to revert that decision

He is probably working around viewLog (menu or toolbar). It is not obvious that you can load a log from open file and then get an extra option to clear that flag.

The only sensible way seems to be to 
1) notify in messages what has happened (i.e. why the log is not displayed)
2) offer a way to remedy directly there.

octaeder:
Whether the user made the decision by mistake or simply wants to revert it later does not really matter.
In any case, when they continue opening large files, we have no way of knowing whether their remembered decision is still what they want.

So the question is: at which point, or in which way, do you want to offer the user a way to revert their decision without risking that they feel interrupted in their workflow? I don't see it.

octaeder:
Maybe the question in the first dialog is still not sufficient in the situation where a large file is opened but will  not be loaded because the remembered answer is “No”, even though the user now selects “Yes”.

Solution: Give the necessary information in the box and maybe deactivate "Yes" until the checkbox is checked. This avoids that the user needs to open the file again after clearing the remembered answer, that was "No".

Possible text: You cannot choose “Yes” because you previously asked TeXstudio to remember your decision not to load large log files into the Log Viewer.

octaeder:
It seems that you are going your own way without further discussion (configuration.md was changed)

sunderme:
that I did already before your PR and it is independent on any GUI solution.

octaeder:
Okay, then I’ll wait for further information

sunderme:
Prelude: 
The log size limitation is old (pre 2017), probably too small for today's computer, so we can increase it to 32 MB ?
I don't see real harm, when a user selects always to load large logs. But there is harm otherwise (do not load large logs) as no logs are shown anymore (for that particular document)

I have emulated the behavior (log size to large)
So, 2 fixes are necessary:
1) the viewer acts as if a log is opened,i.e. it switches the view in the message pane. It should just show, log not loaded because and then give relevant information
2) on that message, either instructions what to do/where to look or a button can be shown

It is not helpful to implement something on a completely different use path (load per fileOpen). Extremely unlikely that someone will find that by accident and if it needs to be described, it is not really different from editing the ini file.
And "feel interrupted in their workflow" is no argument here, as it is rare.

sunderme:
implemented my solution

PR #4392, opened 2026-03-28
Update nicematrix.cwl

fpantigny:
The keys `color-inside` and  `colortbl-like` are now deprecated in `nicematrix`.


sunderme:
I can't merge this.
txs needs to work with old tex installations (e.g. linux) and user still may use those commands.

fpantigny:
Right. I understand.

PR #4394, opened 2026-03-28
check Qt${QT_VERSION_MAJOR} at first

PPN-SD:
to respect an exported value of QT_VERSION_MAJOR and avoid conflict if both Qt5/Qt6 are installed

Initially reported in Gentoo: https://bugs.gentoo.org/940747


sunderme:
thanks

PR #4416, opened 2026-04-08
offer "Clear stored answer" action when loading large logs

octaeder:
Based on the recently introduced info messages in the log panel, this PR aims to provide more convenience to the user. For reference s. #4389.

In cases where the file‑size confirmation dialog was automatically answered with Yes or No based on the last remembered choice, the info message now includes a link. This link allows the user to quickly and easily clear the stored answer with a single click. This gives them back full control in all possible situations. They no longer need to look up in the manual how to modify the ini file accordingly. The section about hidden options is still not obsolete, though.

### DIFF
--- a/src/latexlogwidget.cpp
+++ b/src/latexlogwidget.cpp
@@ -167,11 +167,11 @@ bool LatexLogWidget::loadLogFile(const QString &logname, const QString &compiled
                 config->setOption("LogView/RememberChoiceLargeFile",static_cast<int>(rememberChoice));
             }
             if(!result){
-                if(skipLoadRememberChoice){
-                    setInfo(tr("Log not loaded because of size constraint (%1 MB). User chose not to load it and set it as default option !\nTo revoke that choice, see [manual](%2)").arg(fileSizeMB, 0, 'f', 2).arg("https://texstudio-org.github.io/configuration.html#hidden-settings"));
-                }else{
+                if(rememberChoice==UtilsUi::DontRemember){
                     setInfo(tr("Log not loaded because of size constraint (%1 MB). User chose not to load it !").arg(fileSizeMB, 0, 'f', 2));
                     m_lastIgnoredFilename=logname;
+                }else{
+                    setInfo(tr("Log not loaded because of size constraint (%1 MB). User chose not to load it and set it as default option ! [Clear stored answer](%2)").arg(fileSizeMB, 0, 'f', 2).arg("setToUtilsUi::txsWarningState::RememberFalse"));
                 }
                 return false;
             }
--- a/src/latexlogwidget.cpp
+++ b/src/latexlogwidget.cpp
@@ -129,12 +129,12 @@ void LatexLogWidget::onInfoLinkActivated(const QString &link)
     config->setOption("LogView/RememberChoiceLargeFile",static_cast<int>(rememberChoice));
 
     if (link==QLatin1String("setToUtilsUi::txsWarningState::RememberFalse")) {
-        setInfo(tr("Your remembered answer has been cleared. Please reopen the log to load it."));
+        setInfo(QString());
     }else if (link==QLatin1String("setToUtilsUi::txsWarningState::RememberTrue")) {
-        setInfo(tr("Your remembered answer has been cleared."));
+        setInfo(QString());
     }else{
         m_lastIgnoredFilename.clear();
-        setInfo(tr("Reset done."));
+        setInfo(QString());
     }
 }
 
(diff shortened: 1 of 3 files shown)


sunderme:
so, how do we continue here ?

octaeder:
I made the changes but need some more testing. So this is still ongoing.

sunderme:
ok

octaeder:
Please have a look. The user should now be able to revert restrictions with a mouse click. Checking for same name is back again.

sunderme:
it seems to work. But you have managed to squeeze 7 translations for such a niche functionality in it...

octaeder:
You answered "yes" to my question "So, you’d like me to keep most of your string contents, meaning the majority of the translation can remain unchanged?". But it seems that the question wasn't clear. The intended meaning was that strings could change such that the translation string will only change a bit (maybe at the end). I see:

l. 167 `"The logfile is very large (%1 MB) are you sure you want to load it?"`: unchanged
l. 172 `"Log not loaded because of size constraint (%1 MB). User chose not to load it ! [reset](%2)"`: added string at the end
l. 175 `"Log not loaded because of size constraint (%1 MB). User chose not to load it and set it as default option ! [Clear stored answer](%2)"`: changed string at the end
l. 179 `"Log file size (%1 MB) above limit; loading performed due to your remembered choice. [Clear stored answer](%2)"`: new string

l. 172: Is it ok to use the + operator to add the new string?
l. 175: "manual" is part of the translation. How can this stay unchanged?
l. 179 Is a new string ok?

By the way, you refer to the online manual, but you once mentioned that you prefer referencing the local one (so no internet is required). I had already prepared a change for this, but afterward I thought that the solution discussed here would be the better option.

sunderme:
we can merge it as is, but probably you need to get a feeling what translation actually involves.
I am currently doing de,es,it and nl (with the help of google translate), so this time, maybe you want to do it ?

And to be clear, that work has to be done by another number of people for their languages.

octaeder:
I made many translations on transifex in the past. I assume this should suffice to do it for me with the aid of AI. Yes there are many languages. Because of this I would also like to know if my suggestions would improve things. Why do you see 7 changes?

sunderme:
L 132,134,137 ? plus the ones you mentioned.

octaeder:
if I counted well then two translations remain (one changed, one new).

src/latexlogwidget.cpp
```diff
@@ -142,24 +159,27 @@ bool LatexLogWidget::loadLogFile(const QString &logname, const QString &compiled
...
+                    QString text=tr("Log not loaded because of size constraint (%1 MB). User chose not to load it !").arg(fileSizeMB, 0, 'f', 2);
+                    setInfo(("["+text+"](%1)").arg("resetName"));
                     m_lastIgnoredFilename=logname;
+                }else{
+                    QString text=tr("Log not loaded because of size constraint (%1 MB). User chose not to load it and set it as default option !").arg(fileSizeMB, 0, 'f', 2);
+                    setInfo(("["+text+"](%1)").arg("setToUtilsUi::txsWarningState::RememberFalse"));
                 }
                 return false;
+            }else if(rememberChoice==UtilsUi::RememberTrue) {
+                QString text=tr("Log file size (%1 MB) above limit; loading performed due to your remembered choice.").arg(fileSizeMB, 0, 'f', 2);
```
sunderme: this is a new string, hence you can also make a sensible link

src/latexlogwidget.cpp
```diff
@@ -142,24 +159,27 @@ bool LatexLogWidget::loadLogFile(const QString &logname, const QString &compiled
...
-                    setInfo(tr("Log not loaded because of size constraint (%1 MB). User chose not to load it and set it as default option !\nTo revoke that choice, see [manual](%2)").arg(fileSizeMB, 0, 'f', 2).arg("https://texstudio-org.github.io/configuration.html#hidden-settings"));
-                }else{
-                    setInfo(tr("Log not loaded because of size constraint (%1 MB). User chose not to load it !").arg(fileSizeMB, 0, 'f', 2));
+            if(!m_answer){
+                if(rememberChoice==UtilsUi::DontRemember){
+                    QString text=tr("Log not loaded because of size constraint (%1 MB). User chose not to load it !").arg(fileSizeMB, 0, 'f', 2);
+                    setInfo(("["+text+"](%1)").arg("resetName"));
                     m_lastIgnoredFilename=logname;
+                }else{
+                    QString text=tr("Log not loaded because of size constraint (%1 MB). User chose not to load it and set it as default option !").arg(fileSizeMB, 0, 'f', 2);
```
sunderme: this is a new/changed string, hence you can also make a sensible link

sunderme:
okay, finished ?

octaeder:
please merge

PR #4419, opened 2026-04-11
Cleanup dialog should not constantly add a dot to given extensions

### LINKED ISSUE #4418: Clean Auxiliary Files needs additional flexibility
Please add the ability to specify file extension not separated with a "." to the list of files deleted by "Tools > Clean Auxiliary Files..." .

Currently, users can add additional file extensions to the list of auxiliary files to be removed by "Tools > Clean Auxiliary Files..." but only those separated by a period from the original FILENAME.tex file. This was previously  brought up as too strict of a constraint (see Issue 39, for example).

Some important auxiliary files do not follow the convention of FILENAME.ext. Example: (the motivation for this suggestion) the FILENAME-luamml-mathml.html generated by the new LaTeX Tagging Project modifications to lualatex. When not cleared along with the other auxiliary files, the stale MathML file can cause compile errors. It should be included among the list of auxiliary files deleted by "Tools > Clean Auxiliary Files...".

Many academics are, or will soon, be shifting from PDFLaTeX to LuaLaTeX, due to the need to generate accessible PDFs (this is becoming increasingly important at U.S. educational institutions so that course materials and graduate theses comply with ADA and other accessibility standards). This was also stated as a problem affecting "*-blx.bib" files generated by biblatex in Issue 39, but I'm not sure if this is still an issue or not.

### Any Other info
A few suggestions for implementation: either (1) adding capacity for a special character to signal that no "." separates the file name from the extension provided [e.g., suppose \ is used so that "\-luamml-mathml.html" is interpreted as described above], or perhaps the easier fix is (2) just include the "." as part of the file extensions in the list of default file extensions so that they read as ".log,.aux,.dvi,.lof,.lot..." instead of the current default "log,aux,dvi,lof,lot...".  Another solution might be to have the LaTeX Tagging Project developers alter the naming of this MathML file, if someone wants to submit that request, however more examples than the MathML file exist and justify adding this functionality to TeXStudio.

Huge thank you to the developers for their hard work maintaining this software!! :)

octaeder:
This PR resolves #4418 
a new option `ExtensionsWithDot` is introduced which replaces option `Extensions`

### DIFF
--- a/src/cleandialog.cpp
+++ b/src/cleandialog.cpp
@@ -18,7 +18,7 @@ CleanDialog::CleanDialog(QWidget *parent) :
 	UtilsUi::resizeInFontHeight(this, 38, 22);
 
 	ConfigManagerInterface *config = ConfigManager::getInstance();
-	config->registerOption("CleanDialog/Extensions", &currentExtensions, defaultExtensions);
+	config->registerOption("CleanDialog/ExtensionsWithDot", &currentExtensions, defaultExtensions);
 	config->registerOption("CleanDialog/Scope", &scopeID, 0);
 	if (scopeID < 0 || scopeID >= MAX_SCOPE) scopeID = 0;
 
--- a/src/cleandialog.cpp
+++ b/src/cleandialog.cpp
@@ -28,7 +28,7 @@ CleanDialog::CleanDialog(QWidget *parent) :
 	if (rxValExtensionList->validate(currentExtensions, dummyPos) == QValidator::Acceptable) {
 		ui->leExtensions->setText(currentExtensions);
 	} else {
-		UtilsUi::txsWarning("Invalid extension list found. Resetting to default.");
+		UtilsUi::txsWarning(tr("Invalid extension list %1 found. Resetting to default.").arg(currentExtensions));
 		currentExtensions = defaultExtensions;
 		ui->leExtensions->setText(currentExtensions);
 	}
--- a/src/cleandialog.cpp
+++ b/src/cleandialog.cpp
@@ -22,8 +22,8 @@ CleanDialog::CleanDialog(QWidget *parent) :
 	config->registerOption("CleanDialog/Scope", &scopeID, 0);
 	if (scopeID < 0 || scopeID >= MAX_SCOPE) scopeID = 0;
 
-	QString allowedChars = "[^\\\\/\\?\\%\\*:|\"<>\\s,;\\{\\}\\[\\]\\(\\)]";
-    QRegularExpressionValidator *rxValExtensionList = new QRegularExpressionValidator(QRegularExpression(QString("(%1+\\.)*%1+(,(%1+\\.)*%1+)*").arg(allowedChars)), this);
+	QString disallowedChars = "[^\\\\/\\?\\%\\*:|\"<>\\s,;\\{\\}\\[\\]\\(\\)]";
+    QRegularExpressionValidator *rxValExtensionList = new QRegularExpressionValidator(QRegularExpression(QString("(%1+\\.)*%1+(,(%1+\\.)*%1+)*").arg(disallowedChars)), this);
 	int dummyPos;
 	if (rxValExtensionList->validate(currentExtensions, dummyPos) == QValidator::Acceptable) {
 		ui->leExtensions->setText(currentExtensions);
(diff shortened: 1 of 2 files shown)


sunderme:
and how are old, user defined, extensions translated into new ones ?

octaeder:
Ah, I think then I need to know whether the old option is still there or the new option is already set up. How can I figure out this?

sunderme:
you can check for config keys

octaeder:
ok, I see `existsOption()`

sunderme:
hmm, rather like `if (config->contains("data/styleHint/bold"))` ...

octaeder:
Notes:
1. `config->contains` doesn't work for me [^1]
2. ~the regExp is only checked when reading the option but not after editing the extensions list~
3. the regExp doesn't include `^` and `$`. So you can add points there as many as you like
4. `cleandialog.cpp` uses `#include "ui_cleandialog.h"` but I cannot find this file.

[^1]: 'class ConfigManagerInterface' has no member named 'contains'

sunderme:
1) the translation needs to be done in void `Texstudio::readSettings(bool reread)` 
3) is that any different from current behaviour ?
4) is auto-generated

octaeder:
You may have only read the notification email, but I edited item 2 and it is crossed out now. I assuming that you use the same numbering as I started with:
ad 1) I agree
ad 2) I think I was mistaken.
ad 3) no change here. adding dots at the beginning is possible because the regExp starts with `(%1+\\.)*%1` and the part in parentheses can have no match and `%1` (disallowedChars) doesn't contain the dot.

src/texstudio.cpp
```diff
@@ -4633,6 +4631,18 @@ void Texstudio::readSettings(bool reread)
...
+    // convert to file extensions with dot for cleanup-dialog (#4419)
+    if (!config->contains("CleanDialog/ExtensionsWithDot") && config->contains("CleanDialog/Extensions")) {
+        QString oldExtensions = config->value("CleanDialog/Extensions").toString();
+        QStringList oldExtList = oldExtensions.split(',', Qt::SkipEmptyParts);
+        QStringList convertedList;
+        for (const QString &extension : oldExtList) {
+            convertedList << "." + extension;
+        }
+        QString convertedExtensions = convertedList.join(",");
+        config->setValue("CleanDialog/ExtensionsWithDot", convertedExtensions);
```
sunderme, line 4643: the old entry should be removed

src/texstudio.cpp
```diff
@@ -4633,6 +4631,18 @@ void Texstudio::readSettings(bool reread)
     tobemaximized = config->value("MainWindow/Maximized", false).toBool();
     tobefullscreen = config->value("MainWindow/FullScreen", false).toBool();
 
+    // convert to file extensions with dot for cleanup-dialog (#4419)
+    if (!config->contains("CleanDialog/ExtensionsWithDot") && config->contains("CleanDialog/Extensions")) {
```
sunderme: check only for old entry and remove when done

src/cleandialog.cpp
```diff
@@ -18,17 +18,17 @@ CleanDialog::CleanDialog(QWidget *parent) :
...
-	QString allowedChars = "[^\\\\/\\?\\%\\*:|\"<>\\s,;]";
-    QRegularExpressionValidator *rxValExtensionList = new QRegularExpressionValidator(QRegularExpression(QString("(%1+\\.)*%1+(,(%1+\\.)*%1+)*").arg(allowedChars)), this);
+	QString disallowedChars = "[^\\\\/\\?\\%\\*:|\"<>\\s,;\\{\\}\\[\\]\\(\\)]";
+	QRegularExpressionValidator *rxValExtensionList = new QRegularExpressionValidator(QRegularExpression(QString("(%1+\\.)*%1+(,(%1+\\.)*%1+)*").arg(disallowedChars)), this);
 	int dummyPos;
 	if (rxValExtensionList->validate(currentExtensions, dummyPos) == QValidator::Acceptable) {
 		ui->leExtensions->setText(currentExtensions);
 	} else {
-		UtilsUi::txsWarning("Invalid extension list found. Resetting to default.");
+		UtilsUi::txsWarning(tr("Invalid extension list %1 found. Resetting to default.").arg(currentExtensions));
```
sunderme: there is not much information except for programmer. 
I don't see the necessity to make new translations.

sunderme:
okay. All done ?

octaeder:
yes from my side

sunderme:
thanks

PR #4438, opened 2026-04-25
avoid crash when last editor is moved from 2. to 1. split

### LINKED ISSUE #4436: txs crahses when moving editor back from split screen
### Environment
- TeXstudio:  current dev
- Qt:  6.11.0
- OS:  Win11
- TeX distribution:  miktex

### Expected behavior
the second editor in the right split should be moved back to the left split without crash

### Actual behavior

On doing so txs crashes with a backtrace

### How to reproduce
start txs with option no-session. Use Quick wizard to create two editors with minimal code. Use context menu of second editor tab and move it  to other view. After that move it in a similar way back to the left and txs will crash.

I assume that some resent optimizations to the tabs is the reason.

octaeder:
This PR fixes #4436. The issue describes the reason for the crash. Now the code checks whether the tabGroup is empty.

### DIFF
--- a/src/editors.cpp
+++ b/src/editors.cpp
@@ -262,7 +262,9 @@ QList<LatexEditorView *> Editors::topEditors()
 {
     QList<LatexEditorView *> editors;
     foreach (TxsTabWidget *tabGroup, tabGroups) {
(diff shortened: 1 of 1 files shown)


sunderme:
Thanks

PR #4446, opened 2026-04-26
clarify indentation of an if statement

octaeder:
One of the builds shows this message:
```
D:/a/texstudio/texstudio/src/buildmanager.cpp:1851:17: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
 1851 |                 if(!tf->open()) return; // opening file failed
      |                 ^~
D:/a/texstudio/texstudio/src/buildmanager.cpp:1852:33: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
 1852 |                                 QTextStream out(tf);
```
Here, indentation is done with blanks while tabs are used in the following line. Depending on the size of tabs this might lead to an effective indentation that is larger than the one of the `if`. This could lead to the assumption that the line (1852) is part of the `then` path. The image shows the fixed part:

image

### DIFF
--- a/src/buildmanager.cpp
+++ b/src/buildmanager.cpp
@@ -1848,7 +1848,7 @@ void BuildManager::preview(const QString &preamble, const PreviewSource &source,
 				//write preamble
 				QTemporaryFile *tf = new QTemporaryFile(tempPath + "hXXXXXX.tex");
 				REQUIRE(tf);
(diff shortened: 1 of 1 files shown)


sunderme:
I would prefer not to see more of these pointless changes.
The static checkers have become more a nuisance than a real help.

PR #4466, opened 2026-05-11
Fixes for teamtype collaboration

cargaji:
This PR contains three changes to the teamtype integration in TeXstudio, which now can:

- set the user-configured username from TeXstudio's settings,
- automatically create the client folder when connecting to another user, and
- handle sharing and joining from folders with uri-escapable chars.

The last one is the one I consider the most important, as it was blocking me from sharing from a folder with a space (e.g. `/tmp/my test`), which is encoded as `file:///tmp/my%20test/`, and the decoding process produced `/tmp/my%20test`, which does not exist in the filesystem.


sunderme:
Thanks

PR #4499, opened 2026-06-08
Fix heap-after-free due to document rescan

ppizarror:
Hi! First, thanks a lot for this fantastic project. I've been using it for several years.

I often struggle with crashes due to external updates to .tex files. While debugging, I got a use-after-free in LatexDocument::getListOfDocs(). Its result is cached in the project-shared lp->projectDocuments, but the cache was only invalidated in the two setMasterDocument paths. Every other mutation of the document set / master-child graph (deleteDocument, removeDocs, addDocument, addChild/removeChild) left stale.

So, for fixing it, I invalidated lp->projectDocuments on every graph mutation.
To confirm, added a unit test that confirms the exact crash pattern. Greetings!

### DIFF
--- a/src/latexdocument.cpp
+++ b/src/latexdocument.cpp
@@ -63,6 +63,10 @@ LatexDocument::~LatexDocument()
 	mLineSnapshot.clear();
 
     qDeleteAll(docStructure);
+
+    // make sure this (about to be freed) document can never linger in the
+    // shared project cache as a dangling pointer (see getListOfDocs)
+    if (lp) lp->projectDocuments.clear();
 }
 
 void LatexDocument::setFileName(const QString &fileName)
@@ -1943,11 +1947,13 @@ void LatexDocument::setMasterDocument(LatexDocument *doc, bool recheck)
 void LatexDocument::addChild(LatexDocument *doc)
 {
 	childDocs.insert(doc);
+	lp->projectDocuments.clear(); // graph changed, invalidate cache
 }
 
 void LatexDocument::removeChild(LatexDocument *doc)
 {
 	childDocs.remove(doc);
+	lp->projectDocuments.clear(); // graph changed, invalidate cache
 }
 
 bool LatexDocument::containsChild(LatexDocument *doc) const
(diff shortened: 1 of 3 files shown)


sunderme:
Thanks. I will review later this week.

sunderme:
looks like a `lp->projectDocuments.clear();` was added all over the place.
What action exactly leads to a crash ?

ppizarror:
Hi @sunderme. Often, when TexStudio is opened, and .tex files are rewritten or moved from VS Code (imagine I am using some agent). Or if I have some auto-generated .tex from a Python script, I run that script, regenerate .tex, and then TexStudio crashes.

I am running the latest version on macOS 26.5.1.
To spot the culprit, I just compiled and ran the debug version on my machine. You can see that, without the fix, the unit test I added just collapses to the exact crash I detected.
download

sunderme:
the file is open in txs while changed externally ?

ppizarror:
> the file is open in txs while changed externally ?

exactly. that is a much better description 🤔

sunderme:
Well, I don't get a crash, but let's see.

sunderme:
I think, I would like to see a (simple) test case to recreate the issue.
Currently I do not see a crash (nor failure to invalidate the cache)

or you can check how it gets to "delete document" in your use case.

ppizarror:
Sure I will provide a simple test case by the end of the wee. Thanks!

sunderme:
I still would like to see a real test case.

sunderme:
can you check if txs still crashes on your end with latest changes ?

sunderme:
I have fixed potential crashes.
The PR can not be integrated as it applies cache invalidation in a very unspecific way.
Since no test case is provided, I can't check if my fixes really solve the issue.
But since there is no feedback, I close this.
Please check the latest code, if the crash persists.

ppizarror:
Thanks! I tried to come up with a working testcase, yet I was blown away by work... 
I really appreciate all your efforts for texstudio. Thanks for bringing up this fantastic project.

PR #4537, opened 2026-07-04
Advanced.md

octaeder:
* Some text changes suggested by Copilot.
* Swap two images into the right place.

Hint: Maybe the images mentioned need to be cropped, as my build of the manual shows the following (including changes):

| snapshot |
|-----------|
| grafik |


sunderme:
Thanks

PR #4539, opened 2026-07-04
align images properly

octaeder:
Changes:

|screenshot|
|-|
|grafik|

Question: in `GIT/SVN Support` part `check-in` is used (not checkin) which is also suggested by Copilot (and I assume this is the same with `checkout).` Should I change them several times?


sunderme:
Both variants are understandable. I don't care.

octaeder:
ok, then I won't change and you can merge

PR #4559, opened 2026-07-18
Update packageDatabase.json

octaeder:
latest update from CTAN (7003 packages at 2026-07-18-13.51.14 MESZ)


sunderme:
I ran the script earlier today. So the database is already current.

PR #4583, opened 2026-08-05
fix: only add subdir adwaita-qt if building with adwaita

TomJampen:
### Problem
The subdirectory `srv/adwaita-qt` is added to the build in `CMakeLists.txt` even if `TEXSTUDIO_BUILD_ADWAITA` is set to `OFF` as it's placed outside the `if(TEXSTUDIO_BUILD_ADWAITA)` statement.

### Solution
Move the command `add_subdirectory(src/adwaita-qt)` into the `if` statement, so it's only executed if `TEXSTUDIO_BUILD_ADWAITA` is set to `ON`.


sunderme:
There is a if condition in the subdirectory.
This fix is superfluous

TomJampen:
Well, for the upload to Debian the subdirectory adwaita-qt is not part of the TeXstudio source code as there is already an adwaita-qt package in Debian and this code would be redundant. Thus, the build fails.

sunderme:
I don't see any sense in that fix.

TomJampen:
Thanks for pointing out that there's also an if-statement in the subdirectory (I missed that). I've removed that if-statement in the subdirectory as well. Now, it's handled identically to

- src/hunspell
- src/zlib-1.3.2
- src/quazip

In my opinion it makes sense to only add a subdirectory if it is actually used during build (as already done with hunspell, zlib and quazip).

And in the case of Debian, it avoids maintaining an unnecessary patch as I've removed adwaita-qt from the source code (like hunspell, zlib, quazip, etc.) because there's already a Debian package providing this functionality.

I've updated my branch but I cannot update the PR as it is closed.

Issue #4010, opened 2025-03-01
Syntax highlighting error with tikzpicture (+ beamer ?)


ysalmon:
### Environment
- TeXstudio: 4.8.6
- Qt: Qt 5.15.13 R
- OS: Linux KUbuntu 24.04
- TeX distribution:  texlive

### Expected behavior
Only the tikzpicture code should have tikzpicture syntax highlight.

### Actual behavior
The tikzpicture syntax highlight continues after the end of the tikzpicture environment.

![Image](https://github.com/user-attachments/assets/64c692fc-e0fe-4b76-b81e-d5ec5216c1b9)

In my actual working file, this goes on to the end of the document (more than 2000 lines).

### How to reproduce
Hard to tell exactly. Creating a MWE is difficult because removing some lines will cause the wrong behaviour to cease, but adding them back will not again show the bad behaviour.

This seems to have to do with the multi-line ``\visible``, since replacing it with a ``visibleenv`` environments alleviates the problem.

[document.txt (rename to .tex ; GitHub does not want .tex attachments)](https://github.com/user-attachments/files/19040588/document.txt)

sunderme:
without a step by step instruction on how to reproduce, there is not much I can do.
Out of the box, the highlighting is okay.
![Image](https://github.com/user-attachments/assets/7cec212f-26a8-4317-8b55-557ff3840406)

octaeder:
I think I see it:

![Image](https://github.com/user-attachments/assets/c76e757a-b126-41d0-9792-b82ca619ca3b)

octaeder:
When I use Alt+Return to close open env I get after `\end{frame}`
```
\end{text}
\end{tikzpicture}
\end{frame}
\end{document}
```

sunderme:
where is ` \end{text}`  coming from ?

octaeder:
good question, I did what I said. I hoped this could give some insight, what's going on

octaeder:
I may add that I saved the downloaded file and opened with the open dialog of txs.

sunderme:
The other strange thing us the right-arrow in line 53

octaeder:
> [@octaeder](https://github.com/octaeder) The other strange thing is the right-arrow in line 51

@sunderme you mean @ysalmon ?

sunderme:
> > [@octaeder](https://github.com/octaeder) The other strange thing is the right-arrow in line 51
> 
> @sunderme you mean @ysalmon ?

[...]

Issue #2088, opened 2022-01-28
No more picture environment syntax highlighting for other picture-environments


the-science-guy:
Ever since updating from TeXstudio verstion 4.1.1 to 4.2.0 and later the picture syntax highlighting of tikzpictures and other similar picture environments has disappeared, except for only "picture".

### Environment
- TeXstudio: 4.2.1
- Qt: Version 5.15.2
- OS: Arch Linux
- TeX distribution: texlive

### Expected behavior

-> orange syntax highlighting of picture environment
![old](https://user-images.githubusercontent.com/98602834/151579815-0035bdff-5eed-4a34-915d-243e50fca143.png)
v. 4.1.1

### Actual behavior

-> **NO** orange syntax highlighting of picture environments except only "picture"

![current](https://user-images.githubusercontent.com/98602834/151579838-d111d1ce-09e1-4326-b6ae-30a8f6bdb05f.png)
v. 4.2.0 & v. 4.2.1

### How to reproduce
-Use Version 4.2.0 or later of TeXstudio (be it on Linux or Windows)
-Use the package tikz
\begin{tikzpicture}
-> no orange syntax highlighting
\end{tikzpicture}

sunderme:
that was changed with bf9ab7a0 
@muzimuzhi any comments ?

the-science-guy:
> that was changed with [bf9ab7a](https://github.com/texstudio-org/texstudio/commit/bf9ab7a0967cc1422f124c2664cb479e88caac7f) @muzimuzhi any comments ?

Is there any way for me to turn it on again and revert this change?

sunderme:
generate a line ```\begin{tikzpicture}[options%keyvals]#\picture``` in a new cwl which need to be loaded.

muzimuzhi:
> that was changed with [bf9ab7a](https://github.com/texstudio-org/texstudio/commit/bf9ab7a0967cc1422f124c2664cb479e88caac7f) @muzimuzhi any comments ?

@sunderme Will adding back `\begin{tikzpicture}[options%keyvals]#\picture` make `tikzpicture` env accept commands that are only valid in `picture` env? If not then just do it. If so then perhaps `tikzpicture` and `pspicture(*)` can be registered like `picture` env in (if I guess right)
https://github.com/texstudio-org/texstudio/blob/15ed5bd1e3bcdbb5807b3257a35c3b8599208597/src/latexdocument.cpp#L3201-L3213

sunderme:
[...]

PR #4585, opened 2026-08-07
Preserve the alpha channel when saving a translucent background

vincenzopairona-creator:
# Preserve the alpha channel when saving a translucent background

`QFormatScheme` reads a format background with `QColor(value)`, which accepts Qt's
`#AARRGGBB` notation, but both save paths write it back with `QColor::name()`, which
always produces `#RRGGBB`. The alpha channel is silently dropped.

A format scheme with a translucent background therefore loads correctly, but is
flattened to fully opaque the first time the scheme is written back — turning a faint
shading into a solid block of colour.

## Before / after

The `comment` format is given the background `#40667299`: a 25% tint of the comment
text colour `#667299`. TeXstudio is started, closed so that the scheme is saved, and
started again.

**Before** — the stored value has become `#667299`, exactly the colour of the text it
sits behind, so the comments are no longer readable at all:

download

**After** — the stored value is still `#40667299` and the shading stays a faint hint:

download

Measured against the editor background `rgb(45,45,45)`:

| | comment background | expected |
| --- | --- | --- |
| before | `rgb(102,114,153)` | `#667299`, opaque |
| after | `rgb(60,63,72)` | `#667299` at 25% over `rgb(45,45,45)` |

## The change

`QColor::HexArgb` is used for translucent colours only; opaque colours keep
`QColor::name()`, so existing `.qxf` files and stored settings are written exactly as
before.

## Notes

No format shipped in `utilities/qxs/*.qxf` uses a translucent background today, and the
colour picker calls `QColorDialog::getColor()` without `ShowAlphaChannel`, so the alpha
cannot be set through the UI. The asymmetry is currently reachable by editing a `.qxf`
by hand, or from code adding a format.

`foreground` and `linescolor` are saved the same way and have the same asymmetry. They
are left untouched here, as no shipped or hand-written scheme is known to rely on a
translucent value for them — happy to include them if you prefer.

Verified with Qt 6.11.1 (ucrt64, Debug, Ninja), with an isolated configuration
directory (`--config`).


sunderme:
any particular reason why to use transparent background in the first place. It is expensive for compute.

vincenzopairona-creator:
You are right that the blend is not free, and the numbers below back your instinct on the
relative cost. Let me answer the two halves separately, since they turn out to be
independent questions.

## Why a translucent background in the first place

I am working on **syntax highlighting for TikZ**. Highlighting inside `tikzpicture` is
something users clearly exercise and report on — #2088 when it regressed in 4.2.0, #4010
when it leaked past the end of the environment — so it seemed worth making it actually
understand the language rather than colour it approximately.

Part of that is catching one of the most common mistakes people make when writing TikZ:
**forgetting the `;` at the end of a path command**. The errors TeX reports for this are
sometimes ambiguous — the missing terminator lets the next command be swallowed into the
unfinished path, so the failure can surface far from its cause, or as a message that points
somewhere unrelated. Flagging it in the editor, at the line where the statement actually
starts, is far more direct than reading it out of the log.

The marker for that is a shading over the offending line, and this is where the alpha
channel matters. In `qdocument.cpp` exactly one background is painted per format run — the
highest-priority valid one, via a single `fillRect`; format backgrounds are not composited
with one another. So a marker with an **opaque** colour replaces the line's ground outright
and reads as a hard error block, competing with the syntax colours it sits behind. A
**translucent** one composites over the base, keeps what is underneath legible, and reads
as what it is meant to be: a hint, not a failure.

That is also how this patch came about. It is not a cleanup I went looking for: while
building that marker I gave the format a translucent background, watched it work, and then
watched it turn into a solid block the first time the format scheme was written back. That
looked like a defect to me — loading understands `#AARRGGBB`, saving silently discards it —
so I fixed the asymmetry and proposed it on its own, separately from the feature that ran
into it. If you read it differently, I would rather hear that now than argue the point.

That feature is now #4587, if it helps to see what the shading is actually for — eight files,
+204 −4, with before/after screenshots in the description. It depends on this patch: without
it the shading is flattened to a solid block the first time the format scheme is saved. If you
would rather not take the alpha fix, say so on either thread and I will rework #4587 around an
opaque colour chosen per theme instead; the two are not tied together beyond this.

## What it costs

I measured the exact operation the patch concerns: the background fill in
`QDocumentPrivate::drawText` (`qdocument.cpp` ~3958), reproduced faithfully including the
surrounding `save()` / `setRenderHint(Antialiasing, false)` / `restore()`. Raster engine,
Qt 6.11.1, `-O2`, median of 3000 repaints, three independent runs.

Worst case: a 1200×800 viewport, 40 visible lines, 3 format runs each — 120 fills per
repaint, every single one translucent.

| | per repaint | per fill |
| --- | --- | --- |
| opaque background | 106.9 / 110.3 / 114.3 µs | ~0.92 µs |
| translucent background | 187.8 / 189.4 / 196.3 µs | ~1.59 µs |
| drawing the text alone | 1041 µs | — |

An alpha fill costs **+72–76%** over a solid one, consistently across runs. In absolute
terms that is **+0.67 µs per fill**: +80 µs for a fully shaded viewport, which is **0.5% of
a 60 Hz frame budget**, and under **8% of what drawing the text of those same lines already
costs**.

The realistic case is far below that worst case. The marker shades only the lines that
actually have an unterminated statement — usually one or two — so the added cost is roughly
**2 µs per shaded line per repaint**.

## Caveats

So the numbers are read for what they are: this is a micro-benchmark isolating the fill,
not a whole-frame measurement of the editor, and it does not capture cache or allocation
effects of a real repaint. It measures the operation the patch concerns and nothing else.
Happy to share the benchmark source if you want to run it yourself, or to measure something
closer to a real repaint if you would prefer that.

## What the patch does not change

None of the above is introduced by this patch. It touches no painting code — only
`QFormatScheme::save`. Unmodified TeXstudio already parses `#AARRGGBB`
(`setFormatOption` does `fmt.background = QColor(value)`, `qformatscheme.cpp:72`) and
already paints it: I checked with an unmodified build, and a hand-edited `.qxf` setting
`comment` to `#40667299` renders at `rgb(60,63,72)` over the editor ground `rgb(45,45,45)`,
identical to the patched build. The value is only flattened on save. Whatever the blend
costs is already being paid today by anyone who configures such a colour; the patch only
decides whether that colour survives being written back.

---

*Disclosure: this reply, including the benchmark and its measurements, was generated by
Claude.*

sunderme:
not needed

PR #4587, opened 2026-08-08
Flag a TikZ path command whose terminating ; is missing

vincenzopairona-creator:
Forgetting the `;` at the end of a `\draw`, `\node` or similar inside a `tikzpicture` is one of the most common mistakes in pgf/TikZ code. pgf ships a **dedicated error message** for exactly this slip rather than letting it fall through to a generic TeX failure, which is itself a fair indication of how often it happens.

> Reworked after the review below. The first version shaded the whole statement with a translucent background of its own; this one reports a plain syntax error on the token where the parse breaks. See *What changed* at the end. The original description is kept in a comment below, since the review refers to it.

## What TeX tells you today

The message pgf produces is a good one: it asks *"Did you forget a semicolon?"* outright. What it cannot do is point at the line to edit. TeX only notices the problem where the parser gives up, which is the **next** command — so the reported line is always below the one carrying the unterminated statement:

| | `;` missing at end of | pdflatex reports |
| --- | --- | --- |
| statement on one line | line 7 | `l.8` |
| statement over several lines, starting at line 5 | line 7 | `l.8` — three lines below the command |
| last statement before `\end{tikzpicture}` | line 6 | `l.7`, the `\end` line |

download

The diagnosis is right, the location is where parsing stopped rather than where the fix goes, and you only get it after a compile.

## What this adds

The syntax check tracks, per picture environment, whether a path command has been started and not yet terminated, and flags whatever turns up in place of the `;`: the next path command, or the `\end` of the environment. That is the same token TeX chokes on, so the editor and the compiler agree on where the evidence sits — the difference is that the editor says so while the text is being typed, and says what is missing.

It is a plain syntax error, handled like every other one:

- it uses the `latexSyntaxMistake` format, so it follows whatever the theme defines for syntax errors — no format of its own;
- hovering it explains the problem: **"semicolon in tikz missing"**;
- it is suppressed together with the rest when *Check syntax* is switched off.

**Before**

![nothing is flagged, the statements left open are indistinguishable][prima]

**After**

![the token that cannot follow the unterminated statement is flagged][dopo]

All three shapes are covered. The statement on line 6 occupies a single line, and is flagged on the `\draw` that cannot follow it. The one starting on line 11 runs to line 13, and is flagged on the `\node` below it. When the unterminated statement is the last in the picture, as on line 18, the `\end` of the environment is flagged instead — the command token itself, not the whole `\end{tikzpicture}`. The last picture is correct and stays unmarked.

Hovering the marker says what is missing, wherever it lands:

![tooltip over the flagged command on line 7][messaggio]

![tooltip over the flagged \end on line 19][messaggio-end]

Typing the missing `;` on line 6 clears that marker, while the unterminated statements in the other two pictures keep theirs:

![the marker on line 7 is gone, the other two remain][corretto]

## How it works

One diagram per flow, with the code it describes underneath.

### Where the missing `;` is detected — `src/syntaxcheck.cpp`, `checkLine()` token loop

```mermaid
flowchart TB
  subgraph A["AFTER — one block, +37 lines"]
    direction TB
    A1["for each token of the line"]
    A2{"`a command or punctuation,
with a picture env open?`"}
    A3["`a path command — **\\draw**, **\\node**, **\\fill** … 14 in all
a statement was already open:
**error on this token**`"]
    A4["`the **\\end** of the environment
a statement was open:
**error on this token**`"]
    A5["`a **;** at the statement's own level
statement closed, no error`"]
    A6["`the flagged token is always on the line being checked
no write to another line, no cookie, no vector`"]
    A1 --> A2
    A2 --> A3 --> A6
    A2 --> A4 --> A6
    A2 --> A5 --> A6
  end
  subgraph B["BEFORE — upstream"]
    direction TB
    B1["for each token of the line"]
    B2["`**\\end** pops the environment
a known command gets the environment highlight
an unknown one gives **ERR_unrecognizedCommand**`"]
    B3["`nothing tracks a path command left open
TeX reports it only at compile time,
where the parser gives up`"]
    B1 --> B2 --> B3
  end
  classDef added stroke:#2da44e,stroke-width:2px
  classDef removed stroke:#cf222e,stroke-width:2px
  class A2,A3,A4,A5,A6 added
```

```diff
@@ void SyntaxCheck::checkLine(...)
+        if (tk.type == Token::command || tk.type == Token::punctuation) {
+            const int pictureEnv = pictureEnvIndex(activeEnv);
+            if (pictureEnv >= 0) {
+                static const QSet<QString> pgfPathCommands = {
+                    "\\path", "\\draw", "\\fill", "\\filldraw", "\\shade", "\\shadedraw", "\\pattern",
+                    "\\clip", "\\useasboundingbox", "\\node", "\\coordinate", "\\matrix", "\\pic", "\\graph"
+                };
+                const QString tokenText = tk.getText();
+                const bool startsPath = (tk.type == Token::command && pgfPathCommands.contains(tokenText));
+                const bool endsEnv = (tk.type == Token::command && tokenText == "\\end");
+                if (startsPath || endsEnv) {
+                    if (activeEnv.at(pictureEnv).pictureStatementOpen) {
+                        Error elem;
+                        elem.type = ERR_missingSemicolonInPicture;
+                        elem.range = QPair<int, int>(tk.start, tk.length);
+                        newRanges.append(elem);
+                    }
+                    activeEnv[pictureEnv].pictureStatementOpen = startsPath;
+                    if (startsPath)
+                        activeEnv[pictureEnv].pictureStatementLevel = tk.level;
+                } else if (tk.type == Token::punctuation && tokenText == ";"
+                           && activeEnv.at(pictureEnv).pictureStatementOpen
+                           && tk.level <= activeEnv.at(pictureEnv).pictureStatementLevel) {
+                    activeEnv[pictureEnv].pictureStatementOpen = false;
+                }
+            }
+        }
```

<sup>Comments stripped here for length; the full block, and `pictureEnvIndex()` beside it, are in the diff.</sup>

### Why the tooltip said “no error” — `src/syntaxcheck.cpp`, `getErrorAt()`

```mermaid
flowchart TB
  subgraph AA["AFTER — one skip, +5 −3 lines"]
    direction TB
    D1["`hovering **\\draw** on line 7`"]
    D2["`skip every **ERR_highlight**
while looking for the error`"]
    D3["`1 · environment highlight — skipped
2 · missing-semicolon error — kept
3 · keyword highlight — skipped`"]
    D4["`tooltip: **semicolon in tikz missing**`"]
    D1 --> D2 --> D3 --> D4
  end
  subgraph BB["BEFORE"]
    direction TB
    C1["`hovering **\\draw** on line 7`"]
    C2["`1 · environment highlight, whole line — prepended
2 · missing-semicolon error, on **\\draw**
3 · keyword highlight, on **\\draw** — appended`"]
    C3["`the lookup keeps the last range
covering the column, so no. 3 wins`"]
    C4["`**ERR_highlight** becomes **ERR_none**`"]
    C5["tooltip: no error"]
    C1 --> C2 --> C3 --> C4 --> C5
  end
  classDef added stroke:#2da44e,stroke-width:2px
  classDef removed stroke:#cf222e,stroke-width:2px
  class D2 added
  class C4 removed
```

```diff
@@ QString SyntaxCheck::getErrorAt(...)
 	ErrorType result = ERR_none;
 	foreach (const Error &elem, newRanges) {
+        // A highlight is not an error and has no message of its own. Skipping it here rather than
+        // discarding the result afterwards matters because the highlight of the surrounding
+        // environment spans the whole line and is appended last: left in, it would mask every real
+        // error inside a math or picture environment and report "no error" for it.
+        if (elem.type == ERR_highlight) continue;
 		if (elem.range.second + elem.range.first < pos) continue;
 		if (elem.range.first > pos) break;
 		result = elem.type;
 	}
-    if(result==ERR_highlight){
-        result=ERR_none; // filter out accidental highlight detection (test only)
-    }
```

Only a token that its environment highlights *and* that also carries an error reaches this. No error type in use today does, which is why it could not surface before.

### The state carried from line to line — `src/syntaxcheck.h`

```mermaid
flowchart TB
  subgraph AA["AFTER — +2 fields, +1 error type"]
    direction TB
    F1["`class **Environment** gains two fields
**+ bool pictureStatementOpen** — a path command is open
**+ int pictureStatementLevel** — which **;** closes it`"]
    F2["`**operator==** also compares **pictureStatementOpen**
so the following lines are rechecked when it changes`"]
    F3["`the level is left out on purpose: it differs from
line to line and would stop stacks from ever being equal again`"]
    F4["`enum **ErrorType**
**+ ERR_missingSemicolonInPicture**
message: semicolon in tikz missing`"]
    F1 --> F2 --> F3 --> F4
  end
  subgraph BB["BEFORE — upstream"]
    direction TB
    E1["`class **Environment**
name, origName, id, excessCol, runAway,
dlh, startingColumn, endingColumn, ticket, level`"]
    E2["`**operator==** compares name, id, excessCol, origName, level
equal stacks stop the check cascading down`"]
    E3["`enum **ErrorType** — 16 entries`"]
    E1 --> E2 --> E3
  end
  classDef added stroke:#2da44e,stroke-width:2px
  class F1,F2,F4 added
```

```diff
@@ class Environment
     int level; ///< command level (see tokens) in order to handle nested commands like \shortstack
 
+    bool pictureStatementOpen = false;
+    int pictureStatementLevel = 0;
+
 	bool operator ==(const Environment &env) const
 	{
-        return (name == env.name) && ... && (level == env.level);
+        return (name == env.name) && ... && (level == env.level)
+                && (pictureStatementOpen == env.pictureStatementOpen);
 	}

@@ enum ErrorType
 		ERR_commandOutsideEnv,
+        ERR_missingSemicolonInPicture,
         ERR_spelling,
         ERR_highlight,
 		ERR_MAX
```

## What changed since the review

@sunderme, taking your points in order:

- **"asynchronous processing"** — the check now stays entirely on the line being examined. The flagged token is the one where the parse breaks, and that token is on the line the checker thread is holding, so nothing is written to any other line handle. The previous version marked lines above the evidence, which is what you objected to.
- **"cookie"** — gone. It existed only to let a line remember a verdict reached while checking a later line; with nothing marked out of line, there is nothing to remember. `UNTERMINATED_STATEMENT_COOKIE` is no longer added.
- **"adding fields to `Environment`"** — down from four (including a `QVector` of line handles) to two: a `bool` and an `int`.
- **"it should be an error, not a hint"** — agreed, and this is the change that made the rest fall out. `ERR_missingSemicolonInPicture` with a message of its own, rather than `ERR_highlight` with a background of its own. `ERR_highlight` is meant for colouring kinds of region (math, verbatim, picture); a problem carried by it has no message and cannot be switched off, which was the real cost of the first approach.
- **"similar environments"** — picture environments are not matched by name but through the `#\pictureHighlight` alias the cwl files already give them, so `pgfpicture` and anything a package marks the same way are covered without naming them.
- **"the `\tikz` command"** — still not handled, deliberately. It is not currently treated as a picture environment at all, and giving it a scope that ends at the first `;` needs a runaway guard of its own to keep a forgotten `;` from running to the end of the document. Happy to do it as a follow-up.
- **hardcoded command list** — unchanged in this version: 14 path-opening commands as a `QSet` in the checker. I agree the cwl files are where command knowledge belongs, but as far as I can tell the format has no way to express it yet: `#/env` registers a command under a name and also restricts where it is valid, `#\name` on a line that is not `\begin{...}` is read and dropped, and `#s#%class` means "this command defines entries for that database". It would take a new classifier — I suggest `#%pathcommand`, in the comment below. That changes the cwl format, so it seemed your call rather than something to decide here.

## One thing fixed on the way

`getErrorAt()` now skips `ERR_highlight` while looking for the error under the cursor, instead of discarding the result afterwards. Errors and highlights share one list and the lookup keeps the last range covering the column; a command which its environment highlights gets an `ERR_highlight` range appended *after* the error on the very same token, so the tooltip reported "no error" for `\draw` and `\node`. No error type in use today lands on a token that its environment highlights, so this could not surface before.

## Scope

Two files, +73 −5, against the first version's eight files and +204 −4.

## No longer depends on #4585

The first version needed a translucent background to keep the syntax colours readable underneath, which is why it depended on #4585. With the theme's own error format there is no alpha channel involved and the two are now independent. #4585 still stands on its own — `QFormatScheme::save` drops the alpha of a background that loading accepts — but nothing here waits on it.

[prima]: https://github.com/user-attachments/assets/e5359225-f3a0-4984-b9f9-dfd6a79a0a4e
[dopo]: https://github.com/user-attachments/assets/01ce2fc1-101e-426c-80fa-c6b85baf0b0a
[messaggio]: https://github.com/user-attachments/assets/a42d9128-c0c3-4f93-a3be-6ebcaab8e471
[messaggio-end]: https://github.com/user-attachments/assets/22dadc02-f8a1-40f5-acc7-8c57dc00bd98
[corretto]: https://github.com/user-attachments/assets/2236a3b8-28f8-456f-b9ed-260ed3009f8f


sunderme:
I can't say that I like the code/approach.
1) the latex commands it acts on are hardcoded in the syntax checker. There is a reason why txs has  elaborate cwl files.
2) the highlight is directly written from inside the checkLine function to other lines. That is avoided deliberately because syntax check asynchronously processes the lines and that lines may be outdated.
3) it adds fields to Environment and stores even vectors in it. txs is not memory efficient but that is stretching it.
4) it uses an extra cookie (additionally to the env), I have not yet tried to figure out why
5) I don't follow the approach that it should be a "hint" that the semicolon is missing. The latex compile fails with an error. It should be an error on the next node command or "\end". (semicolon in tikz missing)
6) a bit what aboutism: there are other envs which act like tikzpicture which show the same behaviour and are not handled, same with `\tikz`.

muzimuzhi:
Parsing tikz path is hard. Guessing where a syntactical terminator is missing is even harder.

Instead, it's easier and more robust to report on which line a tikz path started, so the error message can be more helpful.

```tex
\documentclass{article}
\usepackage{tikz}
\usepackage{xpatch}

\makeatletter
\pretocmd\tikz@command@path
  {\edef\tikz@command@start@lineno{\the\inputlineno}}
  {}{\PatchFailed}

\patchcmd\tikz@expand
  {\tikzerror{Giving up on this path. Did you forget a semicolon?}}
  {\tikzerror{Giving up on this path started on line   
   \tikz@command@start@lineno. Did you forget a semicolon?}}
  {}{\Patchfailed}
\makeatother

\begin{document}
\begin{tikzpicture}
  \draw (0,0) -- (1,1)
  \node {a};
\end{tikzpicture}

\begin{tikzpicture}
  \draw (0,0) -- +(1,1)
        (1,0) -- +(1,1)
  \node {a};
\end{tikzpicture}
\end{document}
```

Error messages in log
```diff
-! Package tikz Error: Giving up on this path. Did you forget a semicolon?.
+! Package tikz Error: Giving up on this path started on line 20. Did you forget
+ a semicolon?.

-! Package tikz Error: Giving up on this path. Did you forget a semicolon?.
+! Package tikz Error: Giving up on this path started on line 25. Did you forget
+ a semicolon?.

-! Package tikz Error: Giving up on this path. Did you forget a semicolon?.
+! Package tikz Error: Giving up on this path started on line 32. Did you forget
+ a semicolon?.
```

vincenzopairona-creator:
@muzimuzhi 
> Parsing tikz path is hard

Look at [my attempt](https://github.com/texstudio-org/texstudio/compare/master...vincenzopairona-creator:feature/tikz-syntax-highlighting?expand=1)

vincenzopairona-creator:
Keeping the original description here for the record, since the comments below refer to it. The description above has been replaced with the reworked version.

<details>
<summary>Original description (translucent-shading version, commit f7e0c16)</summary>

# Flag a TikZ path command whose terminating `;` is missing

Forgetting the `;` at the end of a `\draw`, `\node` or similar inside a `tikzpicture` is one of
the most common mistakes in pgf/TikZ code. pgf ships a **dedicated error message** for exactly
this slip rather than letting it fall through to a generic TeX failure, which is itself a fair
indication of how often it happens.

@muzimuzhi, @samcarter — you both work on pgf/TikZ and on TeXstudio, so you probably have a
better sense than most of how often this one comes up. Does it match what you see?

## What TeX tells you today

The message pgf produces is a good one: it asks *"Did you forget a semicolon?"* outright. What
it cannot do is point at the line to edit. TeX only notices the problem where the parser gives
up, which is the **next** command — so the reported line is always below the one carrying the
unterminated statement:

| | `;` missing at end of | pdflatex reports |
| --- | --- | --- |
| statement on one line | line 7 | `l.8` |
| statement over several lines, starting at line 5 | line 7 | `l.8` — three lines below the command |
| last statement before `\end{tikzpicture}` | line 6 | `l.7`, the `\end` line |

download

The diagnosis is right, the location is where parsing stopped rather than where the fix goes,
and you only get it after a compile.

## What this adds

The syntax check tracks, per picture environment, whether a path command has been started and
not yet terminated, and shades the statement until the `;` is typed — while editing, before any
compile.

**Before**

download

**After**

download

Both shapes are covered: the statement on line 7 occupies a single line, the one starting on
line 12 runs to line 14 and all three lines are shaded. Continuation lines are shaded from their
first non-blank character, so the indentation stays clear of the band.

Typing the missing `;` clears the shading from every line of the statement at once, while the
unterminated statement in the other picture keeps its own:

download

Measured against the editor background `rgb(45,45,45)`, every shaded line reads `rgb(84,84,84)`
and the correctly terminated ones stay at `rgb(45,45,45)`.

## How it works

Because the verdict only becomes apparent while checking a *later* line — where the next command
starts, or where the environment ends — the lines the statement spans are collected in the
environment as they are passed, and the column each is shaded from is stored on the line itself
as a cookie. That way the shading survives a recheck of any one of those lines on its own.

Only `pictureStatementOpen` joins `Environment::operator==`. The line handles, columns and
tickets are deliberately left out: they differ from line to line, and comparing them would stop
environment stacks from ever being equal again, which is what lets the syntax check stop
cascading down the document.

## Scope

Eight files, +204 −4. It covers `\begin{tikzpicture}` environments, which TeXstudio already
recognises as picture environments. The short `\tikz ... ;` form is not handled: it is not
currently treated as a picture environment at all, and giving it a scope that ends at the first
`;` needs a runaway guard of its own to keep a forgotten `;` from running to the end of the
document. That seemed better kept out of this change.

## Depends on #4585

The shading uses a translucent background so that it reads as a hint rather than an error block:
`qdocument.cpp` paints a single background per format run, so an opaque colour would replace the
line's ground outright and hide the syntax colours underneath.

That relies on #4585, which stops `QFormatScheme::save` from discarding the alpha channel.
Without it the colour survives loading but is flattened to opaque the first time the format
scheme is written back. If #4585 is not the direction you want to take, this can be reworked
around an opaque colour chosen per theme instead — say so and I will adjust it.

</details>

src/syntaxcheck.cpp
```diff
@@ -308,6 +310,7 @@ QString SyntaxCheck::getErrorAt(QDocumentLineHandle *dlh, int pos, StackEnvironm
 			<< tr("unrecognized key in key option")
 			<< tr("unrecognized value in key option")
             << tr("command outside suitable env")
+            << tr("semicolon in tikz missing")
```
sunderme, line 313: should probably be something like "semicolon in previous tikz command missing" or similar

src/syntaxcheck.h
```diff
@@ -30,13 +30,22 @@ class Environment
 	int ticket;
     int level; ///< command level (see tokens) in order to handle nested commands like \shortstack
 
+    /// A path command inside a picture environment (\draw, \node, ...) has been started and its
+    /// terminating ";" has not been seen yet. Part of operator==, so that a change propagates to the
+    /// following lines: whether they may open a path command of their own depends on it.
+    bool pictureStatementOpen = false;
+    /// command level (see tokens) of that path command, so that a ";" nested deeper -- inside the text
+    /// of a node, say -- is not taken for the terminating one
+    int pictureStatementLevel = 0;
```
sunderme, line 39: I don't see why there needs to be a new variable. 
Just reuse int level from above. It is exactly defined for this case

src/syntaxcheck.h
```diff
@@ -30,13 +30,22 @@ class Environment
 	int ticket;
     int level; ///< command level (see tokens) in order to handle nested commands like \shortstack
 
+    /// A path command inside a picture environment (\draw, \node, ...) has been started and its
+    /// terminating ";" has not been seen yet. Part of operator==, so that a change propagates to the
+    /// following lines: whether they may open a path command of their own depends on it.
+    bool pictureStatementOpen = false;
```
sunderme, line 36: there should be no new variable.
Though names may not match, but any variable above can be used

src/syntaxcheck.cpp
```diff
@@ -690,6 +710,43 @@ void SyntaxCheck::checkLine(const QString &line, Ranges &newRanges, StackEnviron
...
 
+        // A path command inside a picture environment has to be terminated by ";". Track whether that
+        // ";" is still pending, and flag whatever turns up in its place. This runs before the
+        // \begin/\end handling further down, so that the picture environment is still on the stack when
+        // its \end is seen.
+        if (tk.type == Token::command || tk.type == Token::punctuation) {
+            const int pictureEnv = pictureEnvIndex(activeEnv);
+            if (pictureEnv >= 0) {
+                // the commands of pgf/TikZ which open a path and need a ";" of their own
+                static const QSet<QString> pgfPathCommands = {
```
sunderme, line 721: again, no hard coded commands.
We can extend cwl format and define a "nodes" class.
Typical check is is `ltxCommands->possibleCommands["%node"].contains(word)`

src/syntaxcheck.cpp
```diff
@@ -690,6 +710,43 @@ void SyntaxCheck::checkLine(const QString &line, Ranges &newRanges, StackEnviron
...
+                    "\\path", "\\draw", "\\fill", "\\filldraw", "\\shade", "\\shadedraw", "\\pattern",
+                    "\\clip", "\\useasboundingbox", "\\node", "\\coordinate", "\\matrix", "\\pic", "\\graph"
+                };
+                const QString tokenText = tk.getText();
+                const bool startsPath = (tk.type == Token::command && pgfPathCommands.contains(tokenText));
+                const bool endsEnv = (tk.type == Token::command && tokenText == "\\end");
```
sunderme, line 727: this code fails: example
```
\node  at (0,1) {one line, no semicolon \begin{Large}
		content...
\end{Large}};
```

src/syntaxcheck.cpp
```diff
@@ -568,6 +571,23 @@ bool SyntaxCheck::checkCommand(const QString &cmd, const StackEnvironment &envs)
...
 }
 
+/*!
+ * \brief index of the innermost picture environment on the stack, -1 if there is none
+ *
+ * Picture environments are not recognized by name but by the "#\pictureHighlight" alias the cwl files
+ * give them, so that tikzpicture, pgfpicture and anything a package adds are all covered.
+ * \param envs environment stack
+ */
+int SyntaxCheck::pictureEnvIndex(const StackEnvironment &envs)
```
sunderme, line 581: this should be a bit more agnostic, in case we want to use that approach in other aspects.
agnostic: -> getEnvIndex(envName, &stack)

src/syntaxcheck.cpp
```diff
@@ -690,6 +710,43 @@ void SyntaxCheck::checkLine(const QString &line, Ranges &newRanges, StackEnviron
...
+                        elem.type = ERR_missingSemicolonInPicture;
+                        elem.range = QPair<int, int>(tk.start, tk.length);
+                        newRanges.append(elem);
+                    }
+                    activeEnv[pictureEnv].pictureStatementOpen = startsPath;
+                    if (startsPath)
+                        activeEnv[pictureEnv].pictureStatementLevel = tk.level;
+                } else if (tk.type == Token::punctuation && tokenText == ";"
+                           && activeEnv.at(pictureEnv).pictureStatementOpen
+                           && tk.level <= activeEnv.at(pictureEnv).pictureStatementLevel) {
```
sunderme, line 743: It is unclear how nested node should look like but the following expression compiles but gets flagged:
```
\node  at (0,1) {one line, no semicolon \tikz \node (0,0) {content};};
``

sunderme (changes requested):
From maintaining point of view, this is better.
It is probably a good idea to add tests.
In general, an alternative approach would be to use a new "env" which is added for node commands and taken out for semicolon.
Then an extra check is needed for "\end env" which takes out any nodes and marks an error. (also adding new node on the same level gives an error)

src/syntaxcheck.cpp
vincenzopairona-creator: Ok.

sunderme:
I have generated a PR which basically implements the function more aligned with my usual style.
It is probably simpler to test that and give feedback (or extend the tests)
PR: #4589

sunderme:
I am closing this.
I have taken it as a feature request and moved to PR #4589

PR #4589, opened 2026-08-10
Tikz semicolon

sunderme:
check if tikz node are missing semicolons.

Inspired by #4587

### DIFF
--- a/src/latexpackage.cpp
+++ b/src/latexpackage.cpp
@@ -235,6 +235,9 @@ LatexPackage loadCwlFile(const QString fileName, LatexCompleterConfig *config, Q
 					cd.bracketCommand=true;
 					valid.remove("K");
 				}
+                if(line.endsWith(";")){
+                    package.possibleCommands["%semicolonEnd"]<<cmd;
+                }
 				if (cmd == "\\begin") {
 					if (!package.commandDescriptions.contains(cmd)) {
 						// one insertion of a general \begin-command
--- a/src/syntaxcheck.cpp
+++ b/src/syntaxcheck.cpp
@@ -282,13 +282,11 @@ QString SyntaxCheck::getErrorAt(QDocumentLineHandle *dlh, int pos, StackEnvironm
 	// find Error at Position
 	ErrorType result = ERR_none;
 	foreach (const Error &elem, newRanges) {
+        if (elem.type == ERR_highlight) continue;
 		if (elem.range.second + elem.range.first < pos) continue;
 		if (elem.range.first > pos) break;
 		result = elem.type;
 	}
-    if(result==ERR_highlight){
-        result=ERR_none; // filter out accidental highlight detection (test only)
-    }
 	// now generate Error message
 
 	QStringList messages;  // indices have to match ErrorType
@@ -308,6 +306,7 @@ QString SyntaxCheck::getErrorAt(QDocumentLineHandle *dlh, int pos, StackEnvironm
 			<< tr("unrecognized key in key option")
 			<< tr("unrecognized value in key option")
             << tr("command outside suitable env")
+            << tr("semicolon in previous tikz command missing")
             << tr("spelling")
             << "highlight"; // mock message for arbitrary highlight. Will not be shown.
 	Q_ASSERT(messages.length() == ERR_MAX);
@@ -793,6 +792,14 @@ void SyntaxCheck::checkLine(const QString &line, Ranges &newRanges, StackEnviron
 				elem.type = ERR_MathCommandOutsideMath;
 				newRanges.append(elem);
 			}
+            // closing semicolon for tikz nodes
+            if(word==";" && topEnv("%node", activeEnv)>0){
+                // check if there is a node command in the line
+                if(topEnv("%node",activeEnv)>0 && activeEnv.top().level==tk.level){
+                    activeEnv.pop(); // remove node env from stack
+                }
+            }
+
 		}
         // rainbow delimiter
         if(mShowRainbowDelimiter && tk.type==Token::braces){
@@ -1173,7 +1180,6 @@ void SyntaxCheck::checkLine(const QString &line, Ranges &newRanges, StackEnviron
 			activeEnv.push(tp);
 		}
 
-
         if (tk.type == Token::command) {
             QString word = line.mid(tk.start, tk.length);
             if (word.contains('@')) {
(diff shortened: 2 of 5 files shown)


vincenzopairona-creator:
image

When a command that a cwl file declares with a mandatory argument is used **without** that
argument, the lexer keeps looking for it and consumes what follows. Everything after it on the
line ends up one token level deeper than it should be.

The case that led me here is `\r`. `latex-document.cwl` declares it as the ring accent:

```
\r{letter}
```

so it is known in every document. In pgf/TikZ code `\r` is also a natural name for a `\foreach`
variable, and the pgf manual itself uses it that way:

```tex
\foreach \x / \y / \r in {0 / 0 / 2mm, 1 / 1 / 3mm, 2 / 0 / 1mm}
  \draw (\x,\y) circle [radius=\r];
```

Reduced to two lines:

```tex
\begin{tikzpicture}
\draw (0,0) circle [radius=\r];
\end{tikzpicture}
```

## What is and is not the trigger

I measured the variants one at a time. Inside the same key bracket:

| value inside `circle [radius=...]` | levels after it |
| --- | --- |
| `\zzz`, not declared anywhere | correct |
| `\linewidth`, declared, takes no argument | correct |
| `\r`, declared with a mandatory argument, none given | **shifted** |
| `\r{x}`, its argument given | correct |

And the same `\r` elsewhere on the path:

| position | levels after it |
| --- | --- |
| bare in the path, `\draw \r -- (1,1);` | correct |
| in the option bracket, `\draw[line width=\r] (0,0) -- (1,1);` | correct |
| in a key-value bracket, `circle [radius=\r]` | **shifted** |

So an *unknown* command is harmless — it is a *known* one whose mandatory argument is missing
that does the damage, and only inside a key-value argument.

## Why a user definition does not help

My first thought was that redefining the command in the document would take precedence:

```tex
\newcommand{\r}{1pt}
```

It does not, and the reason is structural. There are two separate stores:

- `LatexParser::commandDefs` holds the argument descriptions, and it is filled **only** from cwl
  packages (`latexParser.commandDefs.unite(pck.commandDescriptions)`);
- `possibleCommands["user"]` receives commands defined in the document, and it is a **set of
  names** with no argument information at all.

The lexer consults the first one:

```cpp
// latexparsing.cpp
if (lp->commandDefs.contains(command) && tk.subtype != Token::definition) {
    CommandDescription cd = lp->commandDefs.value(command);
```

so `\r` still resolves to `\r{letter}` and still expects a letter, whatever the document says.
There is currently no way for a document to override an argument description that came from a
cwl file.

## Why it is worth fixing

On its own the wrong level is invisible: nothing in the current syntax check reports it. It
becomes visible as soon as something relies on the levels being right. In the missing-semicolon
check the `;` of the statement ends up at a higher level than the command that opened it, so the
statement is never closed and an error is reported on the following command or on `\end` —
although the `;` is there.

Any future feature keyed on token levels inherits the same problem, so it seems better to fix it
where it starts rather than to work around it in each caller.

The names involved are not exotic. LaTeX defines a whole family of one-letter commands taking a
mandatory argument — `\r`, `\c`, `\v`, `\t`, `\u`, `\d`, `\b`, `\H` — and those are exactly the
names people pick for loop variables and local shorthands.

## Suggested fix

Where the lexer takes a mandatory argument for a command, it could check that the next token is
actually the expected opening delimiter, and give the argument up when it is not, instead of
treating whatever follows as the argument. A command that has lost its argument would then stop
affecting the tokens after it, which is what happens in TeX as well — there the accent simply
applies to the next character and the group structure is unaffected.

A second, independent improvement would be to let a definition in the document produce a real
`CommandDescription` that shadows the cwl one. `\newcommand{\r}{1pt}` knows it takes no
arguments, and honouring that would resolve this whole class of name collisions rather than this
one command.

I have not implemented either, and the second one clearly needs a decision rather than a patch.

## How it was found

I extracted every `codeexample` block of the pgf manual that contains a `tikzpicture` — 1025 of
them, excluding the deliberately incomplete fragments of the map tutorial — and ran the syntax
check over all of them, comparing the reported lines against an unpatched build. On code that is
correct by construction, every added report is a false positive. Two examples out of 1025 came
out flagged, and this is the cause of one of them.

sunderme:
> image
> 
> When a command that a cwl file declares with a mandatory argument is used **without** that argument, the lexer keeps looking for it and consumes what follows. Everything after it on the line ends up one token level deeper than it should be.
> 
> The case that led me here is `\r`. `latex-document.cwl` declares it as the ring accent:
> 
> ```
> \r{letter}
> ```
> 
> so it is known in every document. In pgf/TikZ code `\r` is also a natural name for a `\foreach` variable, and the pgf manual itself uses it that way:
> 
> ```
> \foreach \x / \y / \r in {0 / 0 / 2mm, 1 / 1 / 3mm, 2 / 0 / 1mm}
>   \draw (\x,\y) circle [radius=\r];
> ```
> 
> Reduced to two lines:
> 
> ```
> \begin{tikzpicture}
> \draw (0,0) circle [radius=\r];
> \end{tikzpicture}
> ```
> 
> ## What is and is not the trigger
> 
> I measured the variants one at a time. Inside the same key bracket:
> value inside `circle [radius=...]` 	levels after it
> `\zzz`, not declared anywhere 	correct
> `\linewidth`, declared, takes no argument 	correct
> `\r`, declared with a mandatory argument, none given 	**shifted**
> `\r{x}`, its argument given 	correct
> 
> And the same `\r` elsewhere on the path:
> position 	levels after it
> bare in the path, `\draw \r -- (1,1);` 	correct
> in the option bracket, `\draw[line width=\r] (0,0) -- (1,1);` 	correct
> in a key-value bracket, `circle [radius=\r]` 	**shifted**
> 
> So an _unknown_ command is harmless — it is a _known_ one whose mandatory argument is missing that does the damage, and only inside a key-value argument.
> ## Why a user definition does not help
> 
> My first thought was that redefining the command in the document would take precedence:
> 
> ```
> \newcommand{\r}{1pt}
> ```
> 
> It does not, and the reason is structural. There are two separate stores:
> 
>     * `LatexParser::commandDefs` holds the argument descriptions, and it is filled **only** from cwl
>       packages (`latexParser.commandDefs.unite(pck.commandDescriptions)`);
> 
>     * `possibleCommands["user"]` receives commands defined in the document, and it is a **set of
>       names** with no argument information at all.
> 
> 
> The lexer consults the first one:
> 
> ```c++
> // latexparsing.cpp
> if (lp->commandDefs.contains(command) && tk.subtype != Token::definition) {
>     CommandDescription cd = lp->commandDefs.value(command);
> ```
> 
> so `\r` still resolves to `\r{letter}` and still expects a letter, whatever the document says. There is currently no way for a document to override an argument description that came from a cwl file.
> ## Why it is worth fixing
> 
> On its own the wrong level is invisible: nothing in the current syntax check reports it. It becomes visible as soon as something relies on the levels being right. In the missing-semicolon check the `;` of the statement ends up at a higher level than the command that opened it, so the statement is never closed and an error is reported on the following command or on `\end` — although the `;` is there.
> 
> Any future feature keyed on token levels inherits the same problem, so it seems better to fix it where it starts rather than to work around it in each caller.
> 
> The names involved are not exotic. LaTeX defines a whole family of one-letter commands taking a mandatory argument — `\r`, `\c`, `\v`, `\t`, `\u`, `\d`, `\b`, `\H` — and those are exactly the names people pick for loop variables and local shorthands.
> ## Suggested fix
> 
> Where the lexer takes a mandatory argument for a command, it could check that the next token is actually the expected opening delimiter, and give the argument up when it is not, instead of treating whatever follows as the argument. A command that has lost its argument would then stop affecting the tokens after it, which is what happens in TeX as well — there the accent simply applies to the next character and the group structure is unaffected.
> 
> A second, independent improvement would be to let a definition in the document produce a real `CommandDescription` that shadows the cwl one. `\newcommand{\r}{1pt}` knows it takes no arguments, and honouring that would resolve this whole class of name collisions rather than this one command.
> 
> I have not implemented either, and the second one clearly needs a decision rather than a patch.
> ## How it was found
> 
> I extracted every `codeexample` block of the pgf manual that contains a `tikzpicture` — 1025 of them, excluding the deliberately incomplete fragments of the map tutorial — and ran the syntax check over all of them, comparing the reported lines against an unpatched build. On code that is correct by construction, every added report is a false positive. Two examples out of 1025 came out flagged, and this is the cause of one of them.

This is basically a new feature. I will handle it as extra request and extra PR.

PR #4630, opened 2026-08-28
Fix thesaurus dialog freeze on empty search input (#4629)

### LINKED ISSUE #4629: [bug]Thesaurus dialog: freeze when clicking "Starts With ..."/"Contains ..." with empty search field
### Environment                                                                                                                        
- TeXstudio: 4.9.7 (also reproducible on current master)                                                                       
- Qt: 6.9.1                                                                                                                          
- OS: platform-independent (reproduced on Windows and Linux)                                                                           
- TeX distribution: not relevant                                                                                                       
                                                                                                                                            
### Expected behavior                                                                                                                  
In the Thesaurus dialog, clicking "Starts With ..." or "Contains ..." with an empty search field should do nothing (or show an empty result) and return immediately.                                                                                                                           
                                                                                                                                            
### Actual behavior                                                                                                                    
Clicking "Starts With ..." or "Contains ..." with an empty input field freezes the whole application (UI thread blocked, needs to be killed).                                                                         
                                                                                                                                            
Root cause: in `ThesaurusDialog::startsWithClicked()` and `ThesaurusDialog::containsClicked()` (src/thesaurusdialog.cpp), an empty `word` makes `skey.startsWith("")` / `skey.contains("")` true for every key,so the loop never breaks and iterates the entire thesaurus map. Each match isadditionally deduplicated via `replacelistWidget->findItems(...)` which is O(n) per item, making the whole operation O(n^2) over the dictionary size — effectively a permanent freeze with large dictionaries.

Suggested fix: return early on empty input in both slots, e.g.`if (word.isEmpty()) return;` after the regex cleanup.                                                                                 
                                                                                                                                            
### How to reproduce                                                                                                                   
1. Open a document, open the Thesaurus dialog (Tools -> Thesaurus).                                                                    
2. Clear the search input field so it is empty.                                                                                        
3. Click "Starts With ..." (or "Contains ...").                                                                                        
4. The application stops responding.

Moximxxx:
Closes #4629

### Problem

In the Thesaurus dialog, clicking "Starts With ..." or "Contains ..." with an
empty search field freezes the whole application (see #4629).

An empty `word` makes `skey.startsWith("")` / `skey.contains("")` true for
every key, so the loop never breaks and iterates the entire thesaurus map.
Each match is additionally deduplicated via `replacelistWidget->findItems(...)`
which is O(n) per item — O(n^2) over the dictionary size, effectively a
permanent freeze with large dictionaries.

### Fix

Return early on empty input in `startsWithClicked()` and `containsClicked()`,
right after the regex cleanup:

```cpp
if (word.isEmpty()) return; // empty prefix matches every key -> O(n^2) freeze
```

### Testing                                                                                                                            
- Empty input + "Starts With ..." / "Contains ...": returns immediately, no freeze                                                     
- Non-empty input: prefix/contains search works as before

### DIFF
--- a/src/thesaurusdialog.cpp
+++ b/src/thesaurusdialog.cpp
@@ -318,6 +318,7 @@ void ThesaurusDialog::containsClicked()
 	if (!thesaurus) return;
 	QString word = replaceWrdLe->text().trimmed().toLower();
     word.replace(QRegularExpression(" \\(.*"), "");
+	if (word.isEmpty()) return; // empty needle matches every key -> O(n^2) freeze
 	classlistWidget->clear();
 	replacelistWidget->clear();
     QMultiMap<QString, ThesaurusDatabaseType::TinyStringRef>::const_iterator i = thesaurus->thesaurus.constBegin();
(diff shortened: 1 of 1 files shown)


sunderme:
thanks

